PR #6, opened 2018-05-14
Do not use the WebKitWebView::web-process-crashed signal

### LINKED ISSUE #3: WebView's 'web-process-crashed' signal needs updating
(cog:1490): GLib-GObject-WARNING **: ../../glib-2.46.2/gobject/gsignal.c:2516: signal 'web-process-crashed' is invalid for instance '0x18a80c8' of type 'WebKitWebView'

aperezdc:
Instead use `WebKitWebView::web-process-terminated`, which provides additional information about the reason of the process termination. The termination reason is used to format the messages and error pages, but other than that the behaviour of the updated code remains the same.

Fixes #3

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -333,16 +333,16 @@ on_create_web_view (CogLauncher *launcher,
             break;
 
         case WEBPROCESS_FAIL_EXIT:
-            cog_web_view_connect_web_process_crashed_exit_handler (web_view, EXIT_FAILURE);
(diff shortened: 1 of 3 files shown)


aperezdc:
@mcatanzaro: Thanks! I'll fix the details you mentioned before merging this PR. Regarding the extra indentation space for the variable names of the parameters, I'll fix it only for the parts that this patch touches — let's slowly try to change the rest of functions when they are modified to avoid introducing noise in this patch.

Issue #9, opened 2018-05-15
Last master fails to work with non-fdo


clopez:
Not sure if this was caused by 17cfde02f44279432c6db82f58cbfbe7b0064431 or it was a previous issue.

I'm testing with ae3e565a3c8d9f48c2bccc5c9ac7aa0fd624e50d and wpebackend-rdk/wayland-egl and cog is currently broken unless a "-P randomstring is passed".

Simply starting cog without parameters fails with:
```
root@qemux86-64:~# cog http://igalia.com
(cog:2550): Cog-DEBUG: platform_setup: Platform name: (null)
(cog:2550): Cog-DEBUG: platform_setup: Platform = (nil)

(cog:2550): Cog-CRITICAL **: cog_platform_get_view_backend: assertion 'platform != NULL' failed
**
Cog:ERROR:/usr/src/debug/cog/20180514-r0/git/cog.c:301:on_create_web_view: 'error' should not be NULL
Aborted
```

Starting it with a random invalid platform plugin name however works
```
root@qemux86-64:~# cog -P nonexistent http://igalia.com
(cog:2557): Cog-DEBUG: platform_setup: Platform name: nonexistent
(cog:2557): Cog-DEBUG: platform_setup: Platform plugin: libcogplatform-nonexistent.so

(cog:2557): Cog-WARNING **: Could not load: libcogplatform-nonexistent.so (possible cause: Socket operation on non-socket).

(cog:2557): Cog-DEBUG: Instantiating default WPE backend as fall-back.
xkbcommon: ERROR: couldn't find a Compose file for locale "C"
Cog-Message: <http://igalia.com/> Load started.
Cog-Message: <http://www.igalia.com/> Redirected.
Cog-Message: <https://www.igalia.com/> Redirected.
Cog-Message: <https://www.igalia.com/> Loading...
Cog-Message: <https://www.igalia.com/> Loaded successfully.
```

macpijan:
@clopez Just want to report similar issue when building from: https://github.com/Igalia/meta-webkit/commit/83b57008f4fe17b2f09c1f133d430b0f0a59efea
`wpebackend-fdo` was used in my case


```
cog  https://people.igalia.com/clopez/wkbug/fps-counter/raf_onload.html
(cog:528): Cog-CRITICAL **: cog_platform_get_view_backend: assertion 'platform != NULL' failed
**
[...]

PR #10, opened 2018-05-15
return FALSE early if platform_name is null in platform_setup() (#9)

clopez:
This fixes a bug where platform_name() was beeing returning TRUE when
no platform_name is defined. Causing it to assert on the calls to
on_create_web_view() -> cog_platform_get_view_backend()


clopez:
This fixes #9 for me

Issue #11, opened 2018-05-15
Using a platform plug-in should not need libWPEBackend-default.so 


aperezdc:
Currently, even if a platform plug-in is loaded (e.g. passing `--platform=fdo` in the command line), it is still needed to have around a `libWPEBackend-default.so` shared object, which typically is a symlink to the actual WPE backend being used.

Ideally, when using a Cog platform plug-in, `libWPEBackend-default.so` should not be needed at all — though of course it could be still be used when running without any platform plug-in.

clopez:
I confirm the issue. 
To get cog working with fdo I need to run on the live system:
```
ln -s /usr/lib/libWPEBackend-fdo-0.1.so.0 /usr/lib/libWPEBackend-default.so
```

And then run cog as
```
cog -P fdo http://igalia.com
```

If i try to rung without specifying a platform value causes it crashes  with
```
# cog http://igalia.com
(cog:895): Cog-DEBUG: platform_setup: Platform name: (null)
(cog:895): Cog-DEBUG: Instantiating default WPE backend as fall-back.

** (cog:895): CRITICAL **: WebKitWebViewBackend* webkit_web_view_backend_new(wpe_view_backend*, GDestroyNotify, gpointer): assertion 'backend' failed

(cog:895): Cog-ERROR **: Could not instantiate any WPE backend.
```

I wonder why platform is null? shouldn't default to fdo when cog has been built with fdo support?

aperezdc:
Like I mentioned in #12 already, I would rather go with the implementation proposed by @carlosgcampos in [WebKit#186841](https://bugs.webkit.org/show_bug.cgi?id=186841) and https://github.com/WebPlatformForEmbedded/WPEBackend/pull/18, which adds API to set the path to the WPE backend library, and passing it down to `WPEWebProcess`.

PR #12, opened 2018-05-15
Install a libWPEBackend-default.so symlink with fdo (#11)

clopez:
* When building with fdo support (-DCOG_PLATFORM_FDO=ON) create
   and install a libWPEBackend-default.so symlink pointing to
   libWPEBackend-fdo-0.1.so.0.

 * Also default to platform fdo is no other one is specified via
   command line argument.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -147,6 +147,13 @@ target_include_directories(cogplatform-fdo PUBLIC wayland ${COGPLATFORM_FDO_INCL
 target_compile_options(cogplatform-fdo PUBLIC ${COGPLATFORM_FDO_CFLAGS})
 target_link_libraries(cogplatform-fdo cogcore ${COGPLATFORM_FDO_LDFLAGS})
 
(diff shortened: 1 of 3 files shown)


zdobersek:
I would propose creating libWPEBackend-cog.so and installing the libWPEBackend-default.so link that links to that libWPEBackend-cog.so library.

Upon libWPEBackend-cog.so being loaded or queried for the first interface implementation via _wpe_loader_interface, the implementation would go out and search for the appropriate libWPEBackend library (e.g. libWPEBackend-fdo.so in case of COG_PLATFORM_FDO being enabled) and then load symbols from there.

aperezdc:
An alternative that I like better is what @carlosgcampos has implemented in [WebKit#186841](https://bugs.webkit.org/show_bug.cgi?id=186841) and https://github.com/WebPlatformForEmbedded/WPEBackend/pull/18 — that is, adding API to set the path to the WPE backend library, and passing it down to `WPEWebProcess`.

aperezdc:
We'll eventually have API to configure which WPE backend implementation to use, so let's not merge this.

PR #19, opened 2018-05-31
Wrap and submit the exported buffer forward instead of blitting

elima:
Since we render the exported buffer to the full size of the surface, we don't need to blit it.

This patch creates a wl_buffer that wraps the buffer resource and attach it to the wayland compositor surface for rendering, thus avoiding texture mapping and GLES altogether.

Tested in Intel/i965 and i.MX6 platforms, in both cases with a noticeable increase in performance.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -23,8 +23,6 @@
 #include <wayland-egl.h>
 #include <EGL/egl.h>
 #include <EGL/eglext.h>
-#include <GLES2/gl2.h>
-#include <GLES2/gl2ext.h>
 
 /* for mmap */
 #include <sys/mman.h>
@@ -90,7 +88,6 @@ static struct {
 static struct {
     struct egl_display *display;
     EGLContext context;
-    PFNGLEGLIMAGETARGETTEXTURE2DOESPROC glEglImageTargetTexture2D;
     PFNEGLCREATEIMAGEKHRPROC eglCreateImage;
     PFNEGLDESTROYIMAGEKHRPROC eglDestroyImage;
     EGLConfig egl_config;
@@ -110,12 +107,6 @@ static struct {
 
     bool is_fullscreen;
 } win_data;
-
-static struct gl_data {
-    GLuint program;
-    GLuint tex;
-    GLuint tex_loc;
-} gl_data = {0, };
 
 static struct {
     struct xkb_context* context;
@@ -142,13 +133,11 @@ static struct {
     WebKitWebView *view;
     struct wl_resource* current_buffer;
     EGLImageKHR image;
+    struct wl_buffer *buffer;
     struct wl_callback *frame_callback;
 
     double zoom_level;
 } wpe_view_data = {NULL, };
-
-
-static void draw (void);
 
 
 struct wl_event_source {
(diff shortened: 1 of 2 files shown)


platform/cog-platform-fdo.c
```diff
@@ -998,6 +987,11 @@ on_surface_frame (void *data, struct wl_callback *callback, uint32_t time)
         egl_data.eglDestroyImage (egl_data.display, wpe_view_data.image);
         wpe_view_data.image = NULL;
     }
+
+    if (wpe_view_data.buffer != NULL) {
+        wl_buffer_destroy (wpe_view_data.buffer);
+        wpe_view_data.buffer = NULL;
+    }
```
aperezdc: Please replace the whole `if`-statement with [g_clear_pointer()](https://developer.gnome.org/glib/unstable/glib-Memory-Allocation.html#g-clear-pointer), which is equivalent:

```c
g_clear_pointer (&wpe_view_data.buffer, wl_buffer_destroy);
```

Most of the code in Cog uses the `g_clear_*` and `g_auto*` macros, and I think once one gets used to them, the code is more concise without being less explicit or unclear.

aperezdc: Please replace this whole `if`-statement with [g_clear_pointer()](https://developer.gnome.org/glib/unstable/glib-Memory-Allocation.html#g-clear-pointer), which is equivalent:

```c
g_clear_pointer (&wpe_view_data.buffer, wl_buffer_destroy);
```

Most of the code in Cog uses the `g_clear_*` and `g_auto*` macros, and I think once one gets used to them, the code is more concise without being less explicit or unclear.

platform/cog-platform-fdo.c
```diff
@@ -1069,7 +1026,26 @@ on_export_buffer_resource (void* data, struct wl_resource* buffer_resource)
                                                    image_attrs);
     assert (wpe_view_data.image != NULL);
 
-    draw ();
+    static PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL
+        eglCreateWaylandBufferFromImageWL;
+    if (eglCreateWaylandBufferFromImageWL == NULL) {
+        eglCreateWaylandBufferFromImageWL =
+            eglGetProcAddress ("eglCreateWaylandBufferFromImageWL");
+        assert (eglCreateWaylandBufferFromImageWL != NULL);
```
aperezdc: Please replace the assertion with:

```c
g_assert_nonnull (eglCreateWaylandBufferFromImageWL);
```

Using `g_assert*` gets us better messages when the assertion checks fail.

platform/cog-platform-fdo.c
```diff
@@ -1069,7 +1026,26 @@ on_export_buffer_resource (void* data, struct wl_resource* buffer_resource)
...
+    }
+
+    wpe_view_data.buffer = eglCreateWaylandBufferFromImageWL (egl_data.display,
+                                                              wpe_view_data.image);
+    assert (wpe_view_data.buffer != NULL);
```
aperezdc: Ditto, please use `g_assert_nonnull()`.

elima:
@aperezdc, thanks for the review. I have addressed all your comments and pushed a new version.

Eventually I will have to sit down and glib-ify and organize platform-fdo.

aperezdc:
@elima Thanks for updating the PR, it's merged now.

BTW, no hurries in cranking up the GLib usage, the code is working pretty well as it is. We can always slowly introduce things like in this patch when possible :+1:

PR #29, opened 2018-06-11
Provide command line options for WebKitSettings properties

### LINKED ISSUE #27: WebView settings command-line pass-through
Something like WebKitGTK's MiniBrowser would be nice to have, along with a `--help-websettings` option.

aperezdc:
This adds code to introspect properties of a given `GObjectClass` and obtain a set of `GOptionEntry` instances which, when added to a `GOptionGroup`, will set the values of the properties from the parsed command line options.

Then, the above utility code is used to expose the properties from `WebKitSettings` in a generic way, instead of having to manually add a `GOptionEntry` and handling for each settings.

The following CLI options are replaced by the new functionality, which covers the corresponding settings:

| Old | New |
|-----|-----|
| `--webgl` | `--enable-webgl=1` |
| `-D` <br> `--dev-tools` | `--enable-developer-extras=1` |
| `-v` <br> `--log-console` | `--enable-write-console-messages-to-stdout=1` |

Closes #27

### DIFF
--- a/core/cog-utils.c
+++ b/core/cog-utils.c
@@ -259,18 +259,18 @@ cog_option_entries_from_class (GObjectClass *klass)
 
         // Pick only properties of basic types we know how to convert.
         const GType prop_type = G_PARAM_SPEC_VALUE_TYPE (prop);
+        const char *type_name = NULL;
         switch (prop_type) {
-            case G_TYPE_BOOLEAN:
+            case G_TYPE_BOOLEAN: type_name = "BOOL"; break;
             case G_TYPE_DOUBLE:
-            case G_TYPE_FLOAT:
+            case G_TYPE_FLOAT: type_name = "FLOAT"; break;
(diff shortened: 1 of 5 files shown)


core/cog-utils.c
```diff
@@ -73,3 +74,218 @@ cog_uri_guess_from_user_input (const char *uri_like,
...
+        return FALSE;
+    }
+
+    const GType prop_type = G_PARAM_SPEC_VALUE_TYPE (prop);
+    switch (prop_type) {
+        case G_TYPE_BOOLEAN: {
+            const gboolean prop_value = !(value &&
+                                          g_ascii_strcasecmp (value, "true") &&
+                                          strcmp (value, "1"));
+            g_printerr ("set %s=%s\n", option, prop_value ? "true" : "false");
```
philn: Is this a stray debug statement?

PR #32, opened 2018-06-12
Add CLI flags for setting cookie handling options

### LINKED ISSUE #31: Support configuring how and when to store cookies
By default WebKit already keeps cookies in transient storage around (otherwise many websites would not work), but it would be good to allow enabling persistent storage of the cookies. In general, it would be great to expose the functionality of [WebKitCookieManager](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitCookieManager.html):

- Enabling persistent storage.
- Setting the cookies storage format (plain text, SQLite).
- Configuring the path to the cookie jar (optional, automatically derive a path into `$XDG_CONFIG_HOME`).
- Pre-filling cookies.

aperezdc:
This adds a new `GOptionGroup` which will contain all the cookies-related command line flags.

Some reorganization was needed to allow early creation of the `WebKitWebContext` instance, which we need at the time when the option group is added to the `GApplication` in order to pass the context's `WebKitCookieManager` to the callback used to parse the command line.

Closes #31

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -341,9 +341,19 @@ on_create_web_view (CogLauncher *launcher,
 int
 main (int argc, char *argv[])
 {
+    // We need to set the program name early because constructing the
+    // CogLauncher instance will use g_get_prgname() to determine where
+    // to store the caches for Web content.
+    {
+        const char *dir_separator = strrchr (argv[0], G_DIR_SEPARATOR);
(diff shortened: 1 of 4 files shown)


aperezdc:
Regarding the compiler warnings, we should definitely fix #7 (which is an easy one) which would make sure CMake makes everybody use the same set of compiler flags. Thanks to you both for reviewing the PR!

PR #36, opened 2018-06-19
Support building cog on older versions of wayland

elima:
Specifically, cog needs these patches to build on a Yocto Krogoth system for i.MX6 with an slightly old version of the NXP BSP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,6 @@ endif()
 add_definitions(-DCOG_INSIDE_COG__=1)
 add_definitions(-DG_LOG_DOMAIN=\"Cog\")
 
-configure_file(core/cog-config.h.in cog-config.h @ONLY)
 include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 find_package(PkgConfig)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -35,11 +36,16 @@ endif()
 add_definitions(-DCOG_INSIDE_COG__=1)
 add_definitions(-DG_LOG_DOMAIN=\"Cog\")
 
+find_package(PkgConfig)
+
+pkg_check_modules(WAYLAND REQUIRED wayland-client)
+if (NOT (WAYLAND_VERSION VERSION_LESS 1.10.0))
+    set(WAYLAND_1_10_OR_GREATER ON)
+endif ()
```
aperezdc: Provided that Wayland is required only by the FDO platform module, it seems to me that it would be more correct to move this check inside the `if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)` conditional block (below, line ~132).

core/cog-launcher.c
```diff
@@ -14,6 +14,7 @@
 # include "cog-gtk-utils.h"
 #endif
 
+#include <stdlib.h>
```
aperezdc: This looks unrelated to the current PR. Could you please post this change in a separate PR if needed?

CMakeLists.txt
elima: Agree, but empirical trials suggested me that it should go above the `configure_file()` directive below, otherwise cog-config.h will not pick up the correct value of the cmake variable. Any idea on how to work around that?

core/cog-launcher.c
elima: Agree, it is not directly related with old versions of wayland. I will move it out.

BTW, without this it fails compilation on glibc 2.23 and gcc 5.3.

aperezdc: Indeed, we still want this change applied. Let's just do it in a separate PR — you can count on an immediate :+1: for merging it :wink:

CMakeLists.txt
aperezdc: The `configure_file()` call has to be moved to the end of the `CMakeLists.txt` file. That is probably the best location for it anyway, and having it at the end will avoid issues in the future: generating build configuration files should be the last thing done to avoid surprises.

elima: Ok, makes sense. I will do that change as part of the new version of this patch.

core/cog-launcher.c
elima: Great. PR on the way, thanks

PR #38, opened 2018-06-20
cmake: Make wayland-egl pkgconfig package an optional dependency for fdo

elima:
Some EGL drivers (e.g, i.MX6 Vivante) don't expose a separate pkgconfig
package for wayland-egl, so this patch makes this optional. This should
be perfectly safe because we are already requiring egl and
wayland-client.


CMakeLists.txt
```diff
@@ -142,11 +142,24 @@ if (NOT (WAYLAND_VERSION VERSION_LESS 1.10))
...
+set(COGPLATFORM_FDO_CFLAGS
+    ${COGPLATFORM_FDO_DEPS_CFLAGS_OTHER}
+    ${WAYLAND_CFLAGS_OTHER}
+    ${WAYLAND_EGL_CFLAGS_OTHER}
+)
+set(COGPLATFORM_FDO_LDFLAGS
+    ${COGPLATFORM_FDO_DEPS_LDFLAGS}
+    ${WAYLAND_LDFLAGS}
+    ${WAYLAND_EGL_LDFLAGS}
+)
```
aperezdc: This looks nicer after your reformatting :smiley:

elima:
An EGL driver that installs egl.pc will also install wayland-egl.pc if it supports wayland EGL extensions. Such is the case of Mesa. If the development package is not installed, but the system can still find egl.pc, it means it is provided by another vendor, and unlikely to work. `wayland-egl` is fully dependent on a specific implementation of EGL. I'm not sure this is a possible scenario, that's what I tried to convey in the commit message.

aperezdc:
Okay, let me give a concrete example. This situation is possible in Debian:

- Package `libegl1-mesa-dev` contains `wayland-egl.pc`
- Package `libwayland-dev` contains `wayland-client.pc`
- Package `libwayland-dev` does **not** depend on `libegl1-mesa-dev`.
   - File `wayland-client.pc` can be installed and `wayland-egl.pc` missing.

In the above situation, CMake will check `wayland-egl`, and it will **not** be found, but configuration will continue because the module is optional (there's no `REQUIRED` keyword for it). In this particular case (Debian), `egl.pc` is provided by package `libegl1-mesa-dev` as well, so things will work.

Other packagers may chose to have a separate `libwayland-egl-dev` package. For example the *meta-genivi* Yocto layer has a [separate package for libwayland-egl-dev](https://github.com/GENIVI/meta-genivi-dev/blob/e9c013951feb5c78519da04513ab79baf400264e/meta-raspberrypi-gdp/recipes-graphics/mesa/mesa.inc#L187) — in this case, CMake will complete the configuration without a `wayland-egl.pc` file being present: the source will compile because the EGL headers define the `wl_egl_*` functions, but linking will fail because the `-lwayland-egl` linker flag is pulled from the missing `waylang-egl.pc` file.

Anyway, I merged the PR because it seems more important to me to be able to build and link in configurations where `-lwayland-egl`/`wayland-egl.pc` are unneeded than detecting a corner case during the CMake configuration which is quite unlikely to happen :wink:

PR #41, opened 2018-06-21
[RFC] Support loading plugins from multiple shared modules

aperezdc:
Instead of loading a single `CogPlatform`, provide support for loading multiple modules, each of which can register any number of `CogPlugin` instances.

`CogPluginRegistry` keeps a mapping of names to `CogPlugin` instances, which can be registered manually with `cog_plugin_registry_add()`, or loaded from shared objects with `cog_plugin_registry_load()`. The latter resolves the `cog_module_initialize` symbol, and calls into it passing itself and allowing the module to register as many plugins as it wishes.

Plugin names cannot be reused, and trying to register a plugin with the same name again is not possible. It is possible to lookup a `CogPlugin` by name, which is used to check whether a `"platform"` plugin is registered. The FDO platform module is reworked to register the `"platform"` plugin.

This is a decent first pass at multiple plugin support, but there are a number of things which still need improving:

- The loadable module path has to be specified. It would be better to expand module names into `<prefix>/lib/cog-<apiversion>/libcog<name>.so`

- Module/plugin parameters are not yet passed, an empty string is always received. Some thinking is needed to figure out whether it's better to pass the parameters to the module, or to save them around to be passed to each plugin's `setup()` callback.

- So far only the `setup()`, `teardown()`, and `get_view_backend()` callbacks are defined. In the future it would be better to use GObject interfaces for different kinds of plug-ins, and allow querying by the implemented interfaces as well. This would remove the awkwardness of all plugins having a pointer for a `get_view_backend()` which only the `"platform"` plugins will ever use. While this is not much of a problem at the moment  because there aren't many callback pointer slots, as the interface grows it will be needed to tackle this.

- Currently the plugins have access to all of `CogLauncher` and the rest of libcogcore APIs. While this is very flexible, it can also be a  potential footgun and may need re-evaluation.

As a bonus, the implementation of `CogPlugin` / `CogPluginRegistry` also works when building against WebKitGTK+, while previously `CogPlatform` could only be used with WPE WebKit.

----

There's a couple of things to resolve before this would allow building modules from external source trees:

- [x] Installation of API headers for `libcogcore`.
- [x] Installation of a `.pc` file for `libcogcore`.
- [ ] Decide on a plug-in installation directory, and expose it in the `.pc` file, so externally built modules know where to install themselves .
- [ ] Versioned names for `libcogcore`, its headers, the modules directory, and the `.pc` file.
- [ ] Slate some plugin names (e.g. `"platform"`) as reserved. Probably even better use reverse domain notation for plugin, and reserve the whole `com.igalia.cog.*` namespace.
- [ ] Expansion of module names into their full paths.
- [ ] Passing of initialization parameters to modules/plugins.


aperezdc:
The `cogcore.pc` file was already being installed, since commit a35ad11. Ticked a box without writing any code!

aperezdc:
The API headers for `libcogcore` are installed since commit f035cc93. Another box ticket.

aperezdc:
FWIW, this will need to be rebased on top of #49, and add a `CogPluginRegistry` into `CogShell` to make all its users benefit from the same ability of loading plug-ins.

(Side note: Maybe we have to think a bit if there is some use-case in which plugin loading may be disabled, and if so allow creating instances of `CogShell` which do not make use of plugins.)

aperezdc:
Rebased on top of `master` (currently f79cae0983f8623a337c0047e20f461192d1f5ed).

aperezdc:
I will be reworking this in the future, but this PR has bitrotted for too long, so let's just close it.

PR #49, opened 2018-07-16
Rework CogLauncher to avoid forcing usage of GApplication

### LINKED ISSUE #46: CogLauncher should probably not be a GApplication
Right now, CogLauncher inherits from GApplication, which is convenient for Cog because it moves all the useful logic into libcogcore, simplifying the actual program and make it play well on Freedesktop.org systems.

However, in platforms where the features brought in by GApplication are not needed, this inheritance becomes a problem, making libcogcore less flexible to systems outside Freedesktop.org.

The solution would be making CogLauncher not inherit from GApplication, and implement the GApplication logic in cog.c instead.

aperezdc:
This splits the functionality from `CogLauncher` which is independent from `GApplication` into a new `CogShell` class. This way, programs which do not wish (or cannot) use `GApplication` can still easily make use of most of the facilities provided by `libcogcore`:

- A new `CogShell` class includes the functionality split from `CogLauncher`. Instead of the `CogView` name suggested by myself in #46, I think this name is better, in case at some point we support multiple Web views and/or have our own `CogView` class. 
- `CogLauncher` is rewritten so it makes use of `CogShell` instead.

Fixes #46 

----

Pending to address with this PR:

- Platform plug-ins should receive a `CogShell` instead of a `CogLauncher`.


aperezdc:
@elima My plan for platform plugins is making them a `CogShell`. Would it be okay for the FDO one to use the `GApplication` API to figure out whether an instance is available, and if not use some hardcoded application identifier instead of the one stored in the `GApplication`?

elima:
@aperezdc, changes look good. I'm ready to approve this pull request, but I want to clarify first the plan with platform plugins and CogShell. I don't fully understand what's the proposal above.
As far as I understood the new architecture, there will be different types of plugins (frame-source/sink, input-source/sink, ipc, etc). CogShell seems to essentially wrap a WebView. What exactly means _making platform plugins a CogShell_ ?

aperezdc:
@elima That comment was very badly written, I have edited it to clarify what I had in mind, please check it out. Basically it's what I have implemented in commit https://github.com/Igalia/cog/pull/49/commits/c2225a308fabe5a916fe5d006ee50bfe435c343d.

*Additionally*, after this is merged I want to update PR #41 to remove the usage of `CogPluginRegistry` from `cog.c` and put it inside `CogShell`. That way any application using `CogShell` will automatically be able to use plugins.

PR #50, opened 2018-07-16
fdo: Introduce error handling, cleanups, and GLib assertions

aperezdc:
This improves the FDO platform plug-in by:

- Making the `init_*` functions return errors, which get bubbled up from `cog_platform_setup()`. This includes more informative error strings, plus fetching and reporting EGL errors as well.

- The added error handling alone removes many assertions, which is quite a good improvement when it comes to knowing what went South during initialization. Also this means that other fallbacks can be used on failure (e.g. a Wayland compositor is not available), instead of  aborting the process miserably with an assertion message.

- The `clear_egl()`, `destroy_window()`, and `clear_input()` functions are made idempotent, and safe to use on partially initialized data. This allows using them to free resources at different stages of initialization in the event of a failure.

- Wherever possible, `g_clear_pointer()` is used for brevity and clarity.

While there are still many assertions remaining, most of them are quite unlikely of being ever hit, so this is a good stab at mostly fixing issue #1 — reviewing the remaining assertions later on would not hurt, though.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -1,13 +1,14 @@
 /*
  * cog-fdo-platform.c
  * Copyright (C) 2018 Eduardo Lima <elima@igalia.com>
+ * Copyright (C) 2018 Adrian Perez de Castro <aperez@igalia.com>
  *
  * Distributed under terms of the MIT license.
  */
 
 #include <cog.h>
 
-#include <assert.h>
+#include <errno.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <wpe/webkit.h>
@@ -357,7 +358,7 @@ registry_global (void               *data,
                                               name,
                                               &zxdg_shell_v6_interface,
                                               version);
-        assert (wl_data.xdg_shell != NULL);
+        g_assert_nonnull (wl_data.xdg_shell);
         zxdg_shell_v6_add_listener (wl_data.xdg_shell, &xdg_shell_listener, NULL);
     } else if (strcmp (interface,
                        zwp_fullscreen_shell_v1_interface.name) == 0) {
@@ -563,7 +564,7 @@ keyboard_on_enter (void *data,
                    struct wl_surface *surface,
                    struct wl_array *keys)
 {
-    assert (surface == win_data.wl_surface);
+    g_assert (surface == win_data.wl_surface);
     wl_data.keyboard.serial = serial;
 }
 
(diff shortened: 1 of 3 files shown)


platform/cog-platform-fdo.c
```diff
@@ -357,7 +358,7 @@ registry_global (void               *data,
                                               name,
                                               &zxdg_shell_v6_interface,
                                               version);
-        assert (wl_data.xdg_shell != NULL);
+        g_assert_nonnull (wl_data.xdg_shell);
```
mcatanzaro: The `g_assert_*` family of functions is only intended for use in unit tests. It prints the result in a way that would be useful when logging test failures. You should just use normal `g_assert` instead. That's the only form of assertion that should be used outside of your tests.

platform/cog-platform-fdo.c
```diff
@@ -1122,56 +1155,78 @@ init_egl (void)
     };
 
     EGLint num_configs;
-    if (! eglChooseConfig(egl_data.display,
-                          config_attribs,
-                          &egl_data.egl_config,
-                          1,
-                          &num_configs)) {
-        assert (!"Error choosing EGL config\n");
+    if (!eglChooseConfig(egl_data.display,
```
mcatanzaro: Missing space here after `eglChooseConfig`

platform/cog-platform-fdo.c
aperezdc: Why should only `g_assert` used in unit tests? The [documentation](https://developer.gnome.org/glib/stable/glib-Testing.html) does not mention anything about that. If something, it *encourages* using the extended `g_assert_*` functions to get better diagnostics:

> In addition to the traditional g_assert(), the test framework provides an extended set of assertions for comparisons: g_assert_cmpfloat(), g_assert_cmpint(), g_assert_cmpuint(), g_assert_cmphex(), g_assert_cmpstr(), and g_assert_cmpmem(). The advantage of these variants over plain g_assert() is that the assertion messages can be more elaborate, and include the values of the compared entities.

platform/cog-platform-fdo.c
aperezdc: Fixed in update, thanks!

platform/cog-platform-fdo.c
mcatanzaro: That's the documentation of the test framework. In unit tests, you would indeed want to prefer one of the more specific asserts, and avoid using `g_assert`. Outside tests, it's the opposite. Anyway, it's just a rule that GNOME projects use... I thought it was written down somewhere, maybe in the programming guidelines or someplace, but I'm not sure.

PR #51, opened 2018-07-17
CMake: Allow disabling the programs at build time

aperezdc:
This adds a new `COG_BUILD_PROGRAMS` option to CMake, enabled by default, which allows disabling building the `cog` and `cogctl` programs. This can be useful for packagers which are only interested in using the `libcogcore` library.


elima:
Merged

PR #53, opened 2018-08-14
core: Use "about:blank" instead of NULL as default URL

aperezdc:
Using `NULL` as default home URL was causing a critical warning to be logged right after creating the Web view if no valid URL is set with `cog_shell_set_home_uri()` beforehand, when requesting the URL to be loaded.

Also, when passing a `NULL` URL to `cog_shell_set_home_uri()` convert it into `"about:blank"` internally, guaranteeing that `cog_shell_get_home_uri()` always returns a valid pointer.

### DIFF
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -7,6 +7,10 @@
 
 #include "cog-shell.h"
 
+#ifndef COG_SHELL_DEFAULT_HOME_URI
+#define COG_SHELL_DEFAULT_HOME_URI "about:blank"
+#endif // !COG_SHELL_DEFAULT_HOME_URI
+
 
 typedef struct {
     char             *name;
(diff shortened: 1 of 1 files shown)


core/cog-shell.c
```diff
@@ -268,7 +272,7 @@ cog_shell_class_init (CogShellClass *klass)
         g_param_spec_string ("home-uri",
                              "Home URI",
                              "URI initially loaded by the Web view",
-                             NULL,
+                             COG_SHELL_DEFAULT_HOME_URI,
```
aperezdc: Actually after testing this a bit more, I am not very happy about it. Having a default value for `home-uri` and always loading the URI during `cog_shell_startup()` will add the URI to WebKit's “back-forward list”, which means there will be always one entry (the “no content” page, before loading `about:blank`), which breaks checking for `webkit_web_view_can_go_back()` — it always returns `TRUE` after creation of the Web view, and it should instead return `FALSE` if nothing has been explicitly loaded.

At this point I am leaning towards removing the `home-uri` functionality from `CogShell` and leaving that functionality to the application.

aperezdc:
After thinking a bit about this, I would rather remove the `home-uri` property. I'll be sending a different PR for that.

PR #54, opened 2018-08-14
core: Fix building with WebKitGTK+ after CogShell introduction

aperezdc:
Use `cog_launcher_get_shell()` + `cog_shell_get_web_view()` instead of the `cog_launcher_get_web_view()` function, which was removed.


aperezdc:
P.S: This wouldn't have gone unnoticed if we had CI. Hopefully I'll have something in that department in the coming weeks :wink:

PR #58, opened 2018-09-03
cmake: Bump libwpe requirement to version 0.2

philn:


aperezdc:
Thanks, this looks good to me! I would like to keep around a branch for of Cog that can be built with WPE WebKit 2.20.x before landing this. I will get back to this later today and make sure this PR gets merged.

PR #60, opened 2018-09-03
fdo: Explicitly set the backend implementation name

### LINKED ISSUE #11: Using a platform plug-in should not need libWPEBackend-default.so

aperezdc:
Fixes #11

### DIFF
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -11,6 +11,7 @@
 
 
 G_DEFINE_QUARK (COG_PLATFORM_EGL_ERROR, cog_platform_egl_error)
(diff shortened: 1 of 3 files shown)


elima:
Merged.

aperezdc:
@elima No problem at all. Thanks for taking the time for testing the patch, having to rebuild WPE WebKit *is* a chore after all!

PR #62, opened 2018-09-13
Make it possible to build against GLib 2.58 and newer

aperezdc:
Starting from 2.58.0 GLib already includes the definition needed to use `g_autoptr(GEnumClass)`.


aperezdc:
Cherry picked into the `cog-0.1` branch as well (d772990e623b532d35c43c077d59b3ccf329b386).

PR #68, opened 2018-11-27
fdo: Support maximizing the Wayland surface

aperezdc:
Maximizing the surface asks the compositor to let it occupy as much as it would allow, e.g. without covering other important UI elements. This is toggled with the `COG_PLATFORM_FDO_VIEW_MAXIMIZE` environment variable. Note if both fullscreening *and* maximization are requested (using their respective environment variables) then fullscreening is preferred.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -104,7 +104,15 @@ static struct {
     uint32_t height;
 
     bool is_fullscreen;
-} win_data;
+    bool is_maximized;
+} win_data = {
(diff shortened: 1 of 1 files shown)


aperezdc:
@elima The code style change was not intentional, so I will fix that.

aperezdc:
Updated, also I have filed #70 to improve code formatting because it would be nice to get all the files follow the same conventions everywhere.

aperezdc:
Merged in the `cog-0.1` branch as 03ef5cb09165fd16938f2d8cc6fc5b9f4af271ee.

PR #72, opened 2018-12-13
Cog fixes for iMX8M-evk

petegriffin:
With these two patches Cog launcher starts OK using wpewebkit 2.22.2 and backend-fdo on iMX8M-evk v4.9 based BSP release. I notice weston-simple-egl has a similar guard on egl_window resize and also calls wl_surface_commit when using zxdg_surface_v6.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -257,10 +257,11 @@ configure_surface_geometry (int32_t width, int32_t height)
 static void
 resize_window (void)
 {
-    wl_egl_window_resize (win_data.egl_window,
-                          win_data.width,
-                          win_data.height,
-                          0, 0);
+    if (win_data.egl_window)
+        wl_egl_window_resize (win_data.egl_window,
+			      win_data.width,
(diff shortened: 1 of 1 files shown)


platform/cog-platform-fdo.c
```diff
@@ -257,10 +257,11 @@ configure_surface_geometry (int32_t width, int32_t height)
...
 {
-    wl_egl_window_resize (win_data.egl_window,
-                          win_data.width,
-                          win_data.height,
-                          0, 0);
+    if (win_data.egl_window)
+        wl_egl_window_resize (win_data.egl_window,
+			      win_data.width,
+			      win_data.height,
+			      0, 0);
```
aperezdc: Now that you pointed this fix, it *does* seem plausible that a resize event can be triggered during the small time interval before the `.egl_window` member is initialized :scream:

aperezdc: I just got word from @zdobersek by chat that this also looks like a good fix to him :+1:

PR #76, opened 2019-01-11
core: When guess input URIs, directly use valid ones

### LINKED ISSUE #75: Silently fails to load URIs containing authentication credentials
Use-case: `cog -P fdo http://foo:password@example.com`
Expected behavior: Page displays
Current behavior: White page

The problem is that in `cog_uri_guess_from_user_input` we use `soup_uri_to_string()` which, as documented, strips the password out.

aperezdc:
Change the `cog_uri_guess_from_user_input()` function slightly, so if an input URI is valid and the scheme refers to a network protocol, use the passed string directly, decoded from the command line locale. This avoids converting the `SoupURI` resulting from parsing to a string, which is what caused passwords in URIs to be stripped out.

Fixes #75

### DIFF
--- a/core/cog-utils.c
+++ b/core/cog-utils.c
@@ -40,26 +40,41 @@ cog_uri_guess_from_user_input (const char *uri_like,
         utf8_uri_like = g_strdup (uri_like);
     }
 
+    // If the URI can be parsed do not try to guess whether the argument
+    // is a local file or whether a scheme should be added to it. This also
+    // covers the case of custom URI scheme handlers.
     g_autoptr(SoupURI) uri = soup_uri_new (utf8_uri_like);
     if (uri) {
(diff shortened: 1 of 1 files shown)


carlosgcampos:
g_file_new_for_commandline_arg() does resolve relative paths.

aperezdc:
@carlosgcampos True, but it still won't resolve `file:foo` and `file://foo` to `file://$(pwd)/foo` :upside_down_face:

philn:
Ping :)

aperezdc:
Updated. I tried to keep it as simple as possible, but there are still a few cases for which `GFile` is not enough. There are now a coupl of comments in the code explaining that.

philn:
I tested this new version, 👍

PR #78, opened 2019-02-07
Create Wayland event source after the WPE view backend

aperezdc:
Certain Wayland events cause `libwpe` functions to be used, many of which need a valid WPE view backend available. To avoid trying to use a `NULL` view backend, this moves the creation of the `GSource` which integrates Wayland event dispatch with the GLib main loop to be done always after the first view backend is created.

This fixes a crash which was possible to trigger in the short window of time between using `cog_platform_setup()` and `cog_platform_get_view_backend()` by producing using input (e.g. pointer, key, or touch events) during the time window.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -1083,10 +1083,6 @@ init_wayland (GError **error)
               wl_data.shell != NULL ||
               wl_data.fshell != NULL);
 
-    wl_data.event_src =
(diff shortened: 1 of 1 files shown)


aperezdc:
Cherry picked:

- As 2571ffa5c54817f314ff2098b7f0a4fb3bca7705 in the `cog-0.1` branch.
- As c3f893bbabe21100e6805a7efd0e90237031505e in the `cog-0.2` branch.

PR #83, opened 2019-03-11
Background color configuration support

philn:


cog.c
```diff
@@ -75,6 +76,9 @@ static GOptionEntry s_cli_options[] =
     { "config", 'C', 0, G_OPTION_ARG_FILENAME, &s_options.config_file,
         "Path to a configuration file",
         "PATH" },
+    { "bg-color", 'b', 0, G_OPTION_ARG_STRING, &s_options.background_color,
+        "WebView background color to use, by default 'white'. Colors can also be provided with HEX syntax.",
```
aperezdc: I would just write `"Background color, as a CSS name or in #RRGGBB or #RRGGBBAA hex syntax (default: white)"`. Because I think that is what `webkit_color_parse` accepts, right?

It is quite clear that the specified color will affect the background behind the Web content without needing to add the word “WebView” here, and also I would like to have the `#RRGGBB or #RRGGBBAA` bit to make it look a bit like an example and be more explicit about how the color has to be specified without having to guess.

cog.c
```diff
@@ -352,6 +356,23 @@ on_create_view (CogShell *shell, void *user_data G_GNUC_UNUSED)
...
+      has_valid_color = webkit_color_parse (&color, s_options.background_color);
+#endif
+#else
+      GdkRGBA color;
+      has_valid_color = gdk_rgba_parse (&color, s_options.background_color);
+#endif
+
+#if defined(webkit_web_view_set_background_color)
+      if (has_valid_color)
+        webkit_web_view_set_background_color (web_view, &color);
```
aperezdc: If the color cannot be parsed, it would be desirable to print an error and exit instead of ignoring the specified value. Could you please add an `else` block here for that?

philn:
Thanks for the review, PR is updated now :)

PR #84, opened 2019-03-18
Add an option to control permission requests

thiblahute:
This adds the following API function:

  * cog_launcher_add_web_permissions_option_entries

Currently only 2 values are accepted 'all' and 'none', by default
no permissions are granted.

Long term, we will be able to add a more complete synthax to let the
user decide what to accept and what not.

### DIFF
--- a/core/cog-launcher.c
+++ b/core/cog-launcher.c
@@ -22,6 +22,7 @@
 struct _CogLauncher {
     CogLauncherBase parent;
     CogShell       *shell;
+    gboolean       allow_all_requests;
 };
 
 G_DEFINE_TYPE (CogLauncher, cog_launcher, COG_LAUNCHER_BASE_TYPE)
@@ -67,6 +68,27 @@ on_action_open (G_GNUC_UNUSED GAction *action,
     g_return_if_fail (g_variant_is_of_type (param, G_VARIANT_TYPE_STRING));
     webkit_web_view_load_uri (cog_shell_get_web_view (launcher->shell),
                               g_variant_get_string (param, NULL));
+}
+
+static gboolean
+on_permission_request (G_GNUC_UNUSED WebKitWebView *web_view,
+                       WebKitPermissionRequest *request,
+                       CogLauncher *launcher)
+{
+    if (launcher->allow_all_requests)
+        webkit_permission_request_allow (request);
+    else
+        webkit_permission_request_deny (request);
+
+    return TRUE;
+}
+
+static void
+on_notify_web_view (CogShell *shell, GParamSpec * arg G_GNUC_UNUSED, CogLauncher *launcher)
+{
+    WebKitWebView* web_view = cog_shell_get_web_view (shell);
+
+    g_signal_connect (web_view, "permission-request", G_CALLBACK (on_permission_request), launcher);
 }
 
 static void
@@ -148,7 +170,10 @@ cog_launcher_dispose (GObject *object)
 {
     CogLauncher *launcher = COG_LAUNCHER (object);
 
-    g_clear_object (&launcher->shell);
+    if (launcher->shell) {
+        g_signal_handlers_disconnect_by_func (launcher->shell, G_CALLBACK (on_notify_web_view), NULL);
+        g_clear_object (&launcher->shell);
+    }
 
     G_OBJECT_CLASS (cog_launcher_parent_class)->dispose (object);
 }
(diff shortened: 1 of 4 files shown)


core/cog-launcher.c
```diff
@@ -22,6 +22,7 @@
 struct _CogLauncher {
     CogLauncherBase parent;
     CogShell       *shell;
+    gchar          *permissions;
```
aperezdc: I would use a boolean or enum value here, instead of keeping the string around.

core/cog-launcher.c
```diff
@@ -69,6 +70,25 @@ on_action_open (G_GNUC_UNUSED GAction *action,
                               g_variant_get_string (param, NULL));
 }
 
+static gboolean
+decidePermissionRequest (WebKitWebView *webView, WebKitPermissionRequest *request, CogLauncher *launcher)
+{
+    if (!g_strcmp0 (launcher->permissions, "all"))
```
aperezdc: Making this a boolean or enum as suggested above would avoid having to compare strings here.

core/cog-launcher.c
```diff
@@ -69,6 +70,25 @@ on_action_open (G_GNUC_UNUSED GAction *action,
...
+        webkit_permission_request_allow (request);
+    else
+        webkit_permission_request_deny (request);
+
+    return TRUE;
+}
+
+static void
+on_new_webview (CogShell *shell, GParamSpec * arg G_GNUC_UNUSED, CogLauncher *launcher)
```
aperezdc: This is handling the signal `notify::web-view`, so I would call this function `on_notify_web_view` instead.

core/cog-launcher.c
```diff
@@ -142,13 +162,23 @@ cog_launcher_activate (GApplication *application)
 #endif
 }
 
+static void
+cog_launcher_finalize (GObject *object)
+{
+    CogLauncher *launcher = COG_LAUNCHER (object);
+
+    g_free (launcher->permissions);
```
aperezdc: Using a boolean or enum value (as suggested above) would make the `g_free` call unnecessary. Well, the `cog_launcher_finalize` function would not be needed, either.

core/cog-launcher.c
```diff
@@ -346,7 +378,6 @@ cookie_manager_get_accept_policy (WebKitCookieManager *cookie_manager)
     return data.result;
 }
 
-
```
aperezdc: Please remove this bit from the patch, it's only adding noise to the diff.

core/cog-launcher.c
```diff
@@ -638,3 +669,51 @@ cog_launcher_add_web_cookies_option_entries (CogLauncher *launcher)
...
+    if (g_strcmp0 (value, "all") && g_strcmp0 (value, "none")) {
+        g_set_error (error,
+                     G_OPTION_ERROR,
+                     G_OPTION_ERROR_BAD_VALUE,
+                     "Invalid permission value '%s' (allowed values: ['none', 'all'])",
+                     value);
+        return FALSE;
+    }
+
+    launcher->permissions = g_strdup (value);
```
aperezdc: Going with the suggestions above, for example if the value is a boolean, this would be just:

```c
    launcher->permissions = g_strcmp0 (value, "all") == 0;
```

We would avoid the string copy, too :wink:

core/cog-shell.c
```diff
@@ -121,6 +119,7 @@ cog_shell_startup_base (CogShell *shell)
 
     g_signal_emit (shell, s_signals[CREATE_VIEW], 0, &priv->web_view);
     g_object_ref_sink (priv->web_view);
+    g_object_notify (G_OBJECT (shell), "web-view");
```
aperezdc: Good catch, this was missing! :tada:

core/cog-launcher.c
thiblahute: Long term we will need a proper data structure here.

aperezdc: That's absolutely right, bit for now we're keeping it as simple as possible.

core/cog-launcher.c
thiblahute: This line will make noise all the time... anyway.

core/cog-launcher.c
thiblahute: Yeah... and a string wasn't a proper data structure anyway, you were totally right ;)

PR #94, opened 2019-05-12
HiDPI support

### LINKED ISSUE #16: fdo/wayland: Blurry rendering on HiDPI displays
For comparison, Cog on the left and Epiphany (which uses WebKitGTK+) on the right:

webkit-wpe-gtk-hidpi-comparison

tokyovigilante:
Enables Wayland HiDPI support. Also scale pointer
events by scale factor to match display. Closes #16.
Requires libwpe 1.3.0 and WebKit built with this patch - 
https://bugs.webkit.org/attachment.cgi?id=369667&action=diff

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -40,6 +40,11 @@
 
 #define DEFAULT_ZOOM_STEP 0.1f
 
+typedef struct output_metrics {
+  struct wl_output *output;
+  int32_t name;
+  int32_t scale;
+} output_metrics;
 
 static struct {
     struct wl_display *display;
@@ -51,6 +56,11 @@ static struct {
     struct wl_shell *shell;
 
     struct wl_seat *seat;
+    struct output_metrics metrics[16];
+    struct {
+        int32_t scale;
+    } current_output;
+
     struct {
         struct wl_pointer *obj;
         int32_t x;
@@ -257,15 +267,20 @@ configure_surface_geometry (int32_t width, int32_t height)
 static void
 resize_window (void)
 {
+    int32_t pixel_width = win_data.width * wl_data.current_output.scale;
+    int32_t pixel_height = win_data.height * wl_data.current_output.scale;
+
     if (win_data.egl_window)
         wl_egl_window_resize (win_data.egl_window,
-			      win_data.width,
-			      win_data.height,
-			      0, 0);
+			                  pixel_width,
+			                  pixel_height,
+			                  0, 0);
 
     wpe_view_backend_dispatch_set_size (wpe_view_data.backend,
                                         win_data.width,
                                         win_data.height);
+    g_debug ("Resized EGL buffer to: (%u, %u) @%ix\n",
+            pixel_width, pixel_height, wl_data.current_output.scale);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
     .close = xdg_toplevel_on_close,
 };
 
+
+static void noop()
```
aperezdc: Please put the function type in its own line, with a space between the name and the opening paren for the parameter list. Also, I think that the comment inside the function is not needed because the name “noop” already implies that it does nothing: 

```c
static void
noop (void)
{
}
```

platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
...
+}
+
+static const struct wl_output_listener output_listener = {
+        .geometry = noop,
+        .mode = noop,
+        .done = noop,
+        .scale = output_handle_scale,
+};
+
+static void surface_handle_enter (void *data, struct wl_surface *surface, struct wl_output *output)
```
aperezdc: Same here, please put `static void` in its own line.

platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
...
+  // This space intentionally left blank
+}
+
+static void
+output_handle_scale (void *data,
+                     struct wl_output *output,
+                     int32_t factor)
+{
+    bool found = false;
+    for (int i = 0; i < 16; i++)
```
aperezdc: Instead of hardcoding `16` here, please use `G_N_ELEMENTS (wl_data.metrics)`.

platform/cog-platform-fdo.c
```diff
@@ -257,15 +267,20 @@ configure_surface_geometry (int32_t width, int32_t height)
...
-			      win_data.height,
+			      pixel_width,
+			      pixel_height,
 			      0, 0);
 
     wpe_view_backend_dispatch_set_size (wpe_view_data.backend,
                                         win_data.width,
                                         win_data.height);
+    printf ("Resized EGL buffer to: (%u, %u) @%ix\n",
+            pixel_width, pixel_height, wl_data.current_output.scale);
```
aperezdc: There are a few situations in which `printf` is being used where it would be much better to use [g_debug](https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-debug), or [g_message](https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-message) for things which need to be always logged.

(I am aware that the existing code is already using `printf`, but see #1—it is better  to avoid introducing new uses.)

platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
...
+        if (wl_data.metrics[i].output == output)
+        {
+            found = true;
+            wl_data.metrics[i].scale = factor;
+            break;
+        }
+    }
+    if (!found)
+    {
+        printf("Unknown output %p\n", output);
```
aperezdc: Please leave a space between the function name and the opening parenthesis for the argument list:

```c
printf ("Unknown output %p\n", output);
```

This needs to be fixed in a few places more, but I'll comment it here just once.

platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
...
+        .mode = noop,
+        .done = noop,
+        .scale = output_handle_scale,
+};
+
+static void surface_handle_enter (void *data, struct wl_surface *surface, struct wl_output *output)
+{
+    int32_t scale_factor = -1;
+
+    for (int i=0; i<16; i++)
```
aperezdc: Replace `16` with `G_N_ELEMENTS (wl_data.metrics)`, please.

platform/cog-platform-fdo.c
```diff
@@ -340,6 +355,68 @@ static const struct zxdg_toplevel_v6_listener xdg_toplevel_listener = {
...
+    {
+        if (wl_data.metrics[i].output == output)
+        {
```
aperezdc: This opening brace should have been in the previous line.

platform/cog-platform-fdo.c
```diff
@@ -380,6 +457,29 @@ registry_global (void               *data,
                                          name,
                                          &wl_seat_interface,
                                          version);
+    } else if (strcmp (interface, wl_output_interface.name) == 0) {
+        struct wl_output* output = wl_registry_bind (registry,
+                                             name,
+                                             &wl_output_interface,
+                                             version);
```
aperezdc: The indentation of the function arguments in the three last lines looks off. Could you please make sure only spaces are used for indentation?

platform/cog-platform-fdo.c
```diff
@@ -1081,6 +1182,8 @@ init_wayland (GError **error)
         return FALSE;
     }
 
+    bzero(&wl_data.metrics, sizeof(output_metrics) * 16);
```
aperezdc: Nowadays using `memset` is preferred over `bzero`, because it has been removed from the latest POSIX specs. Also, please use `G_N_ELEMENTS` instead of `16`.

aperezdc:
I think this PR is looking good, so I'll wait until Monday for merging—just in case anyone else wants to check it out and give feedback before that. Thanks a lot again, @tokyovigilante!

tokyovigilante:
No problem! Was easy enough to get the code into Cog once all the WPE plumbing was done. Is there any chance an upcoming WPE release will have the relevant HiDPI changes included?

aperezdc:
@tokyovigilante The 2.26.x stable series, and the 2.25.x development ones, will include all the changes, and we will be releasing 2.26.0 sometime in between the end of October and beginning of November. If you would really, *really* like to have it in before, luckily the changes are all backwards-compatible, so we could look into backporting the feature to be in WPE WebKit 2.24.3 and Cog 0.3.1

zdobersek:
Do we also need to apply wl_data.current_output.scale to coordinates for the touch events?

tokyovigilante:
> The 2.26.x stable series, and the 2.25.x development ones, will include all the changes, and we will be releasing 2.26.0 sometime in between the end of October and beginning of November. 

Cool, I'm not fussed personally as I'm just manually applying the patch to the 2.24 series, but it would be nice to have them filter downstream sooner rather than later as I'm eventually hoping to distribute an app based on WPE.

@zdobersek Almost certainly yes, good catch! I haven't implemented touch in my own app yet so missed that. Will update those events and re-push.

tokyovigilante:
Fixed and tested touch events, thanks!

tokyovigilante:
Thanks!

PR #100, opened 2019-06-13
CMake: Make custom commands create directories

### LINKED ISSUE #99: Build fails when configuring with CMake to use Makefiles
Steps to reproduce:

```
% cmake -G 'Unix Makefiles' . && make
...
[ 52%] Generating WaylandProtocols.dir/fullscreen-shell-unstable-v1-client.h
Could not open output file: No such file or directory
make[2]: *** [CMakeFiles/cogplatform-fdo.dir/build.make:74: WaylandProtocols.dir/fullscreen-shell-unstable-v1-client.h] Error 1
make[1]: *** [CMakeFiles/Makefile2:109: CMakeFiles/cogplatform-fdo.dir/all] Error 2
```

(Issue reported by @bertogg — thanks!)

aperezdc:
Ninja will create itself the directories needed by the output files for targets, but Make does not do that, nor does the CMake Makefiles generator. Therefore, this adds a call to CMake's `make_directory` tool to ensure that the output directory will always be available, regardless of the build system being used.

Fixes #99

### DIFF
--- a/cmake/FindWaylandProtocols.cmake
+++ b/cmake/FindWaylandProtocols.cmake
@@ -117,6 +117,7 @@ function(add_wayland_protocol _target _kind _protocol)
         add_custom_command(
             OUTPUT "${proto_code}"
             MAIN_DEPENDENCY "${proto_file}"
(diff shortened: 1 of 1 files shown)


aperezdc:
Merged to the `cog-0.3` branch as well (5d20793aec964765c4cece09112926939cb85d09).

PR #105, opened 2019-06-21
Fix assertions

JPEWdev:
Most of the asserts cleaned up in 619f8f3 ("Replace testing-related
assertions with g_assert()") were done incorrectly with an inverted
check; g_assert_nonnull(foo) should be replaced with g_assert(foo) and
g_assert_null(foo) should map to g_assert(!foo).


philn:
@csaavedra 😂

csaavedra:
OMG, sorry. :-)

PR #108, opened 2019-07-02
fix syntax error in Use device scale API only when available in libwpe

psaavedra:
fix syntax error introduce in `fdo: Use device scale API only when
available in libwpe` (git show 86ba95ba)

```
error: ... platform/cog-platform-fdo.c:478:5: error: expected '}' before 'else'
     else if (strcmp (interface, wl_output_interface.name) == 0) {
```


psaavedra:
Sorry, this one is incompleted. The proper solution is in https://github.com/Igalia/cog/pull/109

PR #109, opened 2019-07-02
fix syntax error in Use device scale API only when available in libwpe

psaavedra:
fix syntax error introduce in `fdo: Use device scale API only when
available in libwpe` (git show 86ba95ba)


aperezdc:
Note to self: we really need some CI to avoid this kind of “build with some configuration went untested and we noticed later” issues.

At any rate, thanks for the follow-up patch @psaavedra!

PR #111, opened 2019-07-03
cog: Add --device-scale command line option

aperezdc:
Add a `--device-scale` command line option which sets the initial scaling factor for the output device. This is different from zooming the Web content with `--scale`, and is directly related to the pixel density of the display, where a scaling factor of 1.0 corresponds to the traditional the de-facto standard 96 DPI pixel density used by most computer displays. If the DPI for a display is known, it can be divided by 96 to find out a suitable device scaling factor for it.

Note that this sets only an initial value for the output device scale: a platform plug-in can change this factor later on by itself. For example the FDO platform plug-in will update it whenever the Wayland compositor reports a change of output device.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -15,6 +15,12 @@
 # include "cog-platform.h"
 #endif
 
+#if defined(WPE_CHECK_VERSION) && WPE_CHECK_VERSION(1, 3, 0)
+# define HAVE_DEVICE_SCALING 1
+#else
+# define HAVE_DEVICE_SCALING 0
+#endif /* WPE_CHECK_VERSION */
+
 
 enum webprocess_fail_action {
     WEBPROCESS_FAIL_UNKNOWN = 0,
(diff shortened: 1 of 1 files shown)


cog.c
```diff
@@ -345,7 +358,12 @@ on_create_view (CogShell *shell, void *user_data G_GNUC_UNUSED)
     // must have succeeded in providing a WebKitWebViewBackend* instance.
     if (!view_backend)
         g_error ("Could not instantiate any WPE backend.");
-#endif
+
+    struct wpe_view_backend* wpe_backend =
+        webkit_web_view_backend_get_wpe_backend (view_backend);
+    wpe_view_backend_dispatch_set_device_scale_factor (wpe_backend,
```
carlosgcampos: This should be guarded inside a #if HAVE_DEVICE_SCALING block, no?

aperezdc: Good catch, thanks! I'll add the guards.

PR #112, opened 2019-07-03
pointer_on_axis: fix the axis-scrolling issues

zdobersek:
The pointer_on_axis callback can receive a floating-point value that
represents the current vector distance of movement along that axis.
Current libwpe API cannot accomodate that value, so instead we should
only take the sign of that value and accordingly pass -1 or 1 as the
scrolling value for the axis event.


aperezdc:
Merged into `cog-0.3` as commit b5d42a0c4e6bba406198f46039a0a2fdd3589470 because it seems like a good idea, and it applied cleanly.

Issue #110, opened 2019-07-02
CI to activate a minimum set of test


psaavedra:
... SSIA

To avoid trivial errors like this:

https://github.com/Igalia/cog/pull/108
https://github.com/Igalia/cog/pull/109

aperezdc:
We basic CI now that #119 has been merged.

@psaavedra Thanks for your help with this issue :muscle:

PR #113, opened 2019-07-05
ci: Created .gitlab-ci.yml file (fix: #110)

psaavedra:
... implemented a simple smoke test unit to test the build process

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,21 @@
+before_script:
+  - ulimit -n 4096
+  - rm -rf ~/ci-cog
+  - mkdir -p ~/ci-cog
+  - cd ~/ci-cog
+  - git clone $CI_REPOSITORY_URL
+  - cd cog
+  - git checkout $CI_COMMIT_REF_NAME
+
+cog-meta-perf-browser-warrior-wandboard-mesa-wpe-weston:
+  tags:
+    - docker
+  script:
+    - wget -O ~/toolchain.sh https://wk-contrib.igalia.com/yocto/meta-perf-browser/browsers/nightly/sdk/wandboard-mesa/browsers-glibc-x86_64-core-image-weston-wpe-armv7at2hf-neon-wandboard-mesa-toolchain-1.0.sh
+    - chmod +x ~/toolchain.sh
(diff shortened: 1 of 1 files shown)


psaavedra:
Still WIP

psaavedra:
> Still WIP

not anymore

.gitlab-ci.yml
```diff
@@ -0,0 +1,21 @@
...
+  - cd ~/ci-cog
+  - git clone $CI_REPOSITORY_URL
+  - cd cog
+  - git checkout $CI_COMMIT_REF_NAME
+
+cog-meta-perf-browser-warrior-wandboard-mesa-wpe-weston:
+  tags:
+    - docker
+  script:
+    - wget -O ~/toolchain.sh https://wk-contrib.igalia.com/yocto/meta-perf-browser/browsers/nightly/sdk/wandboard-mesa/browsers-glibc-x86_64-core-image-weston-wpe-armv7at2hf-neon-wandboard-mesa-toolchain-1.0.sh
```
aperezdc: As far as I can see those commands should run just fine on other GNU/Linux systems and nothing special is required. Why not putting those to run on Travis-CI, for example, or some other CI system which can run on GitHub PRs?

aperezdc: It just took me a few minutes to write down a basic `.travis.yml` using this toolchain, see PR #119 (and note how the CI ran on that PR already).

psaavedra:
Just for curiosity sake. Is Travis CI free of charge?

aperezdc:
> Just for curiosity sake. Is Travis CI free of charge?

Travis-CI is free (as in “free beer”) for Open Source projects hosted in publicly visible repositories. There is an “Enterprise” service with additional features which is paid.

[CircleCI](https://circleci.com) also has a free tier which supports using custom Docker images for the build environment, and supports more complex build pipelines. Travis is very easy to configure so that's why I tried it first, so I think we can use it for now because the features are enough for what we need at the moment—we can switch in the future when Travis is not enough.

aperezdc:
Merged #119 instead of this Anyway thanks a lot to @psaavedra for making the SDK available and making this PR, which have been very useful :muscle:

psaavedra:
I see. Good to know it. Then, it makes much more sense to do it with Travis
CI in this case so far.

BTW, thanks for translate the Gitlab CI file to Travis.

PR #114, opened 2019-07-23
enable -Wall option when supported

charlie-ht:
This also fixes some warnings that existed. The only tricky one was to
do with g_autoptr. I'm not sure why, but I had to hoist and initialize
this further up the scope to stop GCC complaining about it being
potentially uninitialized in _GLIB_AUTOPTR_CLEAR_FUNC_NAME.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,8 +82,13 @@ else ()
     endif ()
 endif ()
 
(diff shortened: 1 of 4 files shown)


aperezdc:
Also merged to the `cog-0.3` branch as commit cbcab37ff4c73c8f574550fd71533145ee934868.

PR #116, opened 2019-07-26
Fixes and improvements related to project versioning

### LINKED ISSUE #115: Project version in CMake is outdated
Currently hardcoded to 0.1.0. Should be 0.3.2 ? Would be nice to have an indication at runtime that we run a git snapshot or not, using `cog --version`.

aperezdc:
This bumps the project version to something newer than the latest release, to make the `master` be “newer”. The other change adds logic in CMake to detect whether a build is done from a Git checkout, and in that case it adds `+git-<commit-id>`, optionally followed by `-dirty]` if there are uncommitted changes in the checkout, to the output of `cog --version`. Note that CMake will also report the 

CMake will also print the Git information during configuration: 

```
% cmake .
-- The C compiler identification is GNU 9.1.0
-- Check for working C compiler: /usr/lib/ccache/bin/cc
-- Check for working C compiler: /usr/lib/ccache/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: /usr/bin/git (found version "2.22.0")
-- Source tree revision: 0.3.2+git-32baa3e-dirty
...
% make
...
% ./cog --version
0.3.2+git-32baa3e
%
```

Closes #115


aperezdc:
Merged the part that picks the information from Git into the `cog-0.3` branch as commit 19c0ad64cc27ee6d28d06c329ec9c314c0093450.

PR #120, opened 2019-08-02
CI: Cache toolchain instead of always downloading it

aperezdc:
This instructs Travis-CI to cache the contents of the `~/toolchain/` directory.  The `install` step is modified to use of `curl --time-cond` passing `~/toolchain/.installed` as the time reference, which makes curl use the `If-Modified-Since` HTTP header to avoid re-downloading the toolchain installer if the already installed toolchain (from the cache) is as new as the installer file hosted by the server.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,22 @@ addons:
       - cmake
       - ninja-build
 
+cache:
+  directories:
+    - $HOME/toolchain
+
 install:
-  - 'wget -O ~/toolchain.sh "${TOOLCHAINS_URL}/${TOOLCHAIN_NAME}"'
-  - chmod +x ~/toolchain.sh && ~/toolchain.sh -d ~/toolchain -y
+  - |-
+    : 'Downloading toolchain'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -16,9 +16,26 @@ addons:
...
+  directories:
+    - $HOME/downloads
+    - $HOME/toolchain
+
 install:
-  - 'wget -O ~/toolchain.sh "${TOOLCHAINS_URL}/${TOOLCHAIN_NAME}"'
-  - chmod +x ~/toolchain.sh && ~/toolchain.sh -d ~/toolchain -y
+  - |-
+    mkdir -p ~/downloads
+    wget --tries=3 -N -O ~/downloads/toolchain.sh "${TOOLCHAINS_URL}/${TOOLCHAIN_NAME}"
```
aperezdc: This doesn't work, I just realized that `wget` prints

```
WARNING: timestamping does nothing in combination with -O. See the manual
```

Using `curl --time-cond` should work, so I'll update the PR to use that instead.

aperezdc: This works just perfect. I tried triggering rebuilds a few times with the cache empty, and with data in the cache. A build using a cached toolchain is around 25s faster than a build that downloads and installs it.

PR #121, opened 2019-08-05
cog: New runtime CLI option for WebExtensions directory configuration

### LINKED ISSUE #91: Support loading Web process extensions
It would be interesting for quick prototyping/testing to have a CLI option to specify a directory from which to load [Web Extensions](https://webkitgtk.org/reference/webkit2gtk/stable/ch02.html), e.g. `--web-extensions-dir=PATH`.

philn:
Fixes #91

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -54,6 +54,7 @@ static struct {
         char *action_name;
         enum webprocess_fail_action action_id;
     } on_failure;
+    char *web_extensions_dir;
 } s_options = {
     .scale_factor = 1.0,
 #if HAVE_DEVICE_SCALING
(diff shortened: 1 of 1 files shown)


cog.c
```diff
@@ -277,6 +281,10 @@ on_handle_local_options (GApplication *application,
         }
     }
 
+    if (s_options.web_extensions_dir != NULL) {
+        cog_shell_set_web_extensions_dir (shell, s_options.web_extensions_dir);
```
aperezdc: I think this should use the existing `cog_shell_get_web_context()` accessor instead:

```c
webkit_web_context_set_web_extensions_directory (cog_shell_get_web_context (shell),
                                                 s_options.web_extensions_dir);
```

This way we avoid adding unneeded public API to the `CogShell` class.

aperezdc:
Cherry-picked to the `cog-0.3` branch as commit `4c836f27695443b37b5a0301f59d6dc8f80d3e1f`).

PR #122, opened 2019-08-16
Handle SIGINT, SIGTERM signals gracefully

zdobersek:
Catch the SIGINT and SIGTERM signals through g_unix_signal_add() and
cause the GApplication to quit, enabling the application to properly
clean up and dispose all of the resources.

### DIFF
--- a/core/cog-launcher.c
+++ b/core/cog-launcher.c
@@ -186,13 +186,8 @@ cog_launcher_dispose (GObject *object)
         g_clear_object (&launcher->shell);
     }
 
-    if (launcher->sigint_source > 0)
-        g_source_remove (launcher->sigint_source);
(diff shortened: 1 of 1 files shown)


core/cog-launcher.c
```diff
@@ -175,6 +186,14 @@ cog_launcher_dispose (GObject *object)
         g_clear_object (&launcher->shell);
     }
 
+    if (launcher->sigint_source > 0)
+        g_source_remove (launcher->sigint_source);
+    launcher->sigint_source = 0;
```
aperezdc: These three lines could be simplified using `g_clear_handle_id()` as follows:

```c
g_clear_handle_id (&launcher->sigint_source, g_source_remove);
```

(Same for `sigterm_source` below.)

PR #136, opened 2019-08-25
drm-experimental platform updates

zdobersek:
Primarily initial input support, along with some misc stuff.

### DIFF
--- a/platform/cog-platform-drm.c
+++ b/platform/cog-platform-drm.c
@@ -107,11 +107,31 @@ clear_drm ()
 static gboolean
 init_drm ()
 {
-    drm_data.fd = open ("/dev/dri/card0", O_RDWR);
-    if (drm_data.fd < 0)
+    drmDevicePtr devices[64];
+    memset(devices, 0, sizeof(drmDevicePtr) * 64);
+
+    int num_devices = drmGetDevices2 (0, devices, 64);
+    if (num_devices < 0)
         return FALSE;
 
-    drmModeRes *resources = drmModeGetResources (drm_data.fd);
+    drmModeRes *resources = NULL;
+    for (int i = 0; i < num_devices; ++i) {
+        drmDevicePtr device = devices[i];
+        if (!(device->available_nodes & (1 << DRM_NODE_PRIMARY)))
+            continue;
+
+        drm_data.fd = open (device->nodes[DRM_NODE_PRIMARY], O_RDWR);
+        if (drm_data.fd < 0)
+            continue;
+
+        resources = drmModeGetResources (drm_data.fd);
+        if (resources)
+            break;
+
+        close (drm_data.fd);
+        drm_data.fd = -1;
+    }
+
     if (!resources)
         return FALSE;
 
--- a/platform/cog-platform-drm.c
+++ b/platform/cog-platform-drm.c
@@ -289,15 +289,13 @@ input_handle_key_event (struct libinput_event_keyboard *key_event)
     uint32_t keysym = xkb_state_key_get_one_sym (state, key);
     uint32_t unicode = xkb_state_key_get_utf32 (state, key);
 
-    struct wpe_input_keyboard_event event;
-    event.time = libinput_event_keyboard_get_time (key_event);
-    event.key_code = keysym;
-    event.hardware_key_code = unicode;
-
     enum libinput_key_state key_state = libinput_event_keyboard_get_key_state (key_event);
-    event.pressed = (key_state == LIBINPUT_KEY_STATE_PRESSED);
-
-    event.modifiers = 0;
+    struct wpe_input_keyboard_event event = {
+            .time = libinput_event_keyboard_get_time (key_event),
+            .key_code = keysym,
+            .hardware_key_code = unicode,
+            .pressed = (key_state == LIBINPUT_KEY_STATE_PRESSED),
+        };
 
     wpe_view_backend_dispatch_keyboard_event (wpe_view_data.backend, &event);
 }
@@ -319,13 +317,13 @@ input_handle_touch_event (enum libinput_event_type touch_type, struct libinput_e
             event_type = wpe_input_touch_event_type_motion;
             break;
         case LIBINPUT_EVENT_TOUCH_FRAME: {
-            struct wpe_input_touch_event event;
-            event.touchpoints = input_data.touch_points;
-            event.touchpoints_length = 10;
-            event.type = input_data.last_touch_type;
-            event.id = input_data.last_touch_id;
-            event.time = time;
-            event.modifiers = 0;
+            struct wpe_input_touch_event event = {
+                    .touchpoints = input_data.touch_points,
+                    .touchpoints_length = 10,
+                    .type = input_data.last_touch_type,
+                    .id = input_data.last_touch_id,
+                    .time = time,
+                };
 
             wpe_view_backend_dispatch_touch_event (wpe_view_data.backend, &event);
 
--- a/platform/cog-platform-drm.c
+++ b/platform/cog-platform-drm.c
@@ -3,6 +3,8 @@
 #include <assert.h>
 #include <fcntl.h>
 #include <gbm.h>
+#include <libinput.h>
+#include <libudev.h>
 #include <wpe/fdo.h>
 #include <wpe/fdo-egl.h>
 #include <xf86drm.h>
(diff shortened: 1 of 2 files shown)


platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
 }
 
 
+static void
+input_handle_key_event (struct libinput_event_keyboard *key_event)
+{
+    struct wpe_input_xkb_context *default_context = wpe_input_xkb_context_get_default ();
+    struct xkb_state *state = wpe_input_xkb_context_get_state (default_context);
+
+    uint32_t key = libinput_event_keyboard_get_key (key_event) + 8;
```
aperezdc: Why do we need a `+ 8` here? I am sure it's needed to make things work, but it would be good to add a comment explaining why, or a `TODO: Fix $FOO` if this is a temporary hack to get things working without delving into more complexity right now.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+    uint32_t keysym = xkb_state_key_get_one_sym (state, key);
+    uint32_t unicode = xkb_state_key_get_utf32 (state, key);
+
+    struct wpe_input_keyboard_event event;
```
aperezdc: I would rather use C99 designated initializers here, as it guarantees that members not listed explicitly for initialization get filled with zeros:

```c
enum libinput_key_state key_state = libinput_event_keyboard_get_key_state (key_event);

struct wpe_input_keyboard_event event = {
    .time = libinput_event_keyboard_get_time (key_event),
    .key_code = keysym,
    .hardware_key_code = unicode,
    .pressed = (key_state == LIBINPUT_KEY_STATE_PRESSED),
};
```

This way we don't risk having member initialized with garbage in case the layout of the `struct` changes in the future.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+    event.time = libinput_event_keyboard_get_time (key_event);
+    event.key_code = keysym;
+    event.hardware_key_code = unicode;
```
aperezdc: Also, this seems odd... Hardware key codes almost never are Unicode code points, aren't they?

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+}
+
+static void
+input_handle_touch_event (enum libinput_event_type touch_type, struct libinput_event_touch *touch_event)
+{
+    uint32_t time = libinput_event_touch_get_time (touch_event);
+
+    enum wpe_input_touch_event_type event_type = wpe_input_touch_event_type_null;
+    switch (touch_type) {
+    case LIBINPUT_EVENT_TOUCH_DOWN:
```
aperezdc: Please add one level of indentation to the `case` labels.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+        event_type = wpe_input_touch_event_type_down;
+        break;
+    case LIBINPUT_EVENT_TOUCH_UP:
+        event_type = wpe_input_touch_event_type_up;
+        break;
+    case LIBINPUT_EVENT_TOUCH_MOTION:
+        event_type = wpe_input_touch_event_type_motion;
+        break;
+    case LIBINPUT_EVENT_TOUCH_FRAME:
+    {
```
aperezdc: This opening brace should be in the previous line.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+        struct wpe_input_touch_event event;
+        event.touchpoints = input_data.touch_points;
+        event.touchpoints_length = 10;
+        event.type = input_data.last_touch_type;
+        event.id = input_data.last_touch_id;
+        event.time = time;
+        event.modifiers = 0;
```
aperezdc: Same as above, I would rather use C99 designated initializers.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+
+        wpe_view_backend_dispatch_touch_event (wpe_view_data.backend, &event);
+
+        for (int i = 0; i < 10; ++i) {
```
aperezdc: This hardcoded array length here is bugging me. We should add this somewhere:

```c
#define N_ELEMENTS(arr)  ((sizeof (arr) / sizeof (arr[0]))
```

and then use `i < N_ELEMENTS (input_data.touch_points)` as the loop condition.

aperezdc: Actually, we are already using GLib which has [G_N_ELEMENTS](https://developer.gnome.org/glib/stable/glib-Standard-Macros.html#G-N-ELEMENTS:CAPS)—please use that.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+
+        return;
+    }
+    default:
+        g_assert_not_reached ();
+        return;
+    }
+
+    int id = libinput_event_touch_get_seat_slot (touch_event);
+    if (id < 0 || id >= 10)
```
aperezdc: Same here: `id >= G_N_ELEMENTS (input_data.touch_points)`

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+        || touch_type == LIBINPUT_EVENT_TOUCH_MOTION) {
+        touch_point->x = libinput_event_touch_get_x_transformed (touch_event,
+                                                                 input_data.input_width);
+        touch_point->y = libinput_event_touch_get_y_transformed (touch_event,
+                                                                 input_data.input_height);
+    }
+}
+
+static void
+input_process_events ()
```
aperezdc: This being plain C, there is a `void` missing: 

```c
static void
input_process_events (void)
{
}
```

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+}
+
+static void
+input_interface_close_restricted (int fd, void *user_data)
+{
+    close (fd);
+}
+
+static void
+clear_input ()
```
aperezdc: Same here: `void` missing inside the argument list.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+{
+    g_clear_pointer (&input_data.libinput, libinput_unref);
+    g_clear_pointer (&input_data.udev, udev_unref);
+}
+
+static gboolean
+init_input ()
```
aperezdc: Here, too: `void` missing inside the argument list.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+{
+    static struct libinput_interface interface = {
+            input_interface_open_restricted,
```
aperezdc: This indentation here looks a bit wonky, please reformat. Consider using designated initializers, too.

platform/cog-platform-drm.c
```diff
@@ -236,6 +279,180 @@ init_egl ()
...
+        return FALSE;
+
+    int ret = libinput_udev_assign_seat (input_data.libinput, "seat0");
+    if (ret)
+        return FALSE;
+
+    input_data.input_width = drm_data.mode->hdisplay;
+    input_data.input_height = drm_data.mode->vdisplay;
+
+    for (int i = 0; i < 10; ++i) {
```
aperezdc: Avoid hardcoded values: `i < G_N_ELEMENTS (input_data.touch_points)`

platform/cog-platform-drm.c
```diff
@@ -263,53 +480,108 @@ drm_source_dispatch (GSource* base, GSourceFunc callback, gpointer user_data)
...
-    struct drm_source* source = (struct drm_source *) glib_data.source;
-    memset (&source->event_context, 0, sizeof (drmEventContext));
-    source->event_context.version = DRM_EVENT_CONTEXT_VERSION;
-    source->event_context.page_flip_handler = drm_page_flip_handler;
-
-    source->pfd.fd = drm_data.fd;
-    source->pfd.events = G_IO_IN | G_IO_ERR | G_IO_HUP;
-    source->pfd.revents = 0;
-    g_source_add_poll (glib_data.source, &source->pfd);
+    static GSourceFuncs drm_source_funcs = {
```
aperezdc: These two `struct` initializations have wonky indentation as well.

platform/cog-platform-drm.c
zdobersek: I don't remember exactly how this came to be, but the same is done in the -fdo plugin and elsewhere.

platform/cog-platform-drm.c
aperezdc: Wow, interesting, thanks for adding the comment—it will prevent us from thinking that it may be a good idea to remove it when going over this code at some point in the future.

PR #138, opened 2019-08-27
Add a runtime option to ignore TLS certificate errors

philn:
By default this option is disabled and an error page will be displayed when the
page load fails due to a TLS error. However, by enabling this option the page
load can resume and thus the TLS error is ignored. Use this option carefully and
consider the security implications.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -53,11 +53,13 @@ static struct {
         enum webprocess_fail_action action_id;
     } on_failure;
     char *web_extensions_dir;
+    gboolean ignore_tls_errors;
 } s_options = {
     .scale_factor = 1.0,
 #if HAVE_DEVICE_SCALING
     .device_scale_factor = 1.0,
 #endif // HAVE_DEVICE_SCALING
+    .ignore_tls_errors = FALSE,
 };
 
 
@@ -103,6 +105,9 @@ static GOptionEntry s_cli_options[] =
     { "web-extensions-dir", '\0', 0, G_OPTION_ARG_STRING, &s_options.web_extensions_dir,
       "Load Web Extensions from given directory.",
       "PATH"},
+    { "ignore-tls-errors", '\0', 0, G_OPTION_ARG_NONE, &s_options.ignore_tls_errors,
+      "Ignore TLS erros (default: disabled)", NULL
+    },
     { G_OPTION_REMAINING, '\0', 0, G_OPTION_ARG_FILENAME_ARRAY, &s_options.arguments,
         "", "[URL]" },
     { NULL }
@@ -283,6 +288,8 @@ on_handle_local_options (GApplication *application,
         webkit_web_context_set_web_extensions_directory (cog_shell_get_web_context (shell),
                                                          s_options.web_extensions_dir);
     }
+
+    g_object_set (shell, "ignore-tls-errors", s_options.ignore_tls_errors, NULL);
 
     return -1;  /* Continue startup. */
 }
--- a/core/cog-launcher.c
+++ b/core/cog-launcher.c
@@ -117,12 +117,38 @@ on_permission_request (G_GNUC_UNUSED WebKitWebView *web_view,
     return TRUE;
 }
 
+static gboolean
+on_load_failed_with_tls_errors (WebKitWebView       *web_view,
+                                char                *failing_uri,
+                                GTlsCertificate     *certificate,
+                                GTlsCertificateFlags errors,
+                                void                *user_data)
+{
+    CogShell *shell = COG_SHELL (user_data);
+    WebKitWebContext *context = webkit_web_view_get_context (web_view);
+    gboolean ignore_tls_errors = FALSE;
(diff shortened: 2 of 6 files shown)


philn:
Well, I'm not sure this should get in now... Video hosted on websites with broken TLS certs fail to load, even with this option enabled.

core/cog-launcher.c
```diff
@@ -117,12 +117,38 @@ on_permission_request (G_GNUC_UNUSED WebKitWebView *web_view,
...
+on_load_failed_with_tls_errors (WebKitWebView       *web_view,
+                                char                *failing_uri,
+                                GTlsCertificate     *certificate,
+                                GTlsCertificateFlags errors,
+                                void                *user_data)
+{
+    CogShell *shell = COG_SHELL (user_data);
+    WebKitWebContext *context = webkit_web_view_get_context (web_view);
+    gboolean ignore_tls_errors = FALSE;
+    g_object_get (shell, "ignore-tls-errors", &ignore_tls_errors, NULL);
```
aperezdc: Is there any reason to avoid using `cog_shell_get_ignore_tls_errors()` here?

core/cog-launcher.c
```diff
@@ -117,12 +117,38 @@ on_permission_request (G_GNUC_UNUSED WebKitWebView *web_view,
...
+
+    if (ignore_tls_errors) {
+        SoupURI *soup_uri = soup_uri_new (failing_uri);
```
aperezdc: Please use `g_autoptr` here:

```c
g_autoptr(SoupURI) soup_uri = soup_uri_new (failing_uri);
/* ... */
```

Then you can remove the line below which calls `soup_uri_free()`.

cog.c
```diff
@@ -284,6 +289,8 @@ on_handle_local_options (GApplication *application,
                                                          s_options.web_extensions_dir);
     }
 
+    g_object_set (shell, "ignore-tls-errors", s_options.ignore_tls_errors, NULL);
```
aperezdc: I would rather add a `cog_shell_set_ignore_tls_errors()` setter and use it here instead.

core/cog-shell.c
```diff
@@ -277,13 +285,21 @@ cog_shell_class_init (CogShellClass *klass)
...
+
     g_object_class_install_properties (object_class, N_PROPERTIES, s_properties);
 }
 
 
 static void
-cog_shell_init (CogShell *shell G_GNUC_UNUSED)
+cog_shell_init (CogShell *shell)
 {
+    PRIV (shell)->ignore_tls_errors = FALSE;
```
aperezdc: This line is not really needed because the object memory region is zeroed when it gets allocated; therefore the value is already `FALSE` due to that.

core/cog-webkit-utils.c
```diff
@@ -98,22 +98,19 @@ format_tls_error (GTlsCertificateFlags errors)
...
-
 gboolean
-cog_handle_web_view_load_failed_with_tls_errors (WebKitWebView       *web_view,
-                                                 char                *failing_uri,
-                                                 GTlsCertificate     *certificate,
-                                                 GTlsCertificateFlags errors,
-                                                 void                *user_data)
+cog_web_view_load_error_page (WebKitWebView *web_view,
+                              char *failing_uri,
+                              GTlsCertificateFlags errors)
```
aperezdc: I would keep the `GTlsCertificate` parameter for future-proofing. We may want to improve the built-in error pages at some point to include some information derived from the certificate.

core/cog-webkit-utils.c
```diff
@@ -98,22 +98,19 @@ format_tls_error (GTlsCertificateFlags errors)
...
 {
+
     g_autofree char *error_string = format_tls_error (errors);
     return load_error_page (web_view,
                             failing_uri,
                             "TLS Error",
```
aperezdc: Now that `cog_web_view_load_error_page()` would be part of the public API, I would rather remove the harcoded page title and instead make it a function parameter.

On a related note, I would only call `format_tls_error()` and add TLS certificate information if the passed `GTlsCertificate` is non-null. Maybe a function argument for an error message would be needed as well :thinking: 

I am not very sure about how to go with this change, it may need a bit more of thinking.

aperezdc:
@philn Hi there! Do you think you may have time to rehash this for [0.6.x](https://github.com/Igalia/cog/milestone/1)?

philn:
> @philn Hi there! Do you think you may have time to rehash this for [0.6.x](https://github.com/Igalia/cog/milestone/1)?

I unlikely will have time for this. If somebody else wants to pick it up though, feel free :)

timemaster5:
I would really appreciate this feature, please keep it going :)

aperezdc:
I'm going to upload soon a version of this patch based on @philn's which addresses my own review comments. Let's try to get at least the basic functionality to just ignore TLS errors merged, we can do the support for things like `--ignore-tls-errors=some.domain.com,someother.host` into a separate PR.

aperezdc:
Closing this in favor of #218

PR #139, opened 2019-09-02
Turn device scaling factor into a CogShell property

aperezdc:
This allows platform plug-ins reading the value of the property during their initialization, and shifts to them the responsibility of applying the device scaling factor in which way and moment is better for the pltform plug-in.

The FDO platform plug-in does not need any changes, as it picks the scaling factor from the Wayland compositor. The experimental DRM plug-in is changed to take into account the value of the `CogShell::device-scale` during view backend creation.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -13,11 +13,6 @@
 
 #if !COG_USE_WEBKITGTK
 # include "cog-platform.h"
-#if defined(WPE_CHECK_VERSION) && WPE_CHECK_VERSION(1, 3, 0)
-# define HAVE_DEVICE_SCALING 1
-#else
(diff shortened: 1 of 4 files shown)


aperezdc:
Currently I am working on the revamp to support multiple web views in the `devel/multiple-views` branch, I have just converted the DRM plug-in, and just added support for calculating the device scaling ration automatically in commit https://github.com/Igalia/cog/commit/99e58bcba29b50d1913a779f01b684b78f00b6bb (setting manually using the `CogDrmView.device-scale` property works, too).

Therefore I think the approach being suggested in this PR won't go anywhere. Let's close it.

PR #140, opened 2019-09-02
cog-shell: Do not leak the name property string

aperezdc:
Deallocate the string used as shell name in the `cog_shell_dispose()` method to avoid leaking it.


aperezdc:
Merged to `cog-0.3` (fd7a18a26c79da7cab3a61e1bdbe8e6c14237baa) and `cog-0.4` (4f39175d4a53fcfda18554e9ebb4c155b1f8dd8a) :heavy_check_mark:

PR #141, opened 2019-09-02
cog-shell: Apply g_get_prgname() fallback in _init method

aperezdc:
This is needed to ensure that `g_get_prgname()` is used as fallback name when instantiation is done through `g_object_new()` and the value of the property is not explicitly set to a non-`NULL` value.

Without this fix, the `g_get_prgname()` fallback would be only applied when instantiation is done using `cog_shell_new()`.


aperezdc:
Cherry picked to the `cog-0.4` branch as commit fba9dd9da5a432cc38071b2ee8efb1eed2437582.

PR #142, opened 2019-09-02
cog-shell: Use g_object_notify_by_pspec()

aperezdc:
Using `g_object_notify_by_pspec()` instead of `g_object_notify()` is preferred in the implementation of GObject classes because it avoids the lookup of the property given its string.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,7 +93,7 @@ set(COGCORE_SOURCES
     core/cog-webkit-utils.c
 )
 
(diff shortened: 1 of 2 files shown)


aperezdc:
> LGTM, I guess you already depend on GLib >= 2.26

Yes, we are also using `g_auto*` in Cog, which was introduced in GLib 2.44.

I will add an explicit version check in `CMakeLists.txt` to make sure we bail there if GLib is too old, instead of failing to build.

PR #146, opened 2019-09-08
core: ignore cancellation errors as the active URI may be changing

aleksander0m:
If the user clicks on a link before the current page has been fully
loaded, the view gets a "network error cancelled". We could safely
ignore these and go on.

Easy to reproduce on high-latency links (e.g. satellite links).


aperezdc:
Merged by hand as commit 43d0abab11a634bc003c0a8fe3916775bd099cff with some indentation fixes.

aperezdc:
Merged to the `cog-0.4` branch as well (commit 81551ce5a15778fa819acb1724889e2462aa3df9).

PR #147, opened 2019-09-12
CMake: Calculate and set a libtool soversion

aperezdc:
Import the `VersioningUtils.cmake` script and use it to configure a libtool-style soversion for the `cogcore` library.


aperezdc:
Cherry-picked into the `cog-0.4` branch as commit ce05d5d8677cdcaca8ee8724f3a4a9e9fc062af1.

PR #161, opened 2019-11-05
drm: make the platform module non-experimental.

zdobersek:


aperezdc:
@zdobersek Do we want to give #139 a re-spin before merging this?

PR #163, opened 2019-11-21
Several reworks done in the multiview-pluggable-shell branch

psaavedra:
* Moved platform/cog-platform-fdo.c to modules/cog-fdo-shell.c
* WL specific from platform/cog-platform-fdo.c moved to platform/pwl
* Cog loading a demo page using the FDO modules
    * Added temporary debug messages for debugging. To be removed before the merge
    * cog_launcher_constructed() creates a WebKitWebViewBackend and attach this to
      the created WebView
    * cog_launcher_constructed() creates a default WebView named "default"
    * Defined the cog_shell_get_view_backend() class function for the CogShell
    * Temporary removed the WAYLAND_1_10_OR_GREATER defines
    * WIP tasks labeled with a TODO mark. All these references must be addresed
      before the merge
    * cog_launcher_activate() loads test page using the webkit_web_view_load_uri()
    * cog_shell_create_view() fixed: Added g_value_init() step for the initialization
      of the G_VALUE items
    * main() calls the cog_launcher_get_default() and launch the result as the
      main GApplication
    * cog_shell_new_from_module() uses g_initable_new() instance of g_object_new() to
      create the CogShell object because the CogFdoShell implements GInitable now


platform/pwl.c
```diff
@@ -9,6 +9,36 @@
 
 #include <errno.h>
 
+/* for mmap */
+#include <sys/mman.h>
+
+#include <gio/gio.h>
+
+// TODO: #if defined(WPE_CHECK_VERSION) && WPE_CHECK_VERSION(1, 3, 0)
```
psaavedra: We dont want WPE specific definitions here.  probably the  DEVICE_SCALING should be defined with CMake

modules/cog-fdo-shell.c
```diff
@@ -81,19 +411,363 @@ cog_fdo_shell_class_init (CogFdoShellClass *klass)
...
 
+
+static bool
+capture_app_key_bindings (uint32_t keysym,
+                          uint32_t unicode,
+                          uint32_t state,
+                          uint8_t modifiers)
+{
+    CogLauncher *launcher = cog_launcher_get_default ();
+    WebKitWebView *web_view = (WebKitWebView *)
```
psaavedra: `cog_shell_get_view` a `CogView` ( what is defined witht `G_DEFINE_TYPE_WITH_PRIVATE (CogView, cog_view, WEBKIT_TYPE_WEB_VIEW`). I don't know the right syntax to get the actual `WebView`

modules/cog-fdo-shell.c
```diff
@@ -81,19 +411,363 @@ cog_fdo_shell_class_init (CogFdoShellClass *klass)
...
+                xdg_toplevel_set_fullscreen (win_data.xdg_toplevel, NULL);
+            else
+                xdg_toplevel_unset_fullscreen (win_data.xdg_toplevel);
+            win_data.is_fullscreen = ! win_data.is_fullscreen;
+            return true;
+        }
+        /* Ctrl+W, exit the application */
+        else if (modifiers == wpe_input_keyboard_modifier_control &&
+                 unicode == 0x17 && keysym == 0x77) {
+            g_application_quit (G_APPLICATION (launcher));
```
psaavedra: it is comes from the actua.  code launcher  (CogLauncher) is a `G_APP...`  .

modules/cog-fdo-shell.c
csaavedra: You can just use the GObject macro to cast to a `WebKitWebView`, that is `WEBKIT_WEB_VIEW`. It's the same as a direct cast but with type-checking.

modules/cog-fdo-shell.c
csaavedra: My reservation is that this is a loadable module in a library, we shouldn't be causing the application to quit, this should be decided by the application. I am not even sure that it's a good idea to handle any key shortcuts here to be honest.

modules/cog-fdo-shell.c
```diff
@@ -30,6 +44,322 @@ typedef struct {
     CogShell parent;
 } CogFdoShell;
 
+struct wpe_input_touch_event_raw touch_points[10];
+
+extern PwlData wl_data;
```
psaavedra: These structures are going to be shareable. All contains structs relative to the keyboard, window and EGL things show  all should be shared independent on the module which use those.

platform/pwl.c
```diff
@@ -9,6 +9,36 @@
 
 #include <errno.h>
 
+/* for mmap */
+#include <sys/mman.h>
```
psaavedra: To Check. Maybe is not longer needed in this file.

modules/cog-fdo-shell.c
```diff
@@ -30,6 +44,322 @@ typedef struct {
...
+extern PwlEGLData egl_data;
+extern PwlWinData win_data;
+extern PwlXKBData xkb_data;
+
+static struct {
+    struct wpe_view_backend_exportable_fdo *exportable;
+} wpe_host_data;
+
+static struct {
```
psaavedra: This information must to be self contained now in a CogView. E.x: each CogView should a particular Backend (for isolation)

platform/pwl.h
```diff
@@ -18,9 +42,192 @@ typedef struct {
...
+gboolean init_wayland (GError **error);
+void clear_wayland (void);
+
+gboolean create_window (GError **error);
+void destroy_window (void);
+
+gboolean init_input (GError **error);
+void clear_input (void);
+
+/* Keyboard */
```
psaavedra: Moved the keyboard specific  code to pwl.c for readability and reuse.

modules/cog-fdo-shell.c
psaavedra: done

modules/cog-fdo-shell.c
psaavedra: Added a TODO comment in this part of the code to address this problem later.

modules/cog-fdo-shell.c
csaavedra: I think they should be encapsulated in some singleton class for cleanliness, but just my opinion.

csaavedra:
I took the freedom to push a handful of commits directly to this merge request. I fixed a few things I found plus I added the cog/cogctl commands to be able to work with different views. Maybe this was not the best idea (since this merge request is about something specific), but it's already done :)

psaavedra:
I just forced the uploading of a new version Git history in the branch related to this PR. It could be great if you could check it. Hopefully we could merge this branch in the `multiview-pluggable-shell` and then continue with fixing the leftovers.

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+    }
+    eglReleaseThread ();
+}
+
+
+gboolean init_wayland (GError **error)
+{
+    g_debug ("Initializing Wayland...");
+
+    if (!(wl_data.display = wl_display_connect (NULL))) {
```
aperezdc: This is creating an additional Wayland connection, with the other being created by `pwl_display_connect()`. We definitely not want to have more than one connection.

Note that `cog_fdo_module_is_supported()` creates a `PwlDisplay` when the GIO module is loaded and the function called to check whether the module is usable. We want to reuse tha `wl_display` contained in `PwlDisplay` for all the Wayland API uses that need a display.

If you need the `wl_display*` from outside `pwl.c`, we could have an accessor like this:

```c
struct wl_display* pwl_display_get_wl_display(const PwlDisplay *self)
{
    g_return_val_if_fail (self, NULL);
   return self->display;
}
```

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+        g_set_error (error,
+                     G_FILE_ERROR,
+                     g_file_error_from_errno (errno),
+                     "Could not open Wayland display");
+        return FALSE;
+    }
+
+    wl_data.registry = wl_display_get_registry (wl_data.display);
```
aperezdc: Obtaining the Wayland registry and obtaining the interfaces of the Wayland extensions that are needed by the FDO shell module should be done either from `pwl_display_connect()` or from some other function called in `cog_fdo_shell_is_supported()` because the module *needs* at least the `xdg_wm_base` or the `zwp_fullscreen_shell_v1` interface. The `_is_supported()` function must return `FALSE` if any of the required Wayland protocols features are not available.

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+
+    g_debug ("New XDG toplevel configuration: (%" PRIu32 ", %" PRIu32 ")", width, height);
+
+    wl_data.resize_window(data);
+}
+
+static void
+xdg_toplevel_on_close (void *data, struct xdg_toplevel *xdg_toplevel)
+{
+    // TODO g_application_quit (g_application_get_default ());
```
aperezdc: Here we will want to define a `CogView:close` signal, and this function will use `g_signal_emit()` so in `cog.c` we can install a callback that gracefully exits the application (or just closes the view if it's in a separate window).

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+}
+
+static const struct xdg_toplevel_listener xdg_toplevel_listener = {
+    .configure = xdg_toplevel_on_configure,
+    .close = xdg_toplevel_on_close,
+};
+
+
+gboolean create_window (void *data, GError **error)
```
aperezdc: As you can imagine, most of this should be moved into a `pwl_window_create()` function :wink:

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+        win_data.xdg_toplevel =
+            xdg_surface_get_toplevel (win_data.xdg_surface);
+        g_assert (win_data.xdg_toplevel);
+
+        xdg_toplevel_add_listener (win_data.xdg_toplevel,
+                                   &xdg_toplevel_listener, data);
+        // TODO: xdg_toplevel_set_title (win_data.xdg_toplevel, COG_DEFAULT_APPNAME);
+
+        const char *app_id = NULL;
+        GApplication *app = g_application_get_default ();
```
aperezdc: Setting the application identifier is going to be tricky without breaking the encapsulation. Maybe we'll have to add it as a property of `CogShell`, so then the `CogFdoShell` implementation can call into a `pwl_window_set_application_id()` in turn.

I don't have a very clear idea how to proceed here, so we can leave a `TODO` around for this.

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+                                       &shell_surface_listener,
+                                       data);
+        wl_shell_surface_set_toplevel (win_data.shell_surface);
+
+        /* wl_shell needs an initial surface configuration. */
+        configure_surface_geometry (0, 0);
+    }
+
+    const char* env_var;
+    if ((env_var = g_getenv ("COG_PLATFORM_FDO_VIEW_FULLSCREEN")) &&
```
aperezdc: We *need* to stop using environment variables here. The idea is to have a `pwl_window_set_fullscreen(PwlWindow*, bool)` function that can be called, and a virtual method `cog_view_set_fullscreen(CogView*, gboolean)` + `CogView.is-fullscreen` property.

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+            wl_shell_surface_set_fullscreen (win_data.shell_surface,
+                                             WL_SHELL_SURFACE_FULLSCREEN_METHOD_SCALE,
+                                             0,
+                                             NULL);
+        } else {
+            g_warning ("No available shell capable of fullscreening.");
+            win_data.is_fullscreen = false;
+        }
+    }
+    else if ((env_var = g_getenv ("COG_PLATFORM_FDO_VIEW_MAXIMIZE")) &&
```
aperezdc: Same for maximization, we want to control it per-view once we have multiple views.

platform/pwl.c
```diff
@@ -39,3 +61,775 @@ pwl_display_destroy (PwlDisplay *self)
...
+            wl_shell_surface_set_maximized (win_data.shell_surface, NULL);
+        } else {
+            g_warning ("No available shell capable of maximizing.");
+            win_data.is_maximized = false;
+        }
+    }
+
+    return TRUE;
+}
+void destroy_window (void)
```
aperezdc: Would be `pwl_window_destroy(PwlWindow*)`.

platform/pwl.h
```diff
@@ -18,9 +41,192 @@ typedef struct {
...
+
+#if HAVE_DEVICE_SCALING
+typedef struct output_metrics {
+  struct wl_output *output;
+  int32_t name;
+  int32_t scale;
+} output_metrics;
+#endif /* HAVE_DEVICE_SCALING */
+
+typedef struct {
```
aperezdc: Most of the things in `wl_data` belong in `PwlDisplay`.

platform/pwl.h
```diff
@@ -18,9 +41,192 @@ typedef struct {
     struct wl_compositor *compositor;
 } PwlDisplay;
 
+typedef struct {
+    struct egl_display *display;
+    EGLContext context;
+    EGLConfig egl_config;
+} PwlEGLData;
```
aperezdc: These are global for the Wayland connection, they should be inside `PwlDisplay`.

platform/pwl.h
```diff
@@ -18,9 +41,192 @@ typedef struct {
...
+    } indexes;
+
+    struct {
+        xkb_mod_index_t control;
+        xkb_mod_index_t alt;
+        xkb_mod_index_t shift;
+    } modifier;
+
+    uint8_t modifiers;
+} PwlXKBData;
```
aperezdc: Ditto, to be moved into `PwlDisplay`.

platform/pwl.h
```diff
@@ -18,9 +41,192 @@ typedef struct {
...
+
+#if HAVE_DEVICE_SCALING
+typedef struct output_metrics {
+  struct wl_output *output;
+  int32_t name;
+  int32_t scale;
+} output_metrics;
```
aperezdc: The device scaling should be tracked for each surface, because each Wayland surface (“window”) can be in a different output, and outputs can have each a different device scaling factor.

platform/pwl.h
```diff
@@ -18,9 +41,192 @@ typedef struct {
...
+    struct xdg_surface *xdg_surface;
+    struct xdg_toplevel *xdg_toplevel;
+    struct wl_shell_surface *shell_surface;
+
+    uint32_t width;
+    uint32_t height;
+
+    bool is_fullscreen;
+    bool is_maximized;
+} PwlWinData;
```
aperezdc: This should be an opaque `PwlWindow` structure, and for each view we allocate one—instead of a global.

CMakeLists.txt
```diff
@@ -61,6 +61,8 @@ set(COG_HOME_URI "" CACHE STRING "Default home URI")
 set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/lib/cog/modules"
     CACHE STRING "Default search path for loadable modules")
 
+set(COG_DEVICE_SCALING 1)
```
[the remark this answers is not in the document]
psaavedra: I did that to avoid dependencies (in the includes) with  wpebackend-fdo-1.0 (`WPE_CHECK_VERSION`) for the platform. In other case `wpebackend-fdo-1.0` has to be added to the required modules `pkg_check_modules` for the platform-wl library as well.

CMakeLists.txt
```diff
@@ -211,12 +215,27 @@ if (COG_MODULE_FDO_SHELL)
         egl
         xkbcommon
     )
+
+    if(ModuleFdoShell_wpebackend-fdo-1.0_VERSION VERSION_LESS "1.3")
```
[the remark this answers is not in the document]
psaavedra: He he, I see the point. I agree. let's uses `WAYLAND_VERSION_*` instance of `WPE_FDO_CHECK_VERSION`

CMakeLists.txt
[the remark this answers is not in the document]
aperezdc: Not really, it's possible to check whether the macro is defined before trying to use it:

```c
#if defined(WPE_FDO_CHECK_VERSION) && WPE_FDO_CHECK_VERSION(1, 3, 0)
  /* ... */
#endif 
```

core/cog-shell.c
```diff
@@ -186,7 +186,7 @@ cog_shell_new_from_module (const char *name,
         return NULL;
     }
 
-    return g_object_new (shell_type, "name", name, NULL);
+    return g_initable_new (shell_type, NULL, NULL, "name", name, NULL);
```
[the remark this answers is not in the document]
psaavedra: It looks like we could deal with this at the beginning. Lets's create a issue to this work on this later, WDYT?

platform/pwl.h
psaavedra: The current implementation is reusing wayland surface. I suppose we could rework this staff restructuring the structures that you commented

CMakeLists.txt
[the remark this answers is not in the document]
psaavedra: The point is this not only checks the `WPE_FDO_CHECK_VERSION` but also the version number. Anyway I'm fine with using `WAYLAND_VERSION` instance of `WPE_FDO_CHECK_VERSION`  as you suggested in the comment below.

psaavedra:
> There's many things here which need improvements, I have a WIP [multiview-continuation](https://github.com/Igalia/cog/tree/aperezdc/multiview-continuation) branch which already has some code in the direction pointed by the review comments.
> 
> I think the most important bit is avoiding having two Wayland connections to the compositor. That should be relatively easy to start with.
> 
> My advice would be to make move all these changes into an integration branch, and continue improving the code with small PRs which will get merged to the integration branch. Once the code is mature, we will be able to merge the integration branch to `master`.
> 
> WDYT?

What do you mean? A) rebase this branch in the top of `aperezdc/multiview-continuation` like a single PR or, on the contrary,  B) create a PR per each differentiated changed: scaling, pwl_data encapsulation,  Wayland connections to the compositor,  allows to register callbacks for keybindings ...

aperezdc:
Thanks a lot for all the work here!

The commits from this PR are already merged in the [devel/multiple-views](https://github.com/Igalia/cog/commits/devel/multiple-views) branch where we will continue improving things until we're ready to merge the feature to `master` — and therefore I am closing this PR.

PR #164, opened 2019-11-21
cog-fdo-shell: Implement the GInitable interface

csaavedra:
Since the FDO backend, wayland, etc. can fail to initialize
it makes sense to have the module implement the GInitable
interface. For now, the initable interface only tries to
load the fdo module.

### DIFF
--- a/modules/cog-fdo-shell.c
+++ b/modules/cog-fdo-shell.c
@@ -30,9 +30,12 @@ typedef struct {
     CogShell parent;
 } CogFdoShell;
 
+static void cog_fdo_shell_initable_iface_init (GInitableIface *iface);
+
 G_DEFINE_DYNAMIC_TYPE_EXTENDED (CogFdoShell, cog_fdo_shell, COG_TYPE_SHELL, 0,
     g_io_extension_point_implement (COG_MODULES_SHELL_EXTENSION_POINT,
-                                    g_define_type_id, "fdo", 100))
+                                    g_define_type_id, "fdo", 100);
+    G_IMPLEMENT_INTERFACE_DYNAMIC (G_TYPE_INITABLE, cog_fdo_shell_initable_iface_init))
 
 G_MODULE_EXPORT
 void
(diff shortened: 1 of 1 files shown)


modules/cog-fdo-shell.c
```diff
@@ -30,9 +30,12 @@ typedef struct {
     CogShell parent;
 } CogFdoShell;
 
+static void cog_fdo_shell_initable_iface_init (GInitableIface *iface);
+
 G_DEFINE_DYNAMIC_TYPE_EXTENDED (CogFdoShell, cog_fdo_shell, COG_TYPE_SHELL, 0,
     g_io_extension_point_implement (COG_MODULES_SHELL_EXTENSION_POINT,
-                                    g_define_type_id, "fdo", 100))
+                                    g_define_type_id, "fdo", 100);
+    G_IMPLEMENT_INTERFACE_DYNAMIC (G_TYPE_INITABLE, cog_fdo_shell_initable_iface_init))
```
psaavedra: For ref:  https://developer.gnome.org/gobject/stable/GTypeModule.html#G-DEFINE-DYNAMIC-TYPE-EXTENDED:CAPS

This is the `CODE` part of the `G_DEFINE_DYNAMIC_TYPE_EXTENDED`, the custom code that gets inserted in the *_get_type() function. This leads in a  `g_type_module_add_interface ()`  which logic will be implemented by the  `cog_fdo_shell_initable_iface_init()` during the registration of the type.

modules/cog-fdo-shell.c
```diff
@@ -92,8 +95,31 @@ cog_fdo_shell_class_finalize (CogFdoShellClass *klass)
 static void
 cog_fdo_shell_init (CogFdoShell *shell)
 {
+    TRACE ("");
+}
+
+gboolean
+cog_fdo_shell_initable_init (GInitable *initable,
```
psaavedra: The code executed during the initialization of the cog_fdo_shell.

This [code](https://github.com/Igalia/cog/pull/163/files?file-filters%5B%5D=.c&file-filters%5B%5D=.txt&hide-deleted-files=true#diff-b0e4f3f94ad14a1e344cae725bb1cd54R687) has to be relocated here.

csaavedra: I saw in some `gio` classes that implement GInitable and have different entry points, that it is also possible to just keep an internal error reference from those entry points if something fails and just use the `initable_init` method to check if something went wrong and then return `FALSE` in that case. I think in our case, that we have only one entry point, it makes more sense to just move the code here as you say.

PR #165, opened 2020-01-15
Support for direct-display video content rendering

philn:
By relying on the direct-display and content-protection Weston Wayland
protocols, the browser is now able to render video frames by passing them
directly to the display controller, without any additional GPU import in the
compositor. DMABufs are wrapped in sub-surfaces and positioned according to the
coordinates provided by WebKit.

Minimal support for multiple videos rendering support was added, although in
most cases the underlying hardware platform will be able to render only one
video at a time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,8 @@ option(COG_PLATFORM_FDO "Build the FDO platform module" ON)
 option(COG_PLATFORM_DRM "Build the DRM platform module" OFF)
 option(COG_BUILD_PROGRAMS "Build and install programs as well" ON)
 option(INSTALL_MAN_PAGES "Install the man(1) pages if COG_BUILD_PROGRAMS is enabled" ON)
+option(COG_WESTON_DIRECT_DISPLAY "Build direct display support for the FDO platform module" OFF)
+
 set(COG_APPID "" CACHE STRING "Default GApplication unique identifier")
 set(COG_HOME_URI "" CACHE STRING "Default home URI")
 
@@ -256,6 +258,50 @@ if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
         add_definitions(-DCOG_IM_API_SUPPORTED=0)
     endif ()
 
+    find_path(WPEBACKEND_FDO_HAS_VIDEO_PLANE_DISPLAY_DMABUF_EXT
+        NAMES wpe/extensions/video-plane-display-dmabuf.h
+        PATHS ${COGPLATFORM_FDO_DEPS_INCLUDE_DIRS}
+        NO_DEFAULT_PATH
+    )
+    if (WPEBACKEND_FDO_HAS_VIDEO_PLANE_DISPLAY_DMABUF_EXT)
+        set(WPEBACKEND_FDO_HAS_VIDEO_PLANE_DISPLAY_DMABUF_EXT ON)
+    else ()
+        set(WPEBACKEND_FDO_HAS_VIDEO_PLANE_DISPLAY_DMABUF_EXT OFF)
+    endif ()
+
+    if (COG_WESTON_DIRECT_DISPLAY AND WPEBACKEND_FDO_HAS_VIDEO_PLANE_DISPLAY_DMABUF_EXT)
+        file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/weston)
+        pkg_check_modules(LIBWESTON REQUIRED libweston-9-protocols)
+        pkg_get_variable(LIBWESTON_PKG_DATA_DIR libweston-9-protocols "pkgdatadir")
+        find_package(WaylandScanner REQUIRED)
+        add_custom_command(
+            OUTPUT ${CMAKE_BINARY_DIR}/weston/weston-direct-display-client-protocol.h
+                   ${CMAKE_BINARY_DIR}/weston/weston-direct-display-protocol.c
+            DEPENDS ${LIBWESTON_PKG_DATA_DIR}/weston-direct-display.xml
+            COMMAND ${WAYLAND_SCANNER} client-header < ${LIBWESTON_PKG_DATA_DIR}/weston-direct-display.xml > ${CMAKE_BINARY_DIR}/weston/weston-direct-display-client-protocol.h
+            COMMAND ${WAYLAND_SCANNER} ${WAYLAND_SCANNER_CODE_ARG} < ${LIBWESTON_PKG_DATA_DIR}/weston-direct-display.xml > ${CMAKE_BINARY_DIR}/weston/weston-direct-display-protocol.c
+            VERBATIM
+        )
+
+        add_custom_command(
+            OUTPUT ${CMAKE_BINARY_DIR}/weston/weston-content-protection-client-protocol.h
+                   ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
+            DEPENDS ${LIBWESTON_PKG_DATA_DIR}/weston-content-protection.xml
+            COMMAND ${WAYLAND_SCANNER} client-header < ${LIBWESTON_PKG_DATA_DIR}/weston-content-protection.xml > ${CMAKE_BINARY_DIR}/weston/weston-content-protection-client-protocol.h
+            COMMAND ${WAYLAND_SCANNER} ${WAYLAND_SCANNER_CODE_ARG} < ${LIBWESTON_PKG_DATA_DIR}/weston-content-protection.xml > ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
+            VERBATIM
+        )
+
+        list(APPEND COGPLATFORM_FDO_SOURCES
+            ${CMAKE_BINARY_DIR}/weston/weston-direct-display-protocol.c
+            ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
+        )
+        list(APPEND COGPLATFORM_FDO_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/weston)
+        set(ENABLE_WESTON_DIRECT_DISPLAY 1)
+    else ()
+        set(ENABLE_WESTON_DIRECT_DISPLAY 0)
+    endif ()
+
     add_library(cogplatform-fdo MODULE ${COGPLATFORM_FDO_SOURCES})
     set_property(TARGET cogplatform-fdo PROPERTY C_STANDARD 99)
     target_include_directories(cogplatform-fdo PUBLIC wayland ${COGPLATFORM_FDO_INCLUDE_DIRS})
@@ -264,6 +310,7 @@ if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
         PUBLIC ${COGPLATFORM_FDO_CFLAGS}
         PRIVATE -DG_LOG_DOMAIN=\"Cog-FDO\"
     )
+    target_compile_options(cogplatform-fdo PRIVATE -DCOG_ENABLE_WESTON_DIRECT_DISPLAY=${ENABLE_WESTON_DIRECT_DISPLAY})
 
     find_package(WaylandProtocols REQUIRED)
     add_wayland_protocol(cogplatform-fdo CLIENT xdg-shell)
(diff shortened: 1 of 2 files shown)


platform/cog-platform-fdo.c
```diff
@@ -722,7 +817,6 @@ keyboard_on_enter (void *data,
                    struct wl_surface *surface,
                    struct wl_array *keys)
 {
-    g_assert (surface == win_data.wl_surface);
```
aperezdc: Does this assertion need to be removed?

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1392,130 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
     wl_surface_commit (win_data.wl_surface);
 }
 
+#if COG_ENABLE_WESTON_DIRECT_DISPLAY
+static void
+create_succeeded(void *data,
+		 struct zwp_linux_buffer_params_v1 *params,
+		 struct wl_buffer *new_buffer)
```
aperezdc: The indentation of the continuation lines looks a bit wonky here :wavy_dash:

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1392,130 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+}
+
+static void
+create_failed(void *data, struct zwp_linux_buffer_params_v1 *params)
+{
+    struct buffer *buffer = data;
+
+    buffer->buffer = NULL;
+    zwp_linux_buffer_params_v1_destroy (params);
+    g_assert_not_reached ();
```
aperezdc: This looks odd. Is it really expectated that `create_failed()` will never be called? If that's the case, why not asserting at the top of the function and removing the rest of the code?

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1392,130 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+}
+
+static const struct zwp_linux_buffer_params_v1_listener params_listener = {
+	create_succeeded,
```
aperezdc: Same as before here: please use named field initializers.

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1392,130 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+    }
+
+    uint64_t modifier = DRM_FORMAT_MOD_INVALID;
+    struct zwp_linux_buffer_params_v1 *params = zwp_linux_dmabuf_v1_create_params (wl_data.dmabuf);
+    if (wl_data.direct_display != NULL)
+        weston_direct_display_v1_enable (wl_data.direct_display, params);
+
+    video_surface *surf = (video_surface*) g_hash_table_lookup (win_data.video_surfaces, GUINT_TO_POINTER(id));
+    if (!surf) {
+        surf = calloc (1, sizeof (video_surface));
```
aperezdc: How about using `g_slice_new0()` here? 

```c
surf = g_slice_new0 (video_surface);
```

And of course `g_slice_free()` where the `video_surface` is deallocated.

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1392,130 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+
+    zwp_linux_buffer_params_v1_add (params, fd, 0, 0, stride, modifier >> 32, modifier & 0xffffffff);
+
+    if ((x + width) > win_data.width)
+        width -= x;
+
+    if ((y + height) > win_data.height)
+        height -= y;
+
+    struct buffer* buffer = calloc(1, sizeof(struct buffer));
```
aperezdc: You could use `g_slice_new0()` here as well.

platform/cog-platform-fdo.c
```diff
@@ -1306,9 +1556,19 @@ clear_wayland (void)
     if (wl_data.shell != NULL)
         wl_shell_destroy (wl_data.shell);
 
+    if (wl_data.subcompositor != NULL)
+        wl_subcompositor_destroy (wl_data.subcompositor);
```
aperezdc: These two lines could just be `g_clear_pointer (&wl_data.subcompositor, wl_subcompositor_destroy);`.

platform/cog-platform-fdo.c
```diff
@@ -1306,9 +1556,19 @@ clear_wayland (void)
...
+
     if (wl_data.compositor != NULL)
         wl_compositor_destroy (wl_data.compositor);
 
+#if COG_ENABLE_WESTON_DIRECT_DISPLAY
+    if (wl_data.protection != NULL)
+        weston_content_protection_destroy (wl_data.protection);
+    if (wl_data.direct_display != NULL)
+        weston_direct_display_v1_destroy (wl_data.direct_display);
```
aperezdc: Same here, `g_clear_pointer()` could be used instead.

platform/cog-platform-fdo.c
philn: I think the assert could be removed actually.

CMakeLists.txt
```diff
@@ -256,6 +258,39 @@ if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
...
+            COMMAND ${WAYLAND_SCANNER} ${WAYLAND_SCANNER_CODE_ARG} < ${LIBWESTON_PKG_DATA_DIR}/weston-content-protection.xml > ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
+            VERBATIM
+        )
+
+        list(APPEND COGPLATFORM_FDO_SOURCES
+            ${CMAKE_BINARY_DIR}/weston/weston-direct-display-protocol.c
+            ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
+        )
+        list(APPEND COGPLATFORM_FDO_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/weston)
+        add_definitions(-DCOG_ENABLE_WESTON_DIRECT_DISPLAY=1)
```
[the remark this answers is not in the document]
philn: That doesn't seem doable because the target is defined afterwards. I can't define the target before either because when direct display is enabled I need to add compilation units in the target.

platform/cog-platform-fdo.c
philn: I had to remove it during the development cycle but checking again now, the assertion is no longer hit, so let's keep it :)

CMakeLists.txt
```diff
@@ -256,6 +258,39 @@ if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
         add_definitions(-DCOG_IM_API_SUPPORTED=0)
     endif ()
 
+    if (COG_WESTON_DIRECT_DISPLAY)
+        file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/weston)
+        pkg_check_modules(LIBWESTON REQUIRED libweston-9-protocols)
+        pkg_get_variable(LIBWESTON_PKG_DATA_DIR libweston-9-protocols "pkgdatadir")
+        find_package(WaylandScanner REQUIRED)
+        add_custom_command(
```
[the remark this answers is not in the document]
philn: My CMake skills are probably not good enough for this, so I'd prefer to handle this in follow-up PR indeed to not block this one further.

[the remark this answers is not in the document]
philn: #178

philn:
@aperezdc ping

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1394,126 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+	buffer->buffer = new_buffer;
+}
+
+static void
+create_failed(void *data, struct zwp_linux_buffer_params_v1 *params)
+{
+	zwp_linux_buffer_params_v1_destroy (params);
+
+	struct buffer *buffer = data;
+    buffer->buffer = NULL;
```
zdobersek: tabs vs. spaces? What do we use?

platform/cog-platform-fdo.c
```diff
@@ -1266,6 +1394,126 @@ on_export_fdo_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
...
+        }
+        return;
+    }
+
+    uint64_t modifier = DRM_FORMAT_MOD_INVALID;
+    struct zwp_linux_buffer_params_v1 *params = zwp_linux_dmabuf_v1_create_params (wl_data.dmabuf);
+    if (wl_data.direct_display != NULL)
+        weston_direct_display_v1_enable (wl_data.direct_display, params);
+
+    video_surface *surf = (video_surface*) g_hash_table_lookup (win_data.video_surfaces, GUINT_TO_POINTER(id));
```
zdobersek: Nit: a simple wl_list would work better, except it's a bit harder to work *with*. Not a big problem though.

platform/cog-platform-fdo.c
```diff
@@ -54,6 +63,26 @@
 # define HAVE_DEVICE_SCALING 0
 #endif /* WPE_CHECK_VERSION */
 
+#if COG_ENABLE_WESTON_DIRECT_DISPLAY
+#define VIDEO_BUFFER_FORMAT DRM_FORMAT_YUYV
+struct buffer {
```
zdobersek: I think this struct would be better named `video_buffer`, given it's used along with `video_surface`.

platform/cog-platform-fdo.c
```diff
@@ -54,6 +63,26 @@
...
+    int fd;
+
+    struct wpe_video_plane_display_dmabuf_export* dmabuf_export;
+};
+
+typedef struct {
+    struct weston_protected_surface *protected_surface;
+    struct wl_surface *wl_surface;
+    struct wl_subsurface *wl_subsurface;
+} video_surface;
```
zdobersek: There a reason why this uses `typedef`?

platform/cog-platform-fdo.c
```diff
@@ -1306,8 +1554,13 @@ clear_wayland (void)
...
 
-    if (wl_data.compositor != NULL)
-        wl_compositor_destroy (wl_data.compositor);
+    g_clear_pointer (&wl_data.subcompositor, wl_subcompositor_destroy);
+    g_clear_pointer (&wl_data.compositor, wl_compositor_destroy);
+
+#if COG_ENABLE_WESTON_DIRECT_DISPLAY
+    g_clear_pointer (&wl_data.protection, weston_content_protection_destroy);
+    g_clear_pointer (&wl_data.direct_display, weston_direct_display_v1_destroy);
+#endif
```
zdobersek: If the modifiers data will be allocated from dynamic memory, it should be freed here.

CMakeLists.txt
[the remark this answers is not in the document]
philn: Yeah, actually I've found a way to do this using an intermediary variable.

platform/cog-platform-fdo.c
```diff
@@ -452,6 +494,36 @@ static const struct wl_surface_listener surface_listener = {
...
 
+#if COG_ENABLE_WESTON_DIRECT_DISPLAY
+static void
+dmabuf_modifiers(void *data, struct zwp_linux_dmabuf_v1 *zwp_linux_dmabuf,
+                 uint32_t format, uint32_t modifier_hi, uint32_t modifier_lo)
+{
+    switch (format) {
+    case VIDEO_BUFFER_FORMAT:
+        ++wl_data.modifiers_count;
+        wl_data.modifiers = realloc(wl_data.modifiers,
```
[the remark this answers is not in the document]
philn: The whole listener seems useless actually.

platform/cog-platform-fdo.c
philn: spaces!

CMakeLists.txt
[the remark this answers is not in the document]
aperezdc: I think it should be even possible to move creation of the target to be done earlier. All the `target_<foo>()` CMake functions are cumulative, and the order in which they are called on the target is irrelevant.

platform/cog-platform-fdo.c
aperezdc: Spaces.

PR #167, opened 2020-01-17
Remove shared pwl globals

psaavedra:


platform/pwl.c
```diff
@@ -364,14 +364,14 @@ init_egl (PwlDisplay *display, GError **error)
...
-        clear_egl ();
+        pwl_egl_destroy ();
         return FALSE;
     }
 
     return TRUE;
 }
 
-void clear_egl (void)
+void pwl_egl_destroy (void)
```
aperezdc: This should be called `pwl_display_egl_deinit()`.

Note that:

- “create” is the opposite of “destroy”.
- “init” is the opposite of “deinit”.

So if we have some `foo_create()` function, the opposite that frees resources should be `foo_destroy()`; and if we have a `bar_init()`, the opposite that frees resources should `bar_deinit()`.

PR #168, opened 2020-01-17
input methods

carlosgcampos:


carlosgcampos:
Build is failing because of the new WebKit API. I tried to avoid bumping the requirements, but I'm not sure it will be possible. There's also new API in cog-platform.h that uses the new webkit type, that's a public header, so we can't add the COG_IM_API_SUPPORTED ifdef there.

PR #169, opened 2020-01-20
pwl: PwlEGLData data is moved to PwlDisplay

psaavedra:
PwlEGLData is removed in favour of PwlDisplay which now contains the data structures contained previously by PwlEGLData.

Also, the destroy_window () and pwl_display_egl_deinit () receives now a PwlDisplay as argument.


aperezdc:
FWIW, I think that the CI is failing because we have to rebase the `multiview-no-globals` branch after #166 was merged to `master`.

PR #170, opened 2020-01-20
no globals: Moving global references and refactorings 

psaavedra:
* pwl: PwlDisplay has a 'void* shell' element what is used to attach the CogShell initialized:
    * create_window() and clear_input uses the PwlDisplay as argument.
    * The WL handlers receives the PwlDisplay as data context
* pwl: wl_seat moved from PwlData to PwlDisplay
* pwl: Remove shared win_data global. Remove the win_data global exported from the "pwl" module.
  * Instead, use the s_win_data as static variable in cog-fdo-shell.
  * Functions in pwl receives the PwlWinData structure as argument or as part of the contexts, for the case of callbacks.
 * pwl: callbacks for keyboard_listener encapsulated in pwl.c  The handlers for the keyboard_listener are defined in the  seat_on_capabilities() during the initialization of the Wayland platform.
* pwl: pointer_on_*()  moved to pwl and added the *on_pointer_*() callbacks in the PwlDisplay. These callback are used to for especific code for the modules.
* pwl: surface_handle_enter() moved to pwl and added the *on_surface_enter() callback in the PwlDisplay.  This callback is used to for specific code for the modules.   In a future, when each CogView uses its own surface, this callback   should be moved to the PwlWinData.
* pwl: Remove shared win_data global. Remove the win_data global exported from the "pwl" module. Instead, use the s_win_data as static variable in cog-fdo-shell.   Functions in pwl receives the PwlWinData struct as argument or   as part of the contexts, for the case of callbacks.

### DIFF
--- a/modules/cog-fdo-shell.c
+++ b/modules/cog-fdo-shell.c
@@ -156,17 +156,17 @@ on_pointer_on_axis (PwlDisplay* display, void *userdata)
 static void
 on_touch_on_down (PwlDisplay* display, void *userdata)
 {
-    CogFdoShellCallbackData *cb_userdata = userdata;
-    struct wpe_view_backend* backend = cog_shell_get_active_wpe_backend (cb_userdata->shell);
+    CogShell *shell = userdata;
+    struct wpe_view_backend* backend = cog_shell_get_active_wpe_backend (shell);
     struct wpe_input_touch_event_raw raw_event = {
         wpe_input_touch_event_type_down,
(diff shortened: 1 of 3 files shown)


aperezdc:
Thanks a lot for all the work here!

The commits from this PR are already merged in the [devel/multiple-views](https://github.com/Igalia/cog/commits/devel/multiple-views) branch where we will continue improving things until we're ready to merge the feature to `master` — and therefore I am closing this PR.

Issue #77, opened 2019-01-15
Support multiple web views simultaneously


aperezdc:
Currently Cog supports only a single web view, and it would be interesting to provide support for more than one. In my mind, the minimum that needs to be done is:

- [x] Allow `CogShell` to contain more than a single web view, one of them considered to be the “active” one.
- [ ] Augment the platform plug-in interface to:
  - [ ] Request from the platform plug-in more than one view backend.
  - [ ] Letting the platform plug-in know which view (and hence view backend) is active.
- [x] Add new API to `CogShell` to:
  - [x] Enumerate web views.
  - [x] Check which view is the active one.
  - [x] Select which view is the active one.
- [ ] Expose multiple web view support in `CogLauncher`.
- [x] Add commands to `cogctl` for managing multiple web views.
- [ ] Port/add platform implementations to the new API:
  - [x] FDO (Wayland)
  - [ ] DRM (Kernel Mode Setting + Direct Rendering Manager).
  - [ ] Simple (only one view supported, fallback for basic WPE backends like the RDK one).

The simplest implementation (on the platform plug-in side) would be to only display frames coming from the active view, and keeping around the most recent frame from the other views, so it can be used as the first frame shown when the active view is changed.

There is one problem: when using the fall-back WPE “default” view backend (that is: no platform plug-in is used), we cannot know what the WPE backend implementation will do, therefore in that case I think the safest approach would be disallow creating more than one web view.

aperezdc:
This is being worked on in the branch [devel/multiple-views](https://github.com/Igalia/cog/commits/devel/multiple-views). It still needs work, but it is already in an usable state.

aperezdc:
The changes in the `devel/multiple-views` branch are too many for a single PR that would apply to `master`—I tried a few times to rebase but it is a Herculean task. I am going to try instead moving things slowly into `master` to make it slowly morph into shape, let's see how that goes.

aperezdc:
While there has been some work to slowly move this code over to the main branch, it's not there yet, so I am changing the milestone to 0.14 🙃

aperezdc:
The infrastructure needed for this is now merged the `master` branch, so let's close this and for the remaining issues we can open new issues.

PR #171, opened 2020-01-24
cog-shell: Add default WebKitWebContext

psaavedra:
The CogShell creates a default WebKitWebContext what it used by the created CogViews when no other context is passed as argument in the constructor.


aperezdc:
This makes sense with the multiple views support (#77) which we will be doing for 0.8, so setting the same milestone for this.

aperezdc:
I think that the way I would like to go forward with the `CogShell` API is “a shell is a collection of views” but not forcing them to share settings. Therefore I think there should not be a default set of `WebKitSettings`/`WebKitWebContext` attached to a `CogShell`. Instead of Cog enforcing that, the application should be responsible for deciding which views share settings. And I am even thinking of moving the `CogLauncher` class out of `core/` going forward.

At any rate, I think we should not merge this. Thanks for filing the PR, though—it got me to reflect about how the API should look and to realize that Cog itself should allow different policies to be implemented.

PR #181, opened 2020-02-18
Adjust plugin entrypoint names

zdobersek:
To avoid cog_platform_try_load() loading equally-named symbols from
libcogcore.so when they are not defined in the specified platform
plugin, the plugin entrypoints are renamed to cog_platform_plugin_*.

### DIFF
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -65,21 +65,21 @@ cog_platform_try_load (CogPlatform *platform,
     if (!platform->so)
         return FALSE;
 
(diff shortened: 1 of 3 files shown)


aperezdc:
For future reference: this kind of thing will not be an issue anymore once we switch to GIO modules, which is already being done as part of the work for #77. Nevertheless, we definitely want this fix landed because it is not still clear whether using GIO modules will be done on time for the 0.6.x release.

Thanks for the patch, @zdobersek!

PR #184, opened 2020-02-25
fdo: Add WPE backends directory to the library RPATH

aperezdc:
Set the `RPATH` of `libcogplatform-fdo.so` to the directory where WPE backends are installed. This is needed to ensure that it will be possible to find `libWPEBackend-fdo-1.0` at runtime because the backends directory is not included in the default dynamic linker search paths.

----

This needs the changes from https://github.com/WebPlatformForEmbedded/libwpe/pull/64 and https://github.com/Igalia/WPEBackend-fdo/pull/99 and therefore the CI is expected to fail. See https://github.com/WebPlatformForEmbedded/libwpe/issues/59 for more details.


aperezdc:
I may suggest a different approach during the next development cycle.

PR #185, opened 2020-03-09
Keep a reference to the config GKeyFile on CogShell

zdobersek:
Store a reference of the GKeyFile configuration file specified on
the command line on the CogShell object, also providing a
cog_shell_get_config_file() accessor.

This makes the configuration file accessible from the platform
module, allowing access to any module-specific configuration.


zdobersek:
> @zdobersek: Do you need/want to have this in Cog 0.6? (To Be Released Soon™)

Perhaps not.

PR #187, opened 2020-03-13
im-context-fdo-v1: convert from bytes to character the offset and len…

carlosgcampos:
…gth of surroundiong text to delete

WebKit expects the offset and length in characters, and wayland
provides them in bytes, so we need to convert them.


aperezdc:
Cherry-picked into the `cog-0.6` branch as commit 9c30bfc3c856138de972fd685b854841b03d2ca9 — thanks!

PR #188, opened 2020-03-17
CMake: Simplify FDO platform plug-in build definitions

### LINKED ISSUE #178: CMake: Improve Wayland protocol handling
It would be nicer to change the `add_wayland_protocol()` function in `cmake/FindWaylandProtocols.cmake` to support specifying the base directory to allow passing `${LIBWESTON_PKG_DATA_DIR}` (and still using `${WAYLAND_PROTOCOLS}` by default).

We probably want some CMake function to check whether the XML source for a certain protocol exists, too.

Those improvements could be done in follow-up PRs, though.

_Originally posted by @aperezdc in https://github.com/Igalia/cog/pull/165_

aperezdc:
Improve `FindWaylandProtocols.cmake` to allow passing full paths (among other goodies) to `add_wayland_protocol()`, then use that to simplify the top-level `CMakeLists.txt`, along with imported targets and commands which modify the target instead of carrying around list variables.

----

Fixes #178

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,110 +222,69 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cogcore.pc
 # libcogplaform-fdo
 
 if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
-    set(WAYLAND_1_10_OR_GREATER OFF)
-    pkg_check_modules(WAYLAND REQUIRED wayland-client)
-    if (NOT (WAYLAND_VERSION VERSION_LESS 1.10))
-        set(WAYLAND_1_10_OR_GREATER ON)
-    endif ()
+    add_library(cogplatform-fdo MODULE platform/cog-platform-fdo.c)
 
-    pkg_check_modules(COGPLATFORM_FDO_DEPS REQUIRED wpe-webkit-1.0>=2.24.0 wpebackend-fdo-1.0>=1.3.1 egl xkbcommon)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -222,110 +222,69 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cogcore.pc
...
-        )
-
-        list(APPEND COGPLATFORM_FDO_SOURCES
-            ${CMAKE_BINARY_DIR}/weston/weston-direct-display-protocol.c
-            ${CMAKE_BINARY_DIR}/weston/weston-content-protection-protocol.c
-        )
-        list(APPEND COGPLATFORM_FDO_INCLUDE_DIRS ${CMAKE_BINARY_DIR}/weston)
-        set(ENABLE_WESTON_DIRECT_DISPLAY 1)
+        pkg_check_modules(LIBWESTON REQUIRED libweston-8-protocols)
+        pkg_get_variable(LIBWESTON_PKG_DATA_DIR libweston-8-protocols "pkgdatadir")
```
zdobersek: Used to be `libweston-9-protocols`?

aperezdc: Indeed, good catch!

I have pushed an update that brings back the :nine:

aperezdc:
@zdobersek Do you think we could merge this now? Thanks in advance!

aperezdc:
Cherry-picked into the `cog-0.6` branch as commit 43d87c71e3e504e21b620d78e6268e1c9cf644b1, thanks!

PR #190, opened 2020-03-18
Options menu support

zdobersek:


platform/cog-platform-fdo.c
```diff
@@ -427,6 +482,38 @@ static const struct xdg_toplevel_listener xdg_toplevel_listener = {
...
+{
+    if (popup_data.popup_menu == NULL)
+        return;
+
+    if (!popup_data.configured) {
+        popup_data.configured = true;
+
+        struct wl_buffer *buffer = cog_popup_menu_get_buffer (popup_data.popup_menu);
+        wl_surface_attach (popup_data.wl_surface, buffer, 0, 0);
+        wl_surface_damage (popup_data.wl_surface, 0, 0, INT_MAX, INT_MAX);
```
aperezdc: Neat, I didn't know that one can pass sizes bigger than the actual surface size. We should pass `INT32_MAX` here instead of `INT_MAX`, because the parameter is has type `int32_t`, though.

platform/cog-platform-fdo.c
```diff
@@ -1094,6 +1207,16 @@ touch_on_down (void *data,
             &raw_event,
             sizeof (struct wpe_input_touch_event_raw));
 
+    if (popup_data.wl_surface) {
+        if (wl_data.touch.surface == popup_data.wl_surface) {
+            cog_popup_menu_handle_event (popup_data.popup_menu,
+                                         1, raw_event.x, raw_event.y);
```
aperezdc: What's does `1` passed to the `state` parameter do here? It would be good to have a symbolic value, like `COG_POPUP_MENU_STATE_<foo>`, or if it is is a boolean flag, use something like:

```c
const bool popup_state_foo = true;
cog_popup_menu_handle_event (…, popup_state_foo, …);
```

platform/cog-popup-menu-fdo.c
```diff
@@ -0,0 +1,345 @@
...
+
+#include "os-compatibility.h"
+#include <cairo.h>
+#include <sys/mman.h>
+#include <unistd.h>
+#include <wpe/webkit.h>
+
+#include <stdio.h>
+
+typedef struct _CogPopupMenu {
```
aperezdc: Nice to see much of the popup menu logic self-contained :heart_eyes:

platform/cog-platform-fdo.c
```diff
@@ -1751,6 +1892,107 @@ destroy_window (void)
 #endif
 }
 
+static void
+create_popup (WebKitOptionMenu *option_menu)
```
aperezdc: There is quite some churn going on in this function, and I wonder whether it might be possible to move it over to `cog-popup-menu-fdo.c` somehow. Anyway, we can leave it here for now and take care of that later on for the `devel/multiple-views` branch, which has some abstraction utility code for handling Wayland “windows”—no need to do that now.

platform/cog-platform-fdo.c
zdobersek: This means that the 'pressed' event is being emitted.

platform/cog-platform-fdo.c
zdobersek: Not sure pushing the Wayland shell logic into `cog-popup-menu-fdo.c` is beneficial. I would rather (in the future, pos-multiview breakage) prefer that the specific shell logic is all combined in separate implementation files.

CMakeLists.txt
```diff
@@ -258,6 +263,29 @@ if (COG_PLATFORM_FDO AND NOT COG_USE_WEBKITGTK)
...
+        list(APPEND COGPLATFORM_FDO_LDFLAGS
+            ${CAIRO_LDFLAGS}
+        )
+
+        list(APPEND COGPLATFORM_FDO_SOURCES
+            platform/cog-popup-menu-fdo.c
+            platform/os-compatibility.c
+        )
+
+        add_definitions(-DCOG_FDO_WIDGETS_SUPPORTED=1)
```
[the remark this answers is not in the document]
zdobersek: This, along with other CMakeLists.txt changes you're proposing, would require moving around the cogplatform-fdo target declaration. Let's not do that in this PR.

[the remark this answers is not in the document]
aperezdc: Sure, we can move around CMake stuff later.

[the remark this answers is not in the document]
zdobersek: Actually we have to do it now -- rebasing upon master brings in changes that make this both possible and necessary.

PR #194, opened 2020-03-26
fdo: Load and set a default pointer cursor

### LINKED ISSUE #191: fdo: Wayland cursor surface is never configured
Hi, 
    I'm trying to solve the subjected problem but nothing get success till now. I guess it maybe caused by some bugs in cog. The problem can be reproduced by following steps,
    
    1. launch weston-terminal from debug console.
    2. mouse click in terminal  window any where.
    3. close weston-terminal.

    It would be appreciated if Igalia support team can help.

Thanks

aperezdc:
Load the default cursor theme and assign a surface with the left arrow image as cursor for the seat's pointer. This is a basic implementation which uses always the same cursor, even in HiDPI mode, so in that case the cursor will be tiny.


aperezdc:
Cherry-picked into the `cog-0.6` branch as commit aafd77a30e6fdd0cab4514ef3dd4fb938e01d7b1.

PR #199, opened 2020-04-02
cog-launcher: Move auto ptr declation before goto statement

kraj:
This fixes error like

/mnt/b/yoe/build/tmp/work/cortexa7t2hf-neon-vfpv4-yoe-linux-gnueabi/cog/0.5.1-r0/cog-0.5.1/core/cog-launcher.c:529:9: error: cannot jump from this goto statement to its label
goto bad_format;
^
/mnt/b/yoe/build/tmp/work/cortexa7t2hf-neon-vfpv4-yoe-linux-gnueabi/cog/0.5.1-r0/cog-0.5.1/core/cog-launcher.c:531:27: note: jump bypasses initialization of variable with __attribute__((cleanup))
g_autoptr(SoupCookie) cookie = soup_cookie_parse (contents, NULL);
^

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/core/cog-launcher.c
+++ b/core/cog-launcher.c
@@ -505,6 +505,7 @@ option_entry_parse_cookie_add (const char          *option G_GNUC_UNUSED,
                                GError             **error G_GNUC_UNUSED)
 {
     g_autoptr(GMainLoop) loop = NULL;
+    g_autoptr(SoupCookie) cookie = NULL;
     g_autofree char *domain = g_strdup (value);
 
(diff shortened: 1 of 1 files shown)


clopez:
This comes from https://github.com/Igalia/meta-webkit/pull/163
It seems clang warns about this problem but GCC not.
Pretty interesting. Here is the GCC bug about this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91951

PR #200, opened 2020-04-02
cmake: Fix passing -DWAYLAND_PROTOCOLS= on command line

JoshuaWatt:
The optimization to check if WAYLAND_PROTOCOLS was set on the cmake
comamnd line was skipping the function implementations, making cmake
fail if it was used. The simple fix is to define the functions, then
bail of out WAYLAND_PROTOCOLS is already defined.

Signed-off-by: Joshua Watt <Joshua.Watt@garmin.com>


aperezdc:
Cherry-picked into the `cog-0.6` branch as commit 28f62e82d8d73d3984ed281ff031b3ff74a415a8 ⛏️

PR #202, opened 2020-04-10
Missing COG_IM_API_SUPPORTED definitions at compile-time

zdobersek:
Since the COG_IM_API_SUPPORTED and other similar macro definitions
are not applied globally any more, a few code blocks in the cog
and cogcore targets were not compiled even when they should've been.


aperezdc:
Cherry-picked into the `cog-0.6` branch as commit 79dbb74ed711f6c7897750e0694c22d51d4bdf05.

PR #203, opened 2020-04-12
fdo: Avoid repeatedly setting opaque regions

aperezdc:
Set only the opaque region for the `wl_surface` used as output the first time a `wl_buffer` is attached, and only once after the surface has been reconfigured, instead of doing it woth every frame.

This improves graphics intensive benchmarks. For example, with this change MotionMark results consistently improve between 6 and 7 points on a laptop with a Core i7-6600U at 2.6GHz and an Intel HD Graphics 520 GPU.


philn:
Nice patch :)

aperezdc:
Thanks @philn :smile:

PR #208, opened 2020-05-28
feat: user defined video mode selection

kvasdopil:
Adding an env variable to specify preferred video mode for DRM platform.

Usage: 
```
export COG_PLATFORM_DRM_VIDEO_MODE=1920x1080
cog --platform drm http://google.com
```


WebReflection:
this PR is awesome, any reason is still on hold? the change looks simple enough to pass through, isn't it?

PR #209, opened 2020-06-03
cog: Load target URI instead of ignoring window.open()

aperezdc:
Load the URI passed as target to `window.open()` instead of leaving the signal unhandled, which results in no navigation at all. For now the same `WebKitWebView` is used because multiple views are not supported yet, but at least this makes certain websites usable.


aperezdc:
Cherry picked into the `cog-0.6` branch as commit e67365b06ea1c7af8536a53766ae2863d36eea80.

PR #217, opened 2020-08-18
pwl: Support multiple wl_seats

psaavedra:
All the resources related to an individual wl_seat are encapsulated into a PwlSeat struct. A PwlDisplay contains a list of PwlSeat each one managing listeners and resources separately.

Signed-off-by: Pablo Saavedra <psaavedra@igalia.com>

### DIFF
--- a/platform/pwl.c
+++ b/platform/pwl.c
@@ -59,24 +59,18 @@ typedef struct {
 } PwlOutput;
 
 
-struct _PwlDisplay {
-    struct wl_display *display;
-    struct wl_registry *registry;
-    struct wl_compositor *compositor;
+typedef struct {
+    PwlDisplay         *display;  /* Reference to the PwlDisplay*/
 
-    struct wl_seat *seat;
-    uint32_t        seat_version;
+    struct wl_seat     *seat;
+    uint32_t            seat_name;
+    uint32_t            seat_version;
 
-    EGLDisplay                               egl_display;
-    PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL     egl_createWaylandBufferFromImageWL;
-    EGLConfig                                egl_config;               /* Lazy init. */
-    EGLContext                               egl_context;              /* Lazy init. */
+    struct wl_pointer  *pointer;
+    PwlWindow          *pointer_target;  /* Current target of pointer events. */
 
-    struct wl_pointer *pointer;
-    PwlWindow *pointer_target;  /* Current target of pointer events. */
-
-    struct wl_touch *touch;
-    PwlWindow *touch_target;  /* Current target of touch events. */
+    struct wl_touch    *touch;
+    PwlWindow          *touch_target;  /* Current target of touch events. */
 
     PwlXKBData          xkb_data;
     struct wl_keyboard *keyboard;
@@ -87,6 +81,20 @@ struct _PwlDisplay {
     uint32_t            keyboard_repeat_key;
     uint32_t            keyboard_repeat_state;
     GSource            *keyboard_repeat_source;
+
+    struct wl_list link;
+} PwlSeat;
+
+
+struct _PwlDisplay {
+    struct wl_display *display;
+    struct wl_registry *registry;
+    struct wl_compositor *compositor;
+
+    EGLDisplay                               egl_display;
+    PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL     egl_createWaylandBufferFromImageWL;
+    EGLConfig                                egl_config;               /* Lazy init. */
+    EGLContext                               egl_context;              /* Lazy init. */
 
     struct wl_shell *shell;
 
@@ -107,6 +115,7 @@ struct _PwlDisplay {
 #endif /* HAVE_IVI_APPLICATION */
 
     struct wl_list outputs; /* wl_list<PwlOutput> */
+    struct wl_list seats; /* wl_list<PwlSeat> */
 
     char *application_id;
     char *window_title;
@@ -276,15 +285,15 @@ pointer_on_enter (void* data,
                   wl_fixed_t fixed_x,
                   wl_fixed_t fixed_y)
 {
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *seat = data;
+    if (pointer != seat->pointer) {
         g_critical ("%s: Got pointer %p, expected %p.",
-                    G_STRFUNC, pointer, display->pointer);
+                    G_STRFUNC, pointer, seat->pointer);
         return;
     }
 
     PwlWindow *window = wl_surface_get_user_data (surface);
-    display->pointer_target = window;
+    seat->pointer_target = window;
 
     window->focus |= PWL_FOCUS_MOUSE;
     if (window->focus_change_callback) {
@@ -324,33 +333,33 @@ pointer_on_leave (void              *data,
                   uint32_t           serial,
                   struct wl_surface *surface)
 {
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *seat = data;
+    if (pointer != seat->pointer) {
         g_critical ("%s: Got pointer %p, expected %p.",
-                    G_STRFUNC, pointer, display->pointer);
+                    G_STRFUNC, pointer, seat->pointer);
         return;
     }
 
-    if (!display->pointer_target) {
+    if (!seat->pointer_target) {
         g_critical ("%s: No current pointer event target!", G_STRFUNC);
         return;
     }
 
-    if (surface != display->pointer_target->wl_surface) {
+    if (surface != seat->pointer_target->wl_surface) {
         g_critical ("%s: Got leave for surface %p, but current window %p "
                     "has surface %p.", G_STRFUNC, surface,
-                    display->pointer_target,
-                    display->pointer_target->wl_surface);
+                    seat->pointer_target,
+                    seat->pointer_target->wl_surface);
         return;
     }
 
 #ifdef WL_POINTER_FRAME_SINCE_VERSION
-    if (display->seat_version < WL_POINTER_FRAME_SINCE_VERSION) {
-        window_dispatch_axis_event (display->pointer_target);
+    if (seat->seat_version < WL_POINTER_FRAME_SINCE_VERSION) {
+        window_dispatch_axis_event (seat->pointer_target);
     }
 #endif /* WL_POINTER_FRAME_SINCE_VERSION */
 
-    PwlWindow *window = display->pointer_target;
+    PwlWindow *window = seat->pointer_target;
 
     window->focus &= ~PWL_FOCUS_MOUSE;
     if (window->focus_change_callback) {
@@ -360,7 +369,7 @@ pointer_on_leave (void              *data,
     }
 
     g_debug ("%s: Pointer left surface %p, window %p.",
-             G_STRFUNC, surface, display->pointer_target);
+             G_STRFUNC, surface, seat->pointer_target);
 }
 
 
@@ -371,14 +380,14 @@ pointer_on_motion (void              *data,
                    wl_fixed_t         x,
                    wl_fixed_t         y)
 {
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *seat = data;
+    if (pointer != seat->pointer) {
         g_critical ("%s: Got pointer %p, expected %p.",
-                    G_STRFUNC, pointer, display->pointer);
+                    G_STRFUNC, pointer, seat->pointer);
         return;
     }
 
-    PwlWindow *window = display->pointer_target;
+    PwlWindow *window = seat->pointer_target;
     if (!window) {
         g_critical ("%s: No current pointer event target!", G_STRFUNC);
         return;
@@ -403,14 +412,14 @@ pointer_on_button (void              *data,
                    uint32_t           button,
                    uint32_t           state)
 {
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *seat = data;
+    if (pointer != seat->pointer) {
         g_critical ("%s: Got pointer %p, expected %p.",
-                    G_STRFUNC, pointer, display->pointer);
+                    G_STRFUNC, pointer, seat->pointer);
         return;
     }
 
-    PwlWindow *window = display->pointer_target;
+    PwlWindow *window = seat->pointer_target;
     if (!window) {
         g_critical ("%s: No current pointer event target!", G_STRFUNC);
         return;
@@ -441,14 +450,14 @@ pointer_on_axis (void              *data,
     TRACE ("%s: pointer @ %p, axis=%" PRIu32 ", value=%g.",
            G_STRFUNC, pointer, axis, wl_fixed_to_double (value));
 
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *seat = data;
+    if (pointer != seat->pointer) {
         g_critical ("%s: Got pointer %p, expected %p.",
-                    G_STRFUNC, pointer, display->pointer);
+                    G_STRFUNC, pointer, seat->pointer);
         return;
     }
 
-    PwlWindow *window = display->pointer_target;
+    PwlWindow *window = seat->pointer_target;
     if (!window) {
         g_critical ("%s: No current pointer event target!", G_STRFUNC);
         return;
@@ -468,7 +477,7 @@ pointer_on_axis (void              *data,
     }
 
 #ifdef WL_POINTER_FRAME_SINCE_VERSION
-    if (display->seat_version < WL_POINTER_FRAME_SINCE_VERSION)
+    if (seat->seat_version < WL_POINTER_FRAME_SINCE_VERSION)
 #endif /* !WL_POINTER_FRAME_SINCE_VERSION */
         window_dispatch_axis_event (window);
 }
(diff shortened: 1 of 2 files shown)


platform/pwl.c
```diff
@@ -276,15 +284,15 @@ pointer_on_enter (void* data,
                   wl_fixed_t fixed_x,
                   wl_fixed_t fixed_y)
 {
-    PwlDisplay *display = data;
-    if (pointer != display->pointer) {
+    PwlSeat *pwl_seat = data;
```
aperezdc, line 287: There is no need for the `pwl_` prefix in this variable, you can just call it `seat`:

```c
PwlSeat *seat = data;
```

platform/pwl.c
```diff
@@ -816,68 +824,68 @@ window_dispatch_key_event (PwlWindow  *self,
 
 
 static gboolean
-display_on_keyboard_repeat_timeout (PwlDisplay *self)
+display_on_keyboard_repeat_timeout (PwlSeat *pwl_seat)
```
aperezdc, line 827: This function acts on a `PwlSeat` so I would rename it as `seat_on_keyboard_repeat_timeout` and continue calling the first parameter `self`, which makes the refactor do less changes:

```c
static gboolean
seat_on_kayboard_repeat_timeout (PwlSeat *self)
{
    /* No need to rename s/self/seat/ inside the function. */
}
```

platform/pwl.c
```diff
@@ -816,68 +824,68 @@ window_dispatch_key_event (PwlWindow  *self,
...
-    g_source_destroy (self->keyboard_repeat_source);
-    g_clear_pointer (&self->keyboard_repeat_source, g_source_unref);
+    g_source_destroy (pwl_seat->keyboard_repeat_source);
+    g_clear_pointer (&pwl_seat->keyboard_repeat_source, g_source_unref);
 }
 
 
 static void
-display_start_keyboard_repeat (PwlDisplay *self,
+display_start_keyboard_repeat (PwlSeat    *pwl_seat,
```
aperezdc, line 860: Same here, I would rename the function to:

```c
static void
seat_start_keyboard_repeat (PwlSeat *self,
                            /* the rest stays the same... */
```

platform/pwl.h
```diff
@@ -125,7 +125,7 @@ struct wl_buffer*
             pwl_display_egl_create_buffer_from_image (const PwlDisplay*, EGLImage);
 bool        pwl_display_egl_has_broken_buffer_from_image (const PwlDisplay*);
 
-PwlXKBData* pwl_display_xkb_get_data (PwlDisplay*);
+PwlXKBData* pwl_display_xkb_get_data (PwlSeat*);
```
aperezdc, line 128: This operates on a `PwlSeat`, therefore it should be renamed to `pwl_seat_xkb_get_data()`.

platform/pwl.c
```diff
@@ -1960,7 +1986,7 @@ pwl_display_egl_create_buffer_from_image (const PwlDisplay *self,
 
 
 PwlXKBData*
-pwl_display_xkb_get_data (PwlDisplay *self)
+pwl_display_xkb_get_data (PwlSeat *self)
```
aperezdc, line 1989: This operates on a `PwlSeat`, therefore it should be renamed to `pwl_seat_xkb_get_data()`.

platform/pwl.c
```diff
@@ -1393,10 +1417,12 @@ pwl_display_destroy (PwlDisplay *self)
...
-        g_clear_pointer (&self->xkb_data.state, xkb_state_unref);
-        g_clear_pointer (&self->xkb_data.compose_state, xkb_compose_state_unref);
-        g_clear_pointer (&self->xkb_data.keymap, xkb_keymap_unref);
-        g_clear_pointer (&self->xkb_data.context, xkb_context_unref);
+        wl_list_for_each_safe (item, tmp, &self->seats, link) {
+            g_clear_pointer (&item->xkb_data.state, xkb_state_unref);
+            g_clear_pointer (&item->xkb_data.compose_state, xkb_compose_state_unref);
+            g_clear_pointer (&item->xkb_data.keymap, xkb_keymap_unref);
+            g_clear_pointer (&item->xkb_data.context, xkb_context_unref);
+        }
```
aperezdc, line 1425: You have to unlink each seat item from the list, and free the memory used by the corresponding `PwlSeat` value. You may want to add a `pwl_seat_destroy()` function which takes care of destroying/unreffing the members of `PwlSeat` and the deallocates the `struct` itself. That way the only work to do here is iterate the `wl_list` unlinking elements and calling `pwl_seat_destroy()` on them.

platform/pwl.c
```diff
@@ -1393,10 +1417,12 @@ pwl_display_destroy (PwlDisplay *self)
         g_clear_pointer (&self->application_id, g_free);
         g_clear_pointer (&self->window_title, g_free);
 
-        g_clear_pointer (&self->xkb_data.state, xkb_state_unref);
-        g_clear_pointer (&self->xkb_data.compose_state, xkb_compose_state_unref);
-        g_clear_pointer (&self->xkb_data.keymap, xkb_keymap_unref);
-        g_clear_pointer (&self->xkb_data.context, xkb_context_unref);
+        wl_list_for_each_safe (item, tmp, &self->seats, link) {
```
aperezdc, line 1420: I don't like the same loop as above (line 1407) being repeated here. There is no reason why more than one loop should be needed.

platform/pwl.c
```diff
@@ -1358,8 +1375,13 @@ pwl_display_destroy (PwlDisplay *self)
     pwl_display_egl_deinit (self);
 
     if (self->display) {
-        display_stop_keyboard_repeat (self);
+        PwlSeat *item, *tmp;
 
+        wl_list_for_each_safe (item, tmp, &self->seats, link) {
+            g_debug ("Seat @ %p: Removed.",
+                     item->seat);
```
aperezdc, line 1382: If you add a `pwl_seat_destroy()` helper function as suggested below, you should move this `g_debug()` call into it, and remove it from here. Also, this is stopping the keyboard repeat timer, not removing the seat, so the debug message is misleading.

platform/pwl.c
```diff
@@ -1358,8 +1375,13 @@ pwl_display_destroy (PwlDisplay *self)
     pwl_display_egl_deinit (self);
 
     if (self->display) {
-        display_stop_keyboard_repeat (self);
+        PwlSeat *item, *tmp;
 
+        wl_list_for_each_safe (item, tmp, &self->seats, link) {
```
aperezdc, line 1380: Here it's fine to use the `wl_list_for_each()` macro. You only need to use `wl_list_for_each_safe()` if you are planning to modify the list while iterating over it.

platform/pwl.c
```diff
@@ -1218,74 +1274,35 @@ static void
...
         }
     }
+    PwlSeat *pwl_seat, *tmp_seat;
+    wl_list_for_each_safe (pwl_seat, tmp_seat, &display->seats, link) {
+        g_debug ("Seat @ %p: Removed.",
+                 pwl_seat->seat);
+        g_clear_pointer (&pwl_seat->seat, wl_seat_release);
+        wl_list_remove (&pwl_seat->link);
+        g_slice_free (PwlSeat, pwl_seat);
+    }
```
aperezdc, line 1295: This loop is removing _all_ the `PwlSeat` instances from the list! What is worse: it's done unconditionally, so if an output is removed, all the seats will be destroyed as well.

You need to compare the `name` integer to check whether the object being destroyed is one of the elements contained in the list of seats. Check how the same is done right above when iterating the list of outputs. You may need to add a `seat_name` integer to `PwlSeat` to store the object identifier when the seat is created, so later on it can be used here for the comparison.

platform/pwl.c
```diff
@@ -1103,6 +1111,53 @@ zxdg_shell_v6_ping (void                 *data G_GNUC_UNUSED,
 #endif /* HAVE_XDG_SHELL_UNSTABLE_V6 */
 
 
+static void
+xkb_data_init(PwlSeat *pwl_seat)
```
aperezdc, line 1115: You may want to rename this as `pwl_seat_create()`, and move allocation here, like this:

```c
static PwlSeat*
pwl_seat_create(struct wl_seat *seat)
{
    PwlSeat *self = /* allocate */;

    self->seat = seat;
    /* initialize more values */
    /* initialize the XKB values, too */
    /* attach the seat listener */
    return self;
}
```

Then later in `registry_global()` you only need to call `pwl_seat_create()` and add the new value to the list of seats.

platform/pwl.c
aperezdc: Ah, wait, I see why you have done this: the `struct wl_seat` in some places is called `seat`… Because the code deals with a `PwlSeat` more often, I would call that `seat` and the `struct wl_seat` I would call it `wl_seat` when needed.

psaavedra: I just pushed another version where I replace struct wl_seat *seat by  ` struct wl_seat *w_seat`. Probably your suggestion it's much better than mine.

platform/pwl.c
psaavedra: done.

platform/pwl.c
psaavedra: done.

platform/pwl.h
psaavedra: done.

platform/pwl.c
psaavedra: done.

platform/pwl.c
psaavedra: Code simplified and encapsulated in the `pwl_seat_destroy()`

platform/pwl.c
psaavedra: Fixed.

platform/pwl.c
psaavedra: Code simplified and encapsulated in the `pwl_seat_destroy()`

platform/pwl.c
psaavedra: Oh, now i understand the reason of keep the `name` value as part of the struct. Thanks for the review.

Code simplified and encapsulated in the `pwl_seat_destroy()`

platform/pwl.c
psaavedra: Done.

platform/pwl.c
psaavedra: Finally, the PR updated with `s/w_seat/wl_seat/g`: https://github.com/Igalia/cog/commit/cf3f303efeaadee9358f158175ae23e33ae20d50

psaavedra:
Please, dont merge this until my confirmation.

platform/pwl.c
```diff
@@ -699,17 +708,17 @@ keyboard_on_keymap (void *data,
...
+    g_clear_pointer (&seat->xkb_data.keymap, xkb_keymap_unref);
 
-    display->xkb_data.keymap = g_steal_pointer (&keymap);
-    display->xkb_data.state = g_steal_pointer (&state);
+    seat->xkb_data.keymap = g_steal_pointer (&keymap);
+    seat->xkb_data.state = g_steal_pointer (&state);
 
-    display->xkb_data.indexes.control = xkb_keymap_mod_get_index (display->xkb_data.keymap,
+    seat->xkb_data.indexes.control = xkb_keymap_mod_get_index (seat->xkb_data.keymap,
                                                                   XKB_MOD_NAME_CTRL);
```
aperezdc, line 718: Plese reindent the dangling lines so they are aligned to the opening parenthesis:

```c
xkb_keymap_mod_get_index (seat->xkb_data.keymap,
                          XKB_MOD_NAME_CTRL);
           // align here ↑
```

Same thing in the lines below.

platform/pwl.c
```diff
@@ -849,30 +861,30 @@ display_stop_keyboard_repeat (PwlDisplay *self)
 
 
 static void
-display_start_keyboard_repeat (PwlDisplay *self,
+seat_start_keyboard_repeat (PwlSeat    *self,
                                uint32_t    timestamp,
                                uint32_t    key,
                                uint32_t    state)
```
aperezdc, line 867: Please realign the rest of function arguments to the opening parenthesis here, too.

platform/pwl.c
```diff
@@ -1103,6 +1115,93 @@ zxdg_shell_v6_ping (void                 *data G_GNUC_UNUSED,
 #endif /* HAVE_XDG_SHELL_UNSTABLE_V6 */
 
 
+static PwlSeat*
+pwl_seat_create(struct wl_seat *seat, uint32_t name, uint32_t version, PwlDisplay *display)
```
aperezdc, line 1119: Add an space between the function name and the `(` parenthesis, and list each argument in its own line aligned to it. Also, I see that the function adds the seat to the `PwlDisplay` already, so a better name would be `display_add_seat()`, and there is no need to return the `PwlSeat`:

```c
static void
display_add_seat (PwlDisplay     *display,
                  struct wl_seat *seat,
                  uint32_t        name,
                  uint32_t        version)
```

platform/pwl.c
```diff
@@ -1103,6 +1115,93 @@ zxdg_shell_v6_ping (void                 *data G_GNUC_UNUSED,
...
+{
+    PwlSeat *self = g_slice_new0 (PwlSeat);
+    self->seat = seat;
+    self->seat_name = name;
+    self->seat_version = version;
+
```
aperezdc, line 1125: This empty line is not needed, please remove it.

platform/pwl.c
```diff
@@ -1103,6 +1115,93 @@ zxdg_shell_v6_ping (void                 *data G_GNUC_UNUSED,
...
+    self->display = display;
+    if (display)
```
aperezdc, line 1127: I think it does not make sense to create a `PwlSeat` which does not have an associated display. I think it's better to add `g_assert (display)` at the beginning of the function, and remove the condition here.

platform/pwl.c
```diff
@@ -1103,6 +1115,93 @@ zxdg_shell_v6_ping (void                 *data G_GNUC_UNUSED,
...
+                                                               XKB_MOD_NAME_CTRL);
+    self->xkb_data.indexes.alt = xkb_keymap_mod_get_index (self->xkb_data.keymap,
+                                                           XKB_MOD_NAME_ALT);
+    self->xkb_data.indexes.shift = xkb_keymap_mod_get_index (self->xkb_data.keymap,
+                                                             XKB_MOD_NAME_SHIFT);
+    return self;
+}
+
+static void
+pwl_seat_destroy(PwlSeat *self)
```
aperezdc, line 1185: Missing space between the function name and `(`.

platform/pwl.c
```diff
@@ -1168,22 +1267,13 @@ registry_global (void               *data,
...
-                .capabilities = display_on_seat_capabilities,
-                .name = display_on_seat_name,
-            };
-            wl_seat_add_listener (display->seat, &seat_listener, display);
-        }
+        uint32_t seat_version = MIN (version, WL_POINTER_FRAME_SINCE_VERSION);
+        struct wl_seat *wl_seat = wl_registry_bind (registry,
+                                                   name,
+                                                   &wl_seat_interface,
+                                                   seat_version);
```
aperezdc, line 1274: The indentation here is missing one space.

platform/pwl.c
```diff
@@ -1218,74 +1308,37 @@ static void
...
         }
     }
+    PwlSeat *seat, *tmp_seat;
+    wl_list_for_each_safe (seat, tmp_seat, &display->seats, link) {
+        if (seat->seat_name == name) {
+            g_debug ("Seat #%"PRIi32" @ %p: Removed.",
+                     seat->seat_name, seat->seat);
+            g_clear_pointer (&seat->seat, wl_seat_release);
+            wl_list_remove (&seat->link);
+            g_slice_free (PwlSeat, seat);
```
aperezdc, line 1329: Just call `pwl_seat_destroy()` here. Also add a `break` because once we have found the item we want to remove from the list we don't need to continue iterating over it.

aperezdc (changes requested):
This is looking very good, and I think it is almost ready for merging. Most of my second round of comments are small coding style issues—the logic looks good.

Thanks for the batch of improvements! :muscle:

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: removed

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: done

platform/pwl.c
psaavedra: done

psaavedra (commented):
> Please, dont merge this until my confirmation.

Ok. I made my own tests in local. The patch is still working fine after the changes so nothing to complain from my side.

platform/pwl.c
```diff
@@ -1960,7 +2006,7 @@ pwl_display_egl_create_buffer_from_image (const PwlDisplay *self,
 
 
 PwlXKBData*
-pwl_display_xkb_get_data (PwlDisplay *self)
+pwl_seat_xkb_get_data (PwlSeat *self)
```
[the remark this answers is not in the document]
psaavedra: Just uploaded a new version of the patch (https://github.com/Igalia/cog/commit/cc8f72cc15023e2e5e09bf428b5864e8951b07e7) addressing this comment.

platform/pwl.h
```diff
@@ -105,7 +105,7 @@ typedef struct {
 } PwlXKBData;
 
 typedef struct _PwlDisplay PwlDisplay;
-
+typedef struct _PwlSeat PwlSeat;
```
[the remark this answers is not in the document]
psaavedra: ditto.

platform/pwl.c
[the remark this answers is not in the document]
aperezdc: Great, thanks a ton!

aperezdc (approved):
@psaavedra: Nice job getting this all ready, let's merge it :muscle:

PR #218, opened 2020-08-25
cog: Add a runtime option to ignore TLS certificate errors

aperezdc:
By default this option is disabled and an error page will be displayed when the page load fails due to a TLS error. However, by enabling this option the page load can resume and thus the TLS error is ignored. Use this option carefully and consider the security implications.

Based on #138, initially written by Philippe Normand (thanks @philn!)

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -53,6 +53,7 @@ static struct {
         enum webprocess_fail_action action_id;
     } on_failure;
     char *web_extensions_dir;
+    gboolean ignore_tls_errors;
 } s_options = {
     .scale_factor = 1.0,
 #if HAVE_DEVICE_SCALING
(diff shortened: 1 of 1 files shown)


core/cog-launcher.c
```diff
@@ -117,12 +117,39 @@ on_permission_request (G_GNUC_UNUSED WebKitWebView *web_view,
...
+                                GTlsCertificateFlags errors,
+                                void                *user_data)
+{
+    CogShell *shell = COG_SHELL (user_data);
+    WebKitWebContext *context = webkit_web_view_get_context (web_view);
+
+    if (cog_shell_get_ignore_tls_errors (shell)) {
+        g_autoptr(SoupURI) uri = soup_uri_new (failing_uri);
+        g_message ("<%s> Ignoring TLS errors", failing_uri);
+        webkit_web_context_allow_tls_certificate_for_host (context,
```
carlosgcampos, line 133: This is thought to add exceptions when the global policy is to fail in case of TLS errors. This patch is adding a global policy, so I trhink it should be easier and more efficient to use webkit_web_context_set_tls_errors_policy()

aperezdc: I have updated the PR to set this web context option instead, thanks for the suggestion!

carlosgcampos (approved):
Much simpler!

PR #222, opened 2020-09-03
core: unset properties after creating the view object

csaavedra:
During collection of the GValues used to initialize the
View object, the reference count of GObject properties
will be increased. Therefore it's necessary to unset the
GValues once done with them to avoid leaking a reference.

For simplicity, let G_VALUE_COLLECT_INIT copy all the values
and just unset them all at once when done. This fixes any
potential leak with GObject-properties.


aperezdc (approved):
Nice work finding this subtle issue :muscle:!

PR #223, opened 2020-09-16
fdo: Handle sizing with fullscreen shell

JoshuaWatt:
Reworks the fullscreen shell handling to respect the environment
variables COG_PLATFORM_FDO_VIEW_{WIDTH,HEIGHT} to set the backend
resolution, or COG_PLATFORM_FDO_VIEW_FULLSCREEN, which will make cog
resize to match the size of the largest connected output.

### DIFF
--- a/platform/cog-platform-fdo.c
+++ b/platform/cog-platform-fdo.c
@@ -97,13 +97,13 @@ typedef struct wl_buffer * (EGLAPIENTRYP PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL) (
 #endif
 
 
-#if HAVE_DEVICE_SCALING
 typedef struct output_metrics {
   struct wl_output *output;
   int32_t name;
   int32_t scale;
+  int32_t width;
+  int32_t height;
 } output_metrics;
-#endif /* HAVE_DEVICE_SCALING */
 
 static struct {
     struct wl_display *display;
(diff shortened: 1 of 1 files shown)


JoshuaWatt:
Ping?

zdobersek:
What happens when, hypothetically, fullscreen shell is used with multiple outputs with differing sizes? AFAIU these changes would always default to resizing the view to the dimensions of the biggest one, even if the Wayland client was on display on the smaller output.

JoshuaWatt:
> What happens when, hypothetically, fullscreen shell is used with multiple outputs with differing sizes? AFAIU these changes would always default to resizing the view to the dimensions of the biggest one, even if the Wayland client was on display on the smaller output.

That's correct. However, we have some sort of problem like that any way you look at it because we are allowing the compositor to choose an output instead of specifying one, so we wouldn't have any control anyway. The choice to choose the largest output was arbitrary, but it most cases I suspect this would cause the compositor to choose to display the content on the largest output.

We could of course specify an output when call zwp_fullscreen_shell_v1_present_surface, but I'm not sure what good that would really do, since I can't think of a sane way to allow the user to specify which output they want. If you are using the fullscreen shell with weston + cog, you are probably better off modifying the weston.ini so that there is only one active output if you really want to show cog on a specific screen.

FWIW, this isn't really changing the "undefined" behavior as to where the output will be displayed and whether it will be cropped and scaled  when multiple outputs are present. That behavior was always present and more or less undefined.

zdobersek:
OK, we can incorporate this change and address any special cases later.

aperezdc (approved):
I think that in the long run we may want as well to allow indicating to the compositor which output Cog prefers to be displayed on when maximized, so I have filed issue #228 for that.

For now let's merge this. Thanks for the patch, @JoshuaWatt!

PR #224, opened 2020-09-30
CogShell: several fixes during initialization

csaavedra:
1. Fail if the preferred module is not available.
2. Change a critical to warning as module unavailability is not a programming error.
3. Fill the GError in case of shell creation failure, also add a GError domain for this.
4.A few compilation warnings fixed.

### DIFF
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -127,7 +127,7 @@ cog_shell_class_init (CogShellClass *klass)
 }
 
 static void
-cog_shell_init (CogShell *shell)
+cog_shell_init (CogShell *shell G_GNUC_UNUSED)
 {
 }
 
@@ -144,7 +144,7 @@ is_construct_property (const char  *func,
                        GType        object_type,
                        const char  *propname,
                        GParamSpec  *pspec,
-                       unsigned     n_properties)
+                       unsigned     n_properties G_GNUC_UNUSED)
 {
     if (G_UNLIKELY (pspec == NULL)) {
         g_critical ("%s: object class '%s' has no property named '%s'",
@@ -329,7 +329,7 @@ cog_shell_get_name (CogShell *shell)
 
 static void
 shell_on_notify_view_focused (CogView    *view,
-                              GParamSpec *pspec,
+                              GParamSpec *pspec G_GNUC_UNUSED,
                               CogShell   *shell)
 {
     CogShellPrivate *priv = cog_shell_get_instance_private (shell);
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -173,6 +173,8 @@ string_array_contains (const char *haystack[],
     return FALSE;
 }
 
+G_DEFINE_QUARK (cog-shell-error-quark, cog_shell_error)
+
 static CogShell*
 cog_shell_new_internal (GError    **error,
                         const char *shell_name,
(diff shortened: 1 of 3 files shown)


core/cog-shell.c
```diff
@@ -127,7 +127,7 @@ cog_shell_class_init (CogShellClass *klass)
 }
 
 static void
-cog_shell_init (CogShell *shell)
+cog_shell_init (G_GNUC_UNUSED CogShell *shell)
```
aperezdc, line 130: Now that I see this it dawned on me that we do not place the `G_GNUC_UNUSED` annotations consistently in all the code base, sometimes they are before the type, sometimes after the variable name… most of them are _after_ the variable names, so I say let's go with that from now on. Could you move this one so it reads `CogShell *shell G_GNUC_UNUSED`?

core/cog-shell.c
```diff
@@ -140,11 +140,11 @@ cog_shell_get_view_class (CogShell *shell)
...
-is_construct_property (const char  *func,
-                       GType        object_type,
-                       const char  *propname,
-                       GParamSpec  *pspec,
-                       unsigned     n_properties)
+is_construct_property (const char            *func,
+                       GType                  object_type,
+                       const char            *propname,
+                       GParamSpec            *pspec,
+                       G_GNUC_UNUSED unsigned n_properties)
```
aperezdc, line 147: Same with this `G_GNUC_UNUSED` annotation.

core/cog-shell.c
```diff
@@ -173,6 +173,8 @@ string_array_contains (const char *haystack[],
     return FALSE;
 }
 
+G_DEFINE_QUARK(cog-shell-error-quark, cog_shell_error)
```
aperezdc, line 176: Small nit: a space is missing between the macro name and the `(` character.

Also: Ouch! We are also using an inconsistent style of the name passed as argument to the `G_DEFINE_QUARK` macro. This one I blame on myself, there's three different styles:

```
% grep -r G_DEFINE_QUARK core/ platform/
core/cog-directory-files-handler.c:G_DEFINE_QUARK (cog-directory-files-handler-data, cog_directory_files_handler);
core/cog-directory-files-handler.c:G_DEFINE_QUARK (CogDirectoryFilesHandlerError, cog_directory_files_handler_error)
platform/pwl.c:G_DEFINE_QUARK (com_igalia_Cog_PwlError, pwl_error)
```

But yours is already using what seems to be “the usual GLib style”, so let's keep it and try to harmonize the rest later on separately.

core/cog-shell.c
```diff
@@ -324,9 +328,9 @@ cog_shell_get_name (CogShell *shell)
 
 
 static void
-shell_on_notify_view_focused (CogView    *view,
-                              GParamSpec *pspec,
-                              CogShell   *shell)
+shell_on_notify_view_focused (CogView      *view,
+                              G_GNUC_UNUSED GParamSpec *pspec,
```
aperezdc, line 332: The `G_GNUC_UNUSED` could be moved after the variable name here, too.

core/cog-shell.h
```diff
@@ -15,6 +15,13 @@
 
 G_BEGIN_DECLS
 
+#define COG_SHELL_ERROR (cog_shell_error_quark ())
+
+typedef enum
+{
```
aperezdc, line 21: The opening brace goes next to the `typedef enum` in the Cog code style, let's move it up :arrow_up:

aperezdc (changes requested):
The added logic looks perfect. I have commented with a few code style nits, but other than we should merge this.

Thanks a lot for the patch!

aperezdc (approved):
Great, thanks for updating the patch :smiley:

PR #226, opened 2020-10-05
dir-handler: Allow usage of URI host as path component

### LINKED ISSUE #225: dir-handler: Support using the URI host as top-level subdirectory
It would be neat to add an option in `CogDirectoryFilesHandler` to use the hos part of the URI authority field as top-level subdirectory. This would allow pointing a handler to a directory, and treating each of its subdirectories as a separate security origin.

aperezdc:
Support using the URI host as the first path component, when the `CogDirectoryFilesHandler.use-host` property is set. The default value of the property is `FALSE` to make the new behaviour opt-in and avoid breaking existing code.

There is also a commit to turn the base path into a construct-only property (which is good practice for `GObject` construction parameters), and another one to remove an unused definition.

Closes #225

### DIFF
--- a/core/cog-directory-files-handler.c
+++ b/core/cog-directory-files-handler.c
@@ -14,7 +14,13 @@ struct _CogDirectoryFilesHandler {
     GFile  *base_path;
 };
 
+enum {
+    PROP_0,
+    PROP_BASE_PATH,
+    N_PROPERTIES,
+};
 
+static GParamSpec *s_properties[N_PROPERTIES] = { NULL, };
 
 static const char s_file_query_attributes[] =
     G_FILE_ATTRIBUTE_STANDARD_CONTENT_TYPE ","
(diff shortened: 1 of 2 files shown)


core/cog-directory-files-handler.c
```diff
@@ -145,26 +147,45 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
...
+            g_autofree char *uri_string = soup_uri_to_string (uri, FALSE);
+            g_autoptr(GError) error = g_error_new (G_FILE_ERROR,
+                                                   G_FILE_ERROR_INVAL,
+                                                   "No host in URI: %s",
+                                                   uri_string);
+            webkit_uri_scheme_request_finish_error (request, error);
+            return;
+        }
+        base_path = g_file_get_child (handler->base_path, host);
+    }
```
csaavedra, line 178: I would probably just add an else here and assign `base_path` the handler's base path, to avoid having the ?: comparison later on, keeping the refcounting in mind of course.

core/cog-directory-files-handler.c
```diff
@@ -145,26 +147,45 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
...
     /*
      * Skip trailing forward slashes: If an absolute child path would
      * be passed to g_file_get_child() it will ignore the base path.
      */
     while (path[0] == '/')
         ++path;
 
-    g_autoptr(GFile) file = g_file_get_child (handler->base_path, path);
+    g_autoptr(GFile) file =
+        g_file_get_child (base_path ? base_path : handler->base_path, path);
```
csaavedra, line 188: Then you wouldn't need this `?:` here.

csaavedra (approved):
Looks fine to me, the comments are just aesthetic and up to you.

core/cog-directory-files-handler.c
aperezdc: Fun thing: First I wrote it in the way you suggest, then I thought that I would change it to use `?:` below and avoid refcount churn :smile: 

I do agree that the code is easier to understand the way you suggest, so I'll change it back to avoid the `?:` before merging.

Thanks for the review!

PR #232, opened 2020-10-16
drm: fix keyboard typing with shift and other func keys

kytart:
Func keys like shift were completely ignored, which made it impossible to type capital letters or special characters.

I basically just copied the logic from wpebackend-rdk where it works as expected:
https://github.com/WebPlatformForEmbedded/WPEBackend-rdk/blob/8e35a98627361898059cc8212e5bc87993a57e0d/src/input/Libinput/LibinputServer.cpp#L84


aperezdc (approved):
This looks fine, thanks for the patch @kytart :building_construction:

PR #233, opened 2020-10-16
drm: optionally limit max video mode width/height

kytart:
Limit max video mode width/height so the selection of best video mode is limited by these boundaries. It's useful, for example, on Raspberry Pi 4 that supports 4K. If I connect it to 4K screen, it will by default automatically select 4K mode but in some cases that's not what I want. When there's a lot of elements on the screen, CSS animations, etc., performance seems to suffer in 4K so I'd prefer to just force it to be max Full HD.

### DIFF
--- a/platform/cog-platform-drm.c
+++ b/platform/cog-platform-drm.c
@@ -203,10 +203,17 @@ init_drm (void)
 
     const char* user_selected_mode = g_getenv("COG_PLATFORM_DRM_VIDEO_MODE");
 
-    const char* user_max_width_str = g_getenv("COG_PLATFORM_DRM_VIDEO_MAX_WIDTH");
-    const char* user_max_height_str = g_getenv("COG_PLATFORM_DRM_VIDEO_MAX_HEIGHT");
-    const int user_max_width = user_max_width_str ? atoi(user_max_width_str) : 0;
-    const int user_max_height = user_max_height_str ? atoi(user_max_height_str) : 0;
+    int user_max_width = 0;
+    int user_max_height = 0;
+    const char *user_mode_max = g_getenv("COG_PLATFORM_DRM_MODE_MAX");
+    if (user_mode_max) {
+        if (sscanf(user_mode_max, "%dx%d", &user_max_width, &user_max_height) != 2 ||
+            user_max_width < 0 || user_max_height < 0) {
+            fprintf(stderr, "invalid value for COG_PLATFORM_DRM_MODE_MAX\n");
+            user_max_width = 0;
+            user_max_height = 0;
+        }
+    }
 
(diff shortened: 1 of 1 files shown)


aperezdc:
We have support for specifying the desired output mode after https://github.com/Igalia/cog/pull/208 was merged, setting for example `COG_PLATFORM_DRM_VIDEO_MODE=1920x1080` in the environment. Wouldn't that be enough for you, @kytart?

I would prefer to keep the mode selection code simpler if possible, but I am open to let myself be convinced that there is a good use case for your proposed patch that cannot be solved with `COG_PLATFORM_DRM_VIDEO_MODE`.

kytart:
@aperezdc I know that `COG_PLATFORM_DRM_VIDEO_MODE` is available but my use case is slightly different. The default logic that iterates over all available modes and picks the one with the highest resolution is what I need because in my use case I'm using Raspberry Pi which can be connected to any type of screen with smaller or bigger resolution and I want it to automatically detect the appropriate resolution. But then there's the problem with Raspberry Pi 4 and performance in 4K.

So, long story short, I want to keep the smart detection but put a cap on it so it doesn't go higher than 1920x1080. Makes sense?

aperezdc:
> So, long story short, I want to keep the smart detection but put a cap on it so it doesn't go higher than 1920x1080. Makes sense?

@kytart: Yes, it does.

platform/cog-platform-drm.c
```diff
@@ -203,12 +203,24 @@ init_drm (void)
 
     const char* user_selected_mode = g_getenv("COG_PLATFORM_DRM_VIDEO_MODE");
 
+    const char* user_max_width_str = g_getenv("COG_PLATFORM_DRM_VIDEO_MAX_WIDTH");
+    const char* user_max_height_str = g_getenv("COG_PLATFORM_DRM_VIDEO_MAX_HEIGHT");
```
aperezdc, line 207: Could we use a single environment variable for this? Something like `COG_PLATFORM_DRM_MODE_MAX=1920x1080` would be nicer, instead of having to define two different variables for width and height. If one only wants to limit height, a value of `0x1080` could be used, or `1920x0` to limit only the width.

aperezdc (changes requested):
I think we can merge this feature, but I think it would be nicer to use only one environment variable. Could you take a look at my comment below? Thanks!

platform/cog-platform-drm.c
kytart: yeah I think that's also ok

kytart: @aperezdc I made the changes, you can check it out

aperezdc (approved):
Thanks for the update, @kytart — let's merge this! :muscle:

PR #235, opened 2020-10-20
Add initial X11-oriented platform module

zdobersek:
The module is very basic in its nature, resampling the exported buffers
into a separate GL scene and using a dumb repaint loop to present that
scene in an X11 window. libxcb is used as the library through which
the X11 protocol is handled.

Very basic input (mouse, keyboard) is supported. The X11 window is
closable, but not (yet) through any keyboard shortcut.

This is intended as a minimal X11-functionality baseline, with
additional attention provided as required.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,7 @@ set(COG_DBUS_OWN_USER "" CACHE STRING
 option(COG_USE_WEBKITGTK "Use WebKitGTK+ instead of WPEWebKit" OFF)
 option(COG_PLATFORM_FDO "Build the FDO platform module" ON)
 option(COG_PLATFORM_DRM "Build the DRM platform module" OFF)
+option(COG_PLATFORM_X11 "Build the DRM platform module" OFF)
 option(COG_BUILD_PROGRAMS "Build and install programs as well" ON)
 option(INSTALL_MAN_PAGES "Install the man(1) pages if COG_BUILD_PROGRAMS is enabled" ON)
 option(COG_WESTON_DIRECT_DISPLAY "Build direct display support for the FDO platform module" OFF)
@@ -345,5 +346,9 @@ if (COG_PLATFORM_DRM AND NOT COG_USE_WEBKITGTK)
         COMPONENT "runtime"
     )
 endif () # COG_PLATFORM_DRM AND NOT COG_USE_WEBKITGTK
+
+if (COG_PLATFORM_X11 AND NOT COG_USE_WEBKITGTK)
+    add_subdirectory(platform/x11)
+endif ()
 
 configure_file(core/cog-config.h.in cog-config.h @ONLY)
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -450,11 +450,11 @@ xcb_source_dispatch (GSource *base, GSourceFunc callback, gpointer user_data)
 {
     struct xcb_source *source = (struct xcb_source *) base;
     if (source->pfd.revents & (G_IO_ERR | G_IO_HUP))
-        return FALSE;
+        return G_SOURCE_REMOVE;
 
     xcb_process_events ();
     source->pfd.revents = 0;
-    return TRUE;
+    return G_SOURCE_CONTINUE;
 }
 
 static xcb_atom_t
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -460,18 +460,16 @@ xcb_source_dispatch (GSource *base, GSourceFunc callback, gpointer user_data)
 static xcb_atom_t
 get_atom (struct xcb_connection_t *connection, const char *name)
 {
-    xcb_intern_atom_cookie_t cookie;
-    xcb_intern_atom_reply_t *reply;
-    xcb_atom_t atom;
+    xcb_intern_atom_cookie_t cookie = xcb_intern_atom (connection, 0, strlen(name), name);
+    xcb_intern_atom_reply_t *reply = xcb_intern_atom_reply (connection, cookie, NULL);
 
-    cookie = xcb_intern_atom (connection, 0, strlen(name), name);
-    reply = xcb_intern_atom_reply (connection, cookie, NULL);
-    if (reply)
+    xcb_atom_t atom;
+    if (reply) {
         atom = reply->atom;
-    else
+        free(reply);
+    } else
         atom = XCB_NONE;
 
-    free(reply);
     return atom;
 }
 
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -179,7 +179,7 @@ xcb_paint_image (struct wpe_fdo_egl_exported_image *image)
     glActiveTexture (GL_TEXTURE0);
     glBindTexture (GL_TEXTURE_2D, s_window->gl.texture);
     s_display->egl.image_target_texture (GL_TEXTURE_2D,
-                                          wpe_fdo_egl_exported_image_get_egl_image (image));
+                                         wpe_fdo_egl_exported_image_get_egl_image (image));
     glUniform1i (s_window->gl.uniform_texture, 0);
 
     glVertexAttribPointer (s_window->gl.attr_pos, 2, GL_FLOAT, GL_FALSE, 0, position_coords);
@@ -639,9 +639,9 @@ init_egl (void)
     }
 
     s_display->egl.context = eglCreateContext (s_display->egl.display,
-                                         s_display->egl.config,
-                                         EGL_NO_CONTEXT,
-                                         context_attribs);
+                                               s_display->egl.config,
+                                               EGL_NO_CONTEXT,
+                                               context_attribs);
     if (s_display->egl.context == EGL_NO_CONTEXT)
         return FALSE;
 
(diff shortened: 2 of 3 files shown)


platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    };
+    wpe_view_backend_dispatch_keyboard_event (wpe_view_data.backend, &input_event);
+}
+
+static void
+xcb_handle_axis (xcb_button_press_event_t *event, int16_t axis_delta[2])
+{
+#if HAVE_2D_AXIS_EVENT
+    struct wpe_input_axis_2d_event input_event = { 0, };
+    input_event.base.type = wpe_input_axis_event_type_mask_2d | wpe_input_axis_event_type_motion_smooth;
```
aperezdc, line 249: This could use a designated initializer, which would guarantee that members not specified will be initialized with zeroes, too.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    }
+
+    wpe_view_backend_dispatch_axis_event (wpe_view_data.backend, &input_event);
+#endif
+}
+
+static void
+xcb_handle_button_press (xcb_button_press_event_t *event)
+{
+    int16_t axis_delta[4][2] = {
```
aperezdc, line 283: This is a read-only lookup table, it would be good to mark it `static const` to help the compiler optimize it away.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    xcb_paint_image (image);
+}
+
+static void
+xcb_process_events (void)
+{
+    xcb_generic_event_t *event = NULL;
+
+    while ((event = xcb_poll_for_event (xcb_data.connection))) {
+        switch (event->response_type & 0x7f) {
```
aperezdc, line 372: Instead of a magic number it is better use the macros defined by the `xcb/xcb_event.h` header:

```c
/**
 * @brief Bit mask to find event type regardless of event source.
 *
 * Each event in the X11 protocol contains an 8-bit type code.
 * The most-significant bit in this code is set if the event was
 * generated from a SendEvent request. This mask can be used to
 * determine the type of event regardless of how the event was
 * generated. See the X11R6 protocol specification for details.
 */
#define XCB_EVENT_RESPONSE_TYPE_MASK (0x7f)
#define XCB_EVENT_RESPONSE_TYPE(e)   (e->response_type &  XCB_EVENT_RESPONSE_TYPE_MASK)
#define XCB_EVENT_SENT(e)            (e->response_type & ~XCB_EVENT_RESPONSE_TYPE_MASK)
```

In this case we would want:

```suggestion
        switch (XCB_EVENT_RESPONSE_TYPE(event)) {
```

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+static gboolean
+xcb_source_dispatch (GSource *base, GSourceFunc callback, gpointer user_data)
+{
+    struct xcb_source *source = (struct xcb_source *) base;
+    if (source->pfd.revents & (G_IO_ERR | G_IO_HUP))
+        return FALSE;
+
+    xcb_process_events ();
+    source->pfd.revents = 0;
+    return TRUE;
```
aperezdc, line 457:
```suggestion
    return G_SOURCE_CONTINUE;
```

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    struct xcb_source *source = (struct xcb_source *) base;
+    return !!source->pfd.revents;
+}
+
+static gboolean
+xcb_source_dispatch (GSource *base, GSourceFunc callback, gpointer user_data)
+{
+    struct xcb_source *source = (struct xcb_source *) base;
+    if (source->pfd.revents & (G_IO_ERR | G_IO_HUP))
+        return FALSE;
```
aperezdc, line 453: This would be more readable using symbolic constants `G_SOURCE_{CONTINUE,REMOVE}`:

```suggestion
        return G_SOURCE_REMOVE;
```

Also, I am missing here code for handling an errored XCB connection, it would be good to do a check with `xcb_connection_has_error()` and return `G_SOURCE_REMOVE` in that case as well.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    source->pfd.revents = 0;
+    return TRUE;
+}
+
+static xcb_atom_t
+get_atom (struct xcb_connection_t *connection, const char *name)
+{
+    xcb_intern_atom_cookie_t cookie;
+    xcb_intern_atom_reply_t *reply;
+    xcb_atom_t atom;
```
aperezdc, line 465: We are building in C99 mode, these variables could be assigned as soon as they are defined, and there is no need to define them all at the beginning of the function. Also we only support Unixish systems, so using `g_autofree`, `g_autoptr()`, and friends is not only allowed, but encouraged.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+    xcb_data.connection = XGetXCBConnection (xcb_data.display);
+    if (xcb_connection_has_error (xcb_data.connection))
+        return FALSE;
+
+    xcb_data.window = xcb_generate_id (xcb_data.connection);
+
+    const struct xcb_setup_t *setup = xcb_get_setup (xcb_data.connection);
+    xcb_data.screen = xcb_setup_roots_iterator (setup).data;
+
+    uint32_t window_values[] = {
```
aperezdc, line 494: This could be `static const`.

aperezdc (changes requested):
Whew! This took a while to read through properly, sorry about the delay! The code works (I did test it), and I would very much like to merge this, but there are many small improvements which could be done all around. In particular I think we should start avoiding the static `foo_data` globals and start using proper abstractions like the suggested `CogX11Display` and `CogX11Window` types—which will make it more amenable to extend the code later on.

Thanks a lot for working on this! :muscle:

CMakeLists.txt
```diff
@@ -346,4 +347,8 @@ if (COG_PLATFORM_DRM AND NOT COG_USE_WEBKITGTK)
     )
 endif () # COG_PLATFORM_DRM AND NOT COG_USE_WEBKITGTK
 
+if (COG_PLATFORM_X11 AND NOT COG_USE_WEBKITGTK)
+    add_subdirectory(platform/x11)
```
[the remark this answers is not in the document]
zdobersek: Later, when some of the pending PRs are handled.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+#include <X11/Xlib.h>
+#include <X11/Xlib-xcb.h>
+
+
+#ifndef EGL_EXT_platform_base
+#define EGL_EXT_platform_base 1
+typedef EGLDisplay (EGLAPIENTRYP PFNEGLGETPLATFORMDISPLAYEXTPROC) (EGLenum platform, void *native_display, const EGLint *attrib_list);
+typedef EGLSurface (EGLAPIENTRYP PFNEGLCREATEPLATFORMWINDOWSURFACEEXTPROC) (EGLDisplay dpy, EGLConfig config, void *native_window, const EGLint *attrib_list);
+typedef EGLSurface (EGLAPIENTRYP PFNEGLCREATEPLATFORMPIXMAPSURFACEEXTPROC) (EGLDisplay dpy, EGLConfig config, void *native_pixmap, const EGLint *attrib_list);
+#endif
```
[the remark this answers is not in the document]
zdobersek: It would make sense to use it, I would do it as a separate change.

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,896 @@
...
+
+#define DEFAULT_WIDTH  1024
+#define DEFAULT_HEIGHT  768
+
+#if defined(WPE_CHECK_VERSION)
+# define HAVE_2D_AXIS_EVENT WPE_CHECK_VERSION(1, 5, 0) && WEBKIT_CHECK_VERSION(2, 27, 4)
+#else
+# define HAVE_2D_AXIS_EVENT 0
+#endif /* WPE_CHECK_VERSION */
```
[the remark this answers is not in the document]
zdobersek: Fine with that, but it should be done separately since this 2D-axis-event check alone is done in multiple modules and in the core application/library.

platform/x11/cog-platform-x11.c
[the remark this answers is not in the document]
aperezdc: Of course, I have filed #240 to remember about it later.

platform/x11/cog-platform-x11.c
zdobersek: In Debian, this header is packaged through the libxcb-util0 package, so that would be a separate dependency. Let's avoid that for now.

platform/x11/cog-platform-x11.c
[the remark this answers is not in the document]
aperezdc: Sure, let's do it as a follow-up later on, I've filed issue #241 to track this one :+1:

CMakeLists.txt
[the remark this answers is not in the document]
aperezdc: Yes, let's handle the change later, I've filed #242 to track this task :ok_hand:

platform/x11/cog-platform-x11.c
```diff
@@ -0,0 +1,914 @@
...
+
+
+struct CogX11Display {
+    Display *display;
```
aperezdc, line 53: Much neater :star_struck:

aperezdc (approved):
Thanks for all the improvements, let's merge this now and we can continue improving things later on :ship:

PR #243, opened 2020-10-27
cmake: Check and link FDO platform dependencies independently

tobias-grasse:
If any of the modules specified in the `COG_PLATFORM_FDO_DEPS` target cannot be found, cmake will error out somewhat unspecific. This ensures each dependency is searched and linked independently, which should be slightly more robust and/or produce more helpful error messages.


tobias-grasse:
Superseded by #244 😊

aperezdc:
Thanks anyway @tobias-grasse, you were going in the good direction here 👌️

PR #247, opened 2020-11-07
drm: implement keyboard event repeating when the key remains pressed

kytart:
inspired by wpebackend-rdk https://github.com/WebPlatformForEmbedded/WPEBackend-rdk/blob/master/src/input/Libinput/KeyboardEventRepeating.cpp

### DIFF
--- a/platform/cog-platform-drm.c
+++ b/platform/cog-platform-drm.c
@@ -33,6 +33,8 @@ typedef EGLDisplay (EGLAPIENTRYP PFNEGLGETPLATFORMDISPLAYEXTPROC) (EGLenum platf
 # define HAVE_DEVICE_SCALING 0
 #endif
 
+#define KEY_STARTUP_DELAY 500000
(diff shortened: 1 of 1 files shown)


aperezdc (changes requested):
Hi @kytart! Thanks for the patch, the logic looks good.

I only have a few comments about code style. I know that it can feel like a drag to take care of those, but it's better in the long run for maintanability. Could you take a look at the comments?

kytart:
@aperezdc I pushed the changes that you suggested. Have a look now pls :)

aperezdc (approved):
This patch looks good as well, and again the WebKitGTK CI build failure is not relevant anymore after #257 — let's get this merged!

Thanks a lot to @kytart for the patch and his patience waiting for me to come back to re-review this :innocent:

kytart:
@aperezdc Amazing! Thanks for coming back to it :)

PR #249, opened 2020-11-10
platform: add a common EGL proc address loader with dlsym fallback

zdobersek:
Provide a common EGL proc address loader function that incorporates
a dlsym-based fallback in case eglGetProcAddress() refuses to find
an otherwise-existing entrypoint.

This should avoid some drivers that fail to handle proc address
requests for specific entrypoints, but have those entrypoints
exported as regular symbols loadable through dlsym().

### DIFF
--- a/platform/common/egl-proc-address.h
+++ b/platform/common/egl-proc-address.h
@@ -1,3 +1,10 @@
+/*
+ * egl-proc-address.h
+ * Copyright (C) 2020 Igalia S.L.
+ *
+ * Distributed under terms of the MIT license.
+ */
+
 #pragma once
 
 #define __USE_GNU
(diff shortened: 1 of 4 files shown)


platform/common/egl-proc-address.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
```
aperezdc, line 1: We should add a license header to this new file :wink:

aperezdc (changes requested):
Just one small thing missing, other than that code looks fine so if you could update the PR taking into account the comment below, this can be approved :ok_hand:

aperezdc (approved):
@zdobersek: thanks for the quick update!

PR #251, opened 2020-11-26
Remove default about:blank load action in the  View constructor

psaavedra:
Signed-off-by: Pablo Saavedra <psaavedra@igalia.com>

### DIFF
--- a/core/cog-view.c
+++ b/core/cog-view.c
@@ -100,16 +100,6 @@ cog_view_constructor (GType                  type,
 }
 
 static void
-cog_view_constructed (GObject *object)
-{
-    G_OBJECT_CLASS (cog_view_parent_class)->constructed (object);
-
-    webkit_web_view_load_html (WEBKIT_WEB_VIEW (object),
-                               "<!DOCTYPE html><html><body></body></html>",
-                               "about:blank");
-}
-
-static void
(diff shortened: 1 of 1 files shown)


core/cog-view.c
```diff
@@ -103,10 +103,6 @@ static void
 cog_view_constructed (GObject *object)
 {
     G_OBJECT_CLASS (cog_view_parent_class)->constructed (object);
-
-    webkit_web_view_load_html (WEBKIT_WEB_VIEW (object),
-                               "<!DOCTYPE html><html><body></body></html>",
-                               "about:blank");
```
aperezdc, line 109: Now that the `.constructed` callback does nothing, we can remove the whole `cog_view_constructed()` function.

aperezdc (changes requested):
The idea sounds fine to me, but let's clean the code a bit more :smile:

core/cog-view.c
psaavedra: Updated. Please take in consideration these changes to be merged in Cog 0.10. Thanks.

aperezdc: We don't need to do anything for 0.10 because `master` does not have the dummy load of `about:blank` :wink:

PR #264, opened 2021-02-24
core: Add new custom URI scheme handler that routes based on path prefixes 

aperezdc:
Add a new `CogPrefixRoutesHandler` class (which implements `CogRequestHandler`) and can dispatch a number of URI path prefixes to different handlers, with an optional fallback handler for unmatched URI paths.

Additionally, complement it with the addition of the `CogDirectoryFilesHandler:strip-components` property, which allows for removing URI path prefixes, which comes in handy after they have been matched by `CogPrefixRoutesHandler`.

### DIFF
--- a/core/cog-directory-files-handler.c
+++ b/core/cog-directory-files-handler.c
@@ -13,12 +13,14 @@ struct _CogDirectoryFilesHandler {
     GObject  parent;
     GFile   *base_path;
     gboolean use_host;
+    unsigned strip_components;
 };
 
 enum {
     PROP_0,
     PROP_BASE_PATH,
     PROP_USE_HOST,
+    PROP_STRIP_COMPONENTS,
     N_PROPERTIES,
 };
 
@@ -186,6 +188,27 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
     while (path[0] == '/')
         ++path;
 
+    /*
+     * Discard non-empty leading path components, ignoring empty path
+     * components (empty component == two consecutive slashes). For
+     * example with "strip-components" being 2, paths would get munged
+     * as follows:
+     *
+     *   - /foo/bar/baz  -> /baz
+     *   - /foo//bar/baz -> /baz
+     *   - /foo/bar/     -> /
+     *   - /foo/bar      -> (empty)
+     *   - /foo          -> (empty)
+     */
+    for (unsigned i = 0; i < handler->strip_components && path[0] != '\0'; i++) {
+        /* Skip up to the next slash. */
+        while (path[0] != '/' && path[0] != '\0')
+            ++path;
+        /* Skip consecutive slashes. */
+        while (path[0] == '/')
+            ++path;
+    }
+
     g_autoptr(GFile) file = g_file_get_child (base_path, path);
 
     /*
@@ -246,6 +269,9 @@ cog_directory_files_handler_get_property (GObject    *object,
         case PROP_USE_HOST:
             g_value_set_boolean (value, cog_directory_files_handler_get_use_host (handler));
             break;
+        case PROP_STRIP_COMPONENTS:
+            g_value_set_uint (value, cog_directory_files_handler_get_strip_components (handler));
+            break;
         default:
             G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
     }
@@ -267,6 +293,10 @@ cog_directory_files_handler_set_property (GObject      *object,
         case PROP_USE_HOST:
             cog_directory_files_handler_set_use_host (handler,
                                                       g_value_get_boolean (value));
+            break;
+        case PROP_STRIP_COMPONENTS:
+            cog_directory_files_handler_set_strip_components (handler,
+                                                              g_value_get_uint (value));
             break;
         default:
             G_OBJECT_WARN_INVALID_PROPERTY_ID (object, prop_id, pspec);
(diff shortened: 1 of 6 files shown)


core/cog-prefix-routes-handler.c
```diff
@@ -0,0 +1,341 @@
...
+
+static void
+cog_prefix_routes_handler_run (CogRequestHandler      *request_handler,
+                               WebKitURISchemeRequest *request)
+{
+    CogPrefixRoutesHandler *self = COG_PREFIX_ROUTES_HANDLER (request_handler);
+
+    const char *uri_path = webkit_uri_scheme_request_get_path (request);
+    if (!uri_path)
+        return cog_prefix_routes_handler_run_fallback (self, request);
```
psaavedra, line 98: if the WebKit can resolve the path, it is not much more simple just end the logic at this point with a `webkit_uri_scheme_request_finish_error()`?

core/cog-prefix-routes-handler.c
```diff
@@ -0,0 +1,341 @@
...
+ *
+ * The following configures a handler to route requests with their path
+ * starting with `/static/images` to a certain directory, and all the
+ * rest of paths to another—note how the
+ * [property@Cog.DirectoryFilesHandler:strip-components] property is
+ * used to remove the matched path components for that route:
+ *
+ * ```c
+ * GFile *content_dir = g_file_new_for_path ("data/content");
+ * CogRequestHandler *fallback_handler = cog_directory_files_handler_new (fallback_dir);
```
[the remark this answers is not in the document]
aperezdc: Ouch, yes, this is a typo.

core/cog-prefix-routes-handler.c
aperezdc: The path can be empty because an URI can have an empty path and even an empty authority section, both `foo:` and `foo://authority are valid, and the path will be `NULL` in that case. The `CogPrefixRoutesHandler` cannot handle an empty path because it always needs to match a path prefix, so if there is no path the idea is to pass the request to the fallback handler, which may support URIs with empty paths.

core/cog-prefix-routes-handler.c
```diff
@@ -0,0 +1,341 @@
...
+        /* Position 0 is always the leading slash. */
+        if (last_slash_pos == 0)
+            break;
+
+        g_string_erase (prefix, last_slash_pos, -1);
+
+        CogRequestHandler *handler =
+            g_hash_table_lookup (self->routes, prefix->str);
+        if (handler)
+            return cog_request_handler_run (handler, request);
```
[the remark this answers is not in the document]
aperezdc: Sure, I will add it. Also a debug message when none of the configured prefixes has been matched is probably a good complement :smile_cat:

core/cog-prefix-routes-handler.c
```diff
@@ -0,0 +1,341 @@
...
+
+/**
+ * cog_prefix_routes_handler_mount_path:
+ * @self: a #CogPrefixRoutesHandler
+ * @path_prefix: URI path prefix to match
+ * @base_path: Path to a local directory
+ *
+ * Adds a route to the handler pointing to a directory.
+ *
+ * This is a convenience method which convenienceonfigures a route
```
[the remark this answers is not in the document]
aperezdc: Let's rewrite it as “This is a utility method which configures […]”, for example. If you ave a suggestion for better wording, I'm all ears.

core/cog-prefix-routes-handler.c
```diff
@@ -0,0 +1,341 @@
...
+
+    unsigned num_components = 0;
+    for (size_t i = 0; path_prefix[i] != '\0'; i++)
+        if (path_prefix[i] == '/')
+            num_components++;
+
+    g_autoptr(CogRequestHandler) handler =
+        g_object_new (COG_TYPE_DIRECTORY_FILES_HANDLER,
+                      "base-path", path_file,
+                      "strip-components", num_components);
```
aperezdc, line 336: I have realized that this call to `g_object_new()` is missing the terminating `NULL` at the end of the variable list of arguments, I had a release build crash at me just a moment ago because this :facepalm: 

The updated PR will fix this as well.

aperezdc:
@psaavedra Thanks for the review, I have pushed an update with the comments addressed :factory_worker:

core/cog-prefix-routes-handler.c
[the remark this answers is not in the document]
psaavedra: I'm not a native speaker so I'm not the right person to suggest anything on that regard. Actually it mor was much more like a typo: `s/convenienceonfigures/configures/`.  Your proposal LGTM.

core/cog-prefix-routes-handler.c
psaavedra: Ok. I see the point. Makes sense.

core/cog-prefix-routes-handler.c
psaavedra: Good catch.

core/cog-directory-files-handler.c
```diff
@@ -186,6 +188,14 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
     while (path[0] == '/')
         ++path;
 
+    /* Discard non-empty leading path components. */
```
psaavedra, line 191: This logic is not trivial. Not for my eyes at least. It could be nice if you could add here what is the expected behavior with some specific examples.

psaavedra (approved):

core/cog-directory-files-handler.c
aperezdc: Comment added to explain things a bit, I hope it makes the code easier to follow.

core/cog-directory-files-handler.c
```diff
@@ -186,6 +188,27 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
     while (path[0] == '/')
         ++path;
 
+    /*
+     * Discard non-empty leading path components, ignoring empty path
```
psaavedra, line 192: Very appreciate this comment. It makes much more simple the understanding of the code. Thanks.

PR #265, opened 2021-02-25
core: Ensure that files URI handler does not escape the base path

aperezdc:
Return early from `cog_directory_files_handler_run()` if the calculated path to the location of the resource on disk happens to not be a descendant of the base path.

### DIFF
--- a/core/cog-directory-files-handler.c
+++ b/core/cog-directory-files-handler.c
@@ -187,6 +187,24 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
         ++path;
 
     g_autoptr(GFile) file = g_file_get_child (base_path, path);
+
+    /*
+     * Check whether the resolved path is contained inside base_path,
+     * to prevent URIs with ".." components from accessing resources
+     * outside of base_path. The g_file_get_relative_path() method
+     * returns NULL when the file path is NOT descendant of base_path.
+     */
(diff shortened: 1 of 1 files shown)


psaavedra (approved):

aperezdc:
Cherry-picked into the `cog-0.8` branch as commit d54eb5277460a0601d6069f851ca726f8067f029

core/cog-directory-files-handler.c
```diff
@@ -187,6 +187,24 @@ cog_directory_files_handler_run (CogRequestHandler      *request_handler,
...
     g_autoptr(GFile) file = g_file_get_child (base_path, path);
+
+    /*
+     * Check whether the resolved path is contained inside base_path,
+     * to prevent URIs with ".." components from accessing resources
+     * outside of base_path. The g_file_get_relative_path() method
+     * returns NULL when the file path is NOT descendant of base_path.
+     */
+    g_autofree char *relative_path = g_file_get_relative_path (base_path, file);
+    if (!relative_path) {
```
aperezdc, line 198: This introduces a regression: the check does not cover the case when `relative_path` and `base_path` end up being the same path, so if the path is empty an error is returned instead of checking whether an `index.html` file is present at the base path. I'll post a PR with a fix momentarily.

aperezdc: Fix in #266

PR #266, opened 2021-03-01
core: Allow resolving base path in URI files handler

aperezdc:
When checking whether a resolved path is a descendant of the base path in `cog_directory_files_handler_run()`, allow the resolved path to be the same as the base path. This is needed to allow URIs which point to the root of the base path (either without a path, or a single slash as path) and attempt to find an `index.html` file there.

This fixes a regression introduced by #265


carlosgcampos (approved):

aperezdc:
Cherry-picked into the `cog-0.8` branch as commit 5a8b0b414ad1ee17564273caba880926d5cd2c8c

Thanks for the approval @carlosgcampos :smile:

PR #268, opened 2021-03-03
drm: support wpe_fdo_shm_exported_buffer exports

zdobersek:
Handle the wpe_fdo_shm_exported_buffer exports that might arrive on
the wpe_view_backend_exportable_fdo_client interface.

When that occurs, we craft a gbm_bo with the necessary flags that
would allow us to map the bo's underlying memory and write into it.
We still create a framebuffer object off of it, pretty much like we
do it for other types of bos.

The new buffer objects use the existing caching facilities. Each
buffer object, cached or newly created, is updated by mapping the bo
memory into the address space and copying over the data contained
in the shm area.

For now, only ARGB8888 and XRGB8888 SHM formats are handled. In both
cases the corresponding gbm bo will use the XRGB8888 format.

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -787,7 +787,7 @@ drm_copy_shm_buffer_into_bo (struct wl_shm_buffer *shm_buffer, struct gbm_bo *bo
             }
         }
     } else
-        memcpy(dst, src, stride * height);
+        memcpy (dst, src, stride * height);
 
     wl_shm_buffer_end_access (shm_buffer);
     gbm_bo_unmap (bo, map_data);
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -46,6 +46,11 @@ struct buffer_object {
     uint32_t fb_id;
     struct gbm_bo *bo;
     struct wl_resource *buffer_resource;
+
+    struct  {
+        struct wl_resource* resource;
+        struct wpe_fdo_shm_exported_buffer *shm_buffer;
+    } export;
 };
 
 static struct {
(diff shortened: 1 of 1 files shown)


platform/cog-platform-drm.c
```diff
@@ -46,6 +46,11 @@ struct buffer_object {
     uint32_t fb_id;
     struct gbm_bo *bo;
     struct wl_resource *buffer_resource;
```
aperezdc, line 48: The intention of the PR seems to be replacing `buffer_resource` with `export.resource`, right? In that case, the `buffer_resource` member should be removed here.

platform/cog-platform-drm.c
```diff
@@ -682,6 +708,91 @@ drm_create_buffer_for_bo (struct gbm_bo *bo, struct wl_resource *buffer_resource
...
+        for (uint32_t y = 0; y < height; ++y) {
+            for (uint32_t x = 0; x < width; ++x) {
+                dst[bo_stride * y + 4 * x + 0] = src[stride * y + 4 * x + 0];
+                dst[bo_stride * y + 4 * x + 1] = src[stride * y + 4 * x + 1];
+                dst[bo_stride * y + 4 * x + 2] = src[stride * y + 4 * x + 2];
+                dst[bo_stride * y + 4 * x + 3] = src[stride * y + 4 * x + 3];
+            }
+        }
+    } else
+        memcpy(dst, src, stride * height);
```
aperezdc, line 790: Nit: missing space between function name and parenthesis:

```suggestion
        memcpy (dst, src, stride * height);
```

aperezdc (commented):
This would need a rebase, otherwise it looks mergeable to me. Thanks a ton for the patch, @zdobersek! :muscle:

platform/cog-platform-drm.c
zdobersek: This pointer is used for the purposes of caching existing imported buffers. It's used for comparison in the `drm_buffer_for_resource()`.

The `export` struct below is used for holding references to whatever object has been presented and needs subsequent release. Holding separate references here makes it cleaner to test and release these resources, since e.g. for the `wpe_fdo_shm_exported_buffer` export both this object as well as the `wl_resource` would be set, but only the first one would have to be released.

aperezdc (approved):
This is all good to go now, thanks @zdobersek :muscle:

PR #271, opened 2021-03-07
CI: Switch from Travis-CI to GitHub Actions

### LINKED ISSUE #270: CI: Switch away from Travis
It's not happening too often, but sometimes we do hit the rate limit allotted to the free tier accounts (which was [changed a little while ago](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing)). We already have `libwpe` and `WPEBackend-fdo` running CI on GitHub Actions so that is probably the path of least effort for switching.

aperezdc:
Fixes #270


psaavedra (approved):

.github/toolchain.sh
```diff
@@ -0,0 +1,16 @@
+#! /bin/sh
+set -e
+
+INSTALL_DIR=${1:-$HOME/toolchain}
+BASEURL=https://wk-contrib.igalia.com/yocto/meta-perf-browser/browsers/nightly/sdk
+FILE=wandboard-mesa/browsers-glibc-x86_64-core-image-weston-browsers-armv7at2hf-neon-wandboard-mesa-toolchain-1.0.sh
```
psaavedra, line 6: Just a sidenote, the nightly backend contains versions from main/HEAD for the Webkit components (wpe, wpebackend-fdo and libwpe). There is also an toolchain with the stable versions: s/nightly/candidate/.

I'm sharing this info just in case you could be also interested in test the component against those stable versions as well.

aperezdc: Oh, that's interesting, thanks for mentioning it! I think we should keep this PR small and only replace Travis. I have filed #272 to maybe work on improvements later.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,41 @@
...
+          sudo apt update
+          sudo apt install -y cmake ninja-build
+      - name: Toolchain Cache
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: ~/toolchain
+          key: tc-${{ runner.os }}-${{ hashFiles('.github/toolchain.sh') }}
+      - name: Install Toolchain
+        if: steps.cache.outputs.cache-hit != 'true'
```
aperezdc, line 25: This is not great, because if a cached toolchain is found, a newer one is never downloaded, even when a new build of the toolchain is available. I need to change this step to be unconditional, and use `curl --time-cond` like we did with Travis.

psaavedra: splited off in https://github.com/Igalia/cog/issues/272

aperezdc: I have updated the PR to fix this issue; what was split in #272 is using the toolchains with the stable packages :wink:

psaavedra (approved):

PR #279, opened 2021-03-09
CI: Add a CI job that runs clang-format on PR changes

aperezdc:
Introduce tooling to check the coding style,and a CI worflow which runs it only on the changes introduced by a PR. The goal of checking only the changed code is to avoid having to run a full reformatting of the tree: instead, changes that modified existing code will slowly make the tree drift towards the desired status without introducing noise in the repository history.

There are is some trickery involved in getting this to work reliably:

- The `clang-format-diff` helper script is installed at different locations depending on the distribution. The `check-style` script checks a few locations that cover popular GNU/Linux distributions (Ubuntu, Debian, Arch Linux, Fedora); it can be improved as we see fit in the future.
- The `actions/checkout` step defaults to shallow clones of the single commit which triggered a PR, which is enough to build, but not to find the differences between it and the base commit. To avoid forcing a full fetch of the history, the job uses a second instance of the action to fetch only the base tree in a separate directory.
- Even with both trees checked out, Git will refuse to calculate a diff because the commits are not linked to one another. But both trees are now checked out, so instead it is possible to use a plain   `diff` and feed its output to `check-style --diff`.

As a bonus, the `check-style` script can be run by developers locally from the command line in order to verify their changes prior to filing merge requests, e.g:

```sh
git co -b my-meature
$EDITOR cog.c
git ci -m'Awesome improvements'
data/check-style master
```

The output from the script is an unified diff, which can be piped back into `patch` or `git apply` to get the suggested changes applied to the tree:

```sh
data/check-style master | git apply -p0 -
git add -p && git commit --amend
```

### DIFF
--- a/data/check-style
+++ b/data/check-style
@@ -1,0 +1,127 @@
+#! /bin/bash
+set -eu -o pipefail
+
+FILE=
+BASE=
+HEAD=
+
+declare -a args=()
+diff_input=false
+show_help=false
+
+for opt in "$@" ; do
+    case "${opt}" in
+        -h | --help)
+            show_help=true
+            ;;
+        -d | --diff)
+            diff_input=true
+            ;;
+        -o*)
+            FILE=${opt#-o}
+            ;;
+        *)
+            args+=("${opt}")
+            ;;
+    esac
+done
+declare -r diff_input show_help
+
+show_help ()
+{
+    echo "Usage: $0 [-o<file>] [-d|--diff] [<committish> [<committish>]]"
+}
+
+if ${show_help} ; then
+    show_help
+    exit
+fi
+
+if [[ ${#args[@]} -eq 1 ]] ; then
+    BASE=${args[0]}
+elif [[ ${#args[@]} -eq 2 ]] ; then
+    BASE=${args[0]}
+    HEAD=${args[1]}
+elif [[ ${#args[@]} -eq 0 ]] && ${diff_input} ; then
+    : no-op
+else
+    show_help 1>&2
+    exit 1
+fi
+declare -r BASE HEAD
+
+declare -ra clang_format_diff_candidates=(
+    # Search in $PATH (Ubuntu, Debian, maybe others)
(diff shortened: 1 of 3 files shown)


aperezdc:
Note that the included `.clang-format` file is mostly a dump (generated with `clang-format --dump-config`) has only a few small modifications, we may want to tweak it a bit before merging this PR; but we can also modify it lazily whenever we think it is suggesting changes that we do not want to follow.

What do people think about this change and trying to use the CI to enforce the coding style?

data/check-style
```diff
@@ -0,0 +1,99 @@
+#! /bin/bash
+set -e
```
clopez, line 2: Better use:
```
set -eu -o pipefail
```

clopez (approved):
LGTM

kkonopko:
Note that you may need to check for a minimum `clang-format` version.  For example if someone re-formats their diff with an older version (installed by their distro) which does not support certain features in the project's format specification, it may get confusing.

data/check-style
aperezdc: I often forget these extras, I'll add them.

aperezdc:
> Note that you may need to check for a minimum `clang-format` version. For example if someone re-formats their diff with an older version (installed by their distro) which does not support certain features in the project's format specification, it may get confusing.

Yes, this is a very valid point. Also it might be that an older `clang-format` cannot load our `.clang-format` configuration file so it seems like a good idea to add a check for the version in the `check-style` script—I'll add it, thanks for the suggestion!

aperezdc:
I am thinking of abandoning the GLib-styled “space between function/macro name and opening parentheses” and the oddball function type-parameter-name spacing going forward, so I will be tweaking a bit more the `.clang-format` file to reflect that.

aperezdc:
Let's merge this, the CI job works as intended. For adjustments to the formatting style I will send another PR in the next days.

PR #280, opened 2021-03-15
drm: enable the platform module by default

zdobersek:
The thing should be mature enough by now to be enabled by default.
Exception to this are different modifier issues that have to be
solved with different approaches around dmabuf management.


aperezdc (approved):
Agreed, it has been part of the tree for a good while, and from the discussions in the chat room and mailing list it is clear some people are using it regularly.

PR #281, opened 2021-03-15
drm: add necessary guards for SHM export handling

zdobersek:
Check for a recent-enough libwpebackend-fdo build dependency
and use the SHM export API and handling logic accordingly.


aperezdc (approved):
Good catch, thanks @zdobersek!

PR #282, opened 2021-03-15
fdo: Use WAYLAND_VERSION_* instead of CMake check

aperezdc:
Use the `WAYLAND_VERSION_{MAJOR,MINOR}` macros to check whether the FDO platform plug-in is being built against Wayland 1.10 or newer. This makes it unnecessary to check the version with CMake.

This also fixes a recent regression which caused `WAYLAND_1_10_OR_GREATER` to not be defined by CMake anymore.

### DIFF
--- a/platform/fdo/cog-platform-fdo.c
+++ b/platform/fdo/cog-platform-fdo.c
@@ -81,6 +81,13 @@
 # define HAVE_SHM_EXPORTED_BUFFER 0
 #endif
 
+#if defined(WAYLAND_VERSION_MAJOR) && defined(WAYLAND_VERSION_MINOR)
+# define WAYLAND_1_10_OR_GREATER ((WAYLAND_VERSION_MAJOR >= 2) || \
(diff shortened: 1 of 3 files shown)


platform/fdo/cog-platform-fdo.c
```diff
@@ -81,6 +81,13 @@
 # define HAVE_SHM_EXPORTED_BUFFER 0
 #endif
 
+#if defined(WAYLAND_VERSION_MAJOR) && defined(WAYLAND_VERSION_MINOR)
+# define WAYLAND_1_10_OR_GREATER ((WAYLAND_VERSION_MAJOR > 10) || \
```
aperezdc, line 85: Ah, this should be `> 1`, let me fix it and reupload.

aperezdc: Or even `>= 2`, to make it slightly clearer :thinking:

zdobersek (approved):

PR #283, opened 2021-03-15
Depend on WPE WebKit 2.28, libwpe+WPEBackend-fdo 1.5

### LINKED ISSUE #241: Bump required libwpe version to 1.5.0 and remove 2D axis event conditionals
There are a few places in the code that have a check similar to this:

```c
#if defined(WPE_CHECK_VERSION)
# define HAVE_2D_AXIS_EVENT WPE_CHECK_VERSION(1, 5, 0) && WEBKIT_CHECK_VERSION(2, 27, 4)
#else
# define HAVE_2D_AXIS_EVENT 0
#endif /* WPE_CHECK_VERSION */
```

Versions 1.6.x of libwpe and 2.28.x have been around for a while, so we could change the minimum requirement to those and remove the conditionals to always use 2D axis events.

aperezdc:
Update the minimum required versions to the following, which are all more than one year old:

- WPE WebKit 2.27.3, the oldest development version of the 2.28 series   which already has most of the features used by Cog.
- libwpe and WPEBackend-fdo 1.5.0, with the same reasoning.

This allows removing a bunch of conditional code.

Fixes #241

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,29 +115,10 @@ set(COGCORE_SOURCES
 
 pkg_check_modules(GIO IMPORTED_TARGET REQUIRED gio-2.0>=2.44)
 pkg_check_modules(SOUP IMPORTED_TARGET REQUIRED libsoup-2.4)
-
-# There is no need to explicitly check wpe-1.0 here because it's a
-# dependency already specified in the wpe-webkit.pc file.
-pkg_check_modules(WEB_ENGINE IMPORTED_TARGET REQUIRED wpe-webkit-1.0>=2.23.91)
-if ("${WEB_ENGINE_VERSION}" VERSION_GREATER "2.23")
-    add_definitions(-DCOG_BG_COLOR_API_SUPPORTED=1)
-else ()
-    add_definitions(-DCOG_BG_COLOR_API_SUPPORTED=0)
-endif ()
-if ("${WEB_ENGINE_VERSION}" VERSION_GREATER "2.27.3")
-    set(COG_IM_API_SUPPORTED ON)
-else ()
-    set(COG_IM_API_SUPPORTED OFF)
-endif ()
-if (COG_PLATFORM_FDO AND "${WEB_ENGINE_VERSION}" VERSION_GREATER "2.27.3")
-    set(COG_FDO_WIDGETS_SUPPORTED ON)
-else ()
-    set(COG_FDO_WIDGETS_SUPPORTED OFF)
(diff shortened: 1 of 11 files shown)


aperezdc:
I think that this could tentatively be part of the 0.10 release. If anybody else thinks it's late for this in the development cycle I am open to merge this after branching for 0.10—which should be soon anyway :smile_cat:

aperezdc:
This turned out to be a bit big of a change, so in a chat with @zdobersek we have agreed to merge it _after_ the branch for the stable 0.10 releases is made.

aperezdc:
Ping reviewers 🏓 — the `cog-0.10` release branch has been created, so we could not merge this into main development branch.

CMakeLists.txt
```diff
@@ -115,30 +115,11 @@ set(COGCORE_SOURCES
...
-    set(COG_IM_API_SUPPORTED ON)
-else ()
-    set(COG_IM_API_SUPPORTED OFF)
-endif ()
-if (COG_PLATFORM_FDO AND "${WEB_ENGINE_VERSION}" VERSION_GREATER "2.27.3")
-    set(COG_FDO_WIDGETS_SUPPORTED ON)
-else ()
-    set(COG_FDO_WIDGETS_SUPPORTED OFF)
-endif ()
+pkg_check_modules(WebKit IMPORTED_TARGET REQUIRED wpe-webkit-1.0>=2.27.3 wpe-1.0>=1.5.0)
```
philn, line 118: s/2.27.3/2.28.0 ?

philn (changes requested):
LGTM, just a minor comment

CMakeLists.txt
philn: I guess I get the initial reasoning though, but I wouldn't really expect anyone to use pre-releases for their products.

aperezdc: Development release 2.27.3 happens to be the very first one that supports all the features that this PR removes the conditional compilation guards for. I don't see a problem in leaving it like that, but also there is no downside using 2.28.0, so I'll update the PR with that 😄

aperezdc: Also, I'll change it things to use `wpe-1.0>=1.6.0` and `wpebackend-fdo-1.6.0` while at it.

aperezdc: Done!

philn (approved):

aperezdc:
Thanks @philn 😃

PR #285, opened 2021-03-18
Support building documentation using gi-docgen

aperezdc:
Add support for building documentation for the `cogcore` library using [gi-docgen](https://gitlab.gnome.org/ebassi/gi-docgen), and the corresponding CI steps to build and archive the documentation as an artifact. In the future we could also setup automatic deployment of the generated content to `wpewebkit.org` or to GitHub pages when a push to the main branch happens and/or a tag is created.

A preview (not kept updated) can be found at https://fugu.perezdecastro.org/Cog-0.0/

For now there is not a lot documented, but at least we need to start somewhere 😄

### DIFF
--- a/.github/workflows/ci-cross.yml
+++ b/.github/workflows/ci-cross.yml
@@ -1,5 +1,5 @@
 ---
-name: CI
+name: "Build - ARM"
 on: [push]
 
 jobs:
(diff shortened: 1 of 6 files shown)


aperezdc:
Ah, on top of cross-toolchain not liking it that the the Python setup sets `LD_LIBRARY_PATH`, now that I think of it GObject Introspection does not play particularly well with cross compilation. Maybe I can add instead a new CI job that does a native build with documentation enabled, and keep the cross-builds with it disabled 🤔

aperezdc:
> Ah, on top of cross-toolchain not liking it that the the Python setup sets `LD_LIBRARY_PATH`, now that I think of it GObject Introspection does not play particularly well with cross compilation. Maybe I can add instead a new CI job that does a native build with documentation enabled, and keep the cross-builds with it disabled

This worked, and as a bonus we get to build Cog on the latest Ubuntu LTS, which is a nice check to have.

zdobersek (approved):

aperezdc:
Thanks @zdobersek, I will keep improving on top of this on future PRs 😄

PR #288, opened 2021-03-23
docs: Assorted documentation improvements

aperezdc:
This will add the following to the documentation:

- Documentation for most entities (classes, signals, functions, etc.) There are still some missing and the text could be a bit more elaborate, but IMO it's already much better than it used to be.
- An “Overview” page, which content lifted from `ARCHITECTURE.md` (which gets removed) plus some additions inspired by the GTK documentation.
- Stub pages for the different platform plug-ins, to be filled later.
- Logo!


aperezdc:
Ping reviewers 🏓

zdobersek (approved):

PR #293, opened 2021-03-29
add gitignore

akvadrako:


.gitignore
```diff
@@ -0,0 +1,12 @@
+CMakeFiles
+Makefile
+cmake_install.cmake
+libcog*.so*
+install_manifest.txt
+cogctl
+cogcore.pc
+cog-config.h
+cof
```
philn, line 9: ?
I'm not sure about this PR. Nowadays I would expect people to build in a separate folder, TBH.

philn (changes requested):

.gitignore
akvadrako: I was just following the README instructions. With meson, I agree, a separate folder is normal. With cmake I have no idea.

aperezdc: With CMake one can either do:

```sh
cmake -S sourcedir -B builddir
```

or, for older versions of CMake:

```sh
mkdir builddir
cd builddir
cmake ../path/to/srcdir
```

@akvadrako How about editing `README.md` to add a note about how to do out-of-tree builds with CMake? We could make that the recommended build instructions to encourage people to have a separate build directory.

akvadrako: @aperezdc Yes, I think that's a better idea.

aperezdc: @akvadrako Do you want to send a new PR for the `README.md` file yourself? That would be super nice 😄 

Also I suppose it's fine to close this one, so let's do that.

PR #299, opened 2021-04-20
Port to libsoup3

carlosgcampos:


philn:
Can we keep soup2 support for a while please? That would fix CI without much trouble, adding libsoup3 there might not be an easy task I'm afraid.

aperezdc (changes requested):
I agree with @philn that it would be neat to support both libsoup2 and libsoup3 for a while.

aperezdc:
I have fixed the code style issues reported by the linter, thanks @carlosgcampos for the initial patch and @philn for updating it with the conditional compilation guards 😺 — this should be ready to land now.

aperezdc (approved):

aperezdc:
I have fixed most of the style checker complaints but now there are false positives which would be avoided with #317 — so I am going to use Superpowers™ to land the PR as-is 😄

PR #300, opened 2021-04-21
drm: Document configuration file options

aperezdc:
SSIA.

PR #302, opened 2021-04-22
CI: Always use clang-format version 11

aperezdc:
This is needed to make the code style checker work as intended (see #300 for an example failure).


philn (approved):

aperezdc:
Thanks @philn!

PR #303, opened 2021-04-23
New platform: xcb (simplified X11)

agordon:
Hello Igalia devs,

Thanks for all your hard-work on WPE/COG - it's fantastic.

I'd like to offer a new platform plugin (temporary name: xcb).
It is conceptually similar to the current X11 platform, but works on minimal X11 servers without any extensions (e.g. GL/XKB).
It also supports some command-line options to enable easy startup in various kiosk modes.

Comments very welcomed.


aperezdc:
@agordon: Thanks for the PR, much appreciated! This new XCB platform plug-in has quite some overlap with the existing X11 one. I think it would make sense to merge both, for example:

- We could make XKB optional (it can be a build option) and use your key input handling from this PR if XKB is disabled and/or unavailable.
- When EGL fails to initialize in the existing X11 code, we could use your code from this PR to use SHM buffer exports.
- The existing X11 plug-in does not support fullscreening, we could import the related code from this PR for that.

Having only one platform plug-in for X11 would make the code easier to maintain and everybody using X11 would be running the same code, so bugs will be caught and fixed earlier.

How does that sound? Would you be willing to help merge both, with some guidance from our side? 🔧

agordon:
Yes, happy to work with you to consolidate it

aperezdc:
There is some nice code in this PR, so I have picked e.g. the `xcb-keysyms` code from here to add it into the existing X11 platform plug-in in #679 😺

PR #304, opened 2021-04-26
FrameBuffer Platform - WPE without X11

agordon:


agordon:
Hi,
I just checked this (framebuffer platform, no X11) on Raspberry PI zero - and it works nicely (although a bit slow and WPEWebProcess is at almost %100 cpu all the time..weird).

This is pretty cool, as it allows a minimal kiosk on rpi-zero, with no X11.

aperezdc:
Hello @agordon! First of all, thanks for the patches, and my apologies for taking time to get to them.

> I just checked this (framebuffer platform, no X11) on Raspberry PI zero - and it works nicely (although a bit slow and WPEWebProcess is at almost %100 cpu all the time..weird).

The slowness and CPU consumption would be explained by using software rendering, which is typically the case when using SHM buffer exports. Using the GPU for rendering would give you a performance boost, and it should be possible to make it work on the RaspberryPi Zero (more info below). Nowadays all [drivers are expected to use the DRM subsystem](https://lkml.org/lkml/2015/9/24/253) and even  “fbtft” drivers (for small e.g. SPI-driven TFT displays) have been [ported to DRM](https://github.com/notro/tinydrm/wiki).

But, if this makes things work in some cases, I would be okay merging this PR if somebody would commit to maintain the code… @agordon, would you be that person?

> This is pretty cool, as it allows a minimal kiosk on rpi-zero, with no X11.

While I do not own a RaspberryPi Zero to test things, I expect that the DRM plug-in would work there using Mesa's vc4 driver, and that would also run directly on the framebuffer, but with acceleration. In theory the GPU model is the same as for the Raspberry Pi3, which I have working with Mesa 😉 

Alternatively, when using the proprietary GPU driver I would recommend [WPEBackend-rdk](https://github.com/WebPlatformForEmbedded/WPEBackend-rdk) built with the `USE_BACKEND_BCM_RPI` option; then make a symlink named `libWPEBackend-default.so` pointing to `libWPEBackend-rdk.so`, and run Cog without passing a `--platorm=` option. This will use [dispmanx](https://jan.newmarch.name/RPi/Dispmanx/) which also works directly on framebuffer, and accelerated (no X11, and no Wayland).

On the other hand, it would be great to have the commit that adds the `-O`/`--platform-options=` command line option as a separate PR—I would merge that in the blink of an eye 😉

agordon:
Hi,
Thanks for the pointers -I'm going to check out the DRM and RDK/BCM method. If it works, the need for a pure "framebuffer" method is minimal.

I'll send a separate new PR for the `-O` thing.

aperezdc:
@agordon I suppose the DRM/KMS platform plug-in worked for you in the end, or some other of the options I suggested.

For now I am going to close this, because the code in the PR is outdated and won't work at the moment. If anybody would still find a legacy fbdev plug-in useful, please open a new PR, possibly using the code in this one as a starting point.

PR #305, opened 2021-04-27
Initial Webdriver support

lauromoura:
These commits add the basic plumbing needed to support running WebDriver scripts with Cog. For example, opening a page, following links and sending basic mouse and keyboard commands.

Still missing some documentation and examples of using it, alongside other bits like some visual feedback that the automation mode is enabled.

On the WebKit side, some harness patch will be needed to make the webdriver tests to run with `-DENABLE_COG` enabled.


core/cog-shell.c
```diff
@@ -215,11 +249,14 @@ cog_shell_constructed (GObject *object)
...
         g_build_filename (g_get_user_cache_dir (), priv->name, NULL);
 
-    g_autoptr(WebKitWebsiteDataManager) manager =
-        webkit_website_data_manager_new ("base-data-directory", data_dir,
+    g_autoptr(WebKitWebsiteDataManager) manager = NULL;
+
+    if (priv->automated)
+        manager = webkit_website_data_manager_new_ephemeral();
+    else
+        manager = webkit_website_data_manager_new ("base-data-directory", data_dir,
```
zdobersek, line 257: Nit: adjust the indentation of the following arguments.

cog.c
```diff
@@ -167,7 +170,9 @@ on_handle_local_options (GApplication *application,
     }
 
     const char *uri = NULL;
-    if (!s_options.arguments) {
+    if (s_options.automation) {
+        uri = "about:blank";
```
zdobersek, line 174: Could we just check whether CogShell object is automated? This would avoid the need for the option list entry.

cog.c
```diff
@@ -442,7 +448,17 @@ main (int argc, char *argv[])
         g_set_application_name ("Cog");
     }
 
-    g_autoptr(GApplication) app = G_APPLICATION (cog_launcher_get_default ());
+    // We need to check whether we'll use automation mode before creating the launcher
+    gboolean automated = FALSE;
+    for (int i=0; i < argc; i++) {
```
zdobersek, line 453: Nit: spacing around `i=0`. Also, it could start from 1 since [0] is the program name (as leveraged just above).

core/cog-shell.c
```diff
@@ -120,7 +126,28 @@ request_handler_map_entry_register (const char             *scheme,
...
+static WebKitWebView*
+cog_shell_create_web_view_for_automation (WebKitAutomationSession *session, CogShell *shell)
+{
+    return PRIV (shell)->web_view;
+}
+
+static void
+cog_shell_automation_started_callback (WebKitWebContext *context, WebKitAutomationSession* session, CogShell* shell)
+{
+    WebKitApplicationInfo* info = webkit_application_info_new();
```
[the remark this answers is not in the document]
lauromoura: Ok. Changed.

cog.c
lauromoura: Checking the CogShell for automation is also possible. I've updated the patch with it.

Regarding the option list entry, I thought about it more like a way to document this flag, keeping consistent with other arguments (even though it has this special check before the application is created).

aperezdc (commented):
Modulo style checker failures, patch looks reasonable to me. You can apply locally the suggestions from the style checker doing something like this, assuming you have a local branch derived from `master`:

```sh
data/check-style master | git apply -p0
```

PR #313, opened 2021-05-28
drm: Correctly set CMake target options

aperezdc:
Fix a typo in the call to `set_target_properties()`, which makes `libcogplatform-drm.so` be built into the expected location, and also correctly sets it to be compiled as C99.


philn (approved):

aperezdc:
Cherry picked into the `cog-0.10` branch as commit 3846cebd556cb6c9b9f77b1958ceb7ebd45fc028

PR #314, opened 2021-05-28
cog: Add support for loading a JSON content filter

aperezdc:
Add a new `-F`/`--content-filter` command line option to the `cog` launcher, which allows specifying the path to a JSON rule set which will be compiled and used as content filter.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -37,6 +37,10 @@ static struct {
         CogPlatform *platform;
     };
     union {
+        char *filter_path;
+        WebKitUserContentFilter *filter;
+    };
+    union {
         char *action_name;
         enum webprocess_fail_action action_id;
     } on_failure;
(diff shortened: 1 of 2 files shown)


cog.c
```diff
@@ -379,6 +394,11 @@ on_create_view (CogShell *shell, void *user_data G_GNUC_UNUSED)
                                                       "backend", view_backend,
                                                       NULL);
 
+    if (s_options.filter) {
+        WebKitUserContentManager *manager = webkit_web_view_get_user_content_manager(web_view);
+        webkit_user_content_manager_add_filter(manager, s_options.filter);
```
TingPing, line 399: You could free this after adding it

cog.c
```diff
@@ -273,6 +268,26 @@ on_handle_local_options (GApplication *application,
...
+        g_clear_pointer(&s_options.filter_path, g_free);
+
+        g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, FALSE);
+        webkit_user_content_filter_store_save_from_file(store,
+                                                        "CogFilter",
+                                                        file,
+                                                        NULL,
+                                                        (GAsyncReadyCallback) on_filter_saved,
+                                                        loop);
+        g_main_loop_run(loop);
```
zdobersek, line 288: Does this work as intended? The GTask spawned in `webkit_user_content_filter_store_save_from_file` will use the thread-default main context, but the global main context will be spun on the basis of passing `NULL` to `g_main_loop_new`.

Should a separate context be spawned and pushed-popped as the thread default? Otherwise spinning the current thread-default one will also be dispatching any other source otherwise associated with execution on this thread (i.e. something in WebKit or any platform-specific source dispatching e.g. windowing events).

aperezdc: Currently this seems to works but it may be as well a case of “works on my computer, dunno elsewhere”. I agree that it may be a good idea to go with your approach to make sure we only get the dispatch for the content filter without risking other events being handled as a side effect.

Thanks for the suggestion!

cog.c
aperezdc: Good catch, I'll post an update later covering this 👌🏼

cog.c
aperezdc: Mmmh, amusingly, creating a new `GMainContext` and doing the push/pop dance gets the process stuck. I suspect that inside WebKit we are doing something that uses the _default_ context (as opposed to the _thread default_ context), and by not spinning a main loop on the _default_ one we never get events dispatched—hence the lockup.

aperezdc: I think there might be some issue with WebKit's `RunLoop`/`WorkQueue` and how they deal with the _default_ main context and the _thread default_ main context (which are different things); but I do not have time at the moment to investigate further.

Truth is, code works fine using the _default_ main context, so unless somebody can find some case in which this won't work, I would say we merge the PR as-is.

Opinions?

aperezdc: Ping? 🏓

TingPing (approved):

PR #315, opened 2021-05-28
codestyle: Use Git's diff on CI with same options

aperezdc:
Make CI code style checks use as input the same diff that would have been generated locally when running `data/check-style` to avoid situations in which running the script on a developer's machine would report no mistakes but the CI would insist that there are still some code style issues.

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -17,7 +17,7 @@ jobs:
           sudo add-apt-repository \
             'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main'
           sudo apt update
-          sudo apt install -y clang-format-11 diffutils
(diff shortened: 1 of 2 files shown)


philn (approved):

aperezdc:
Sadly this unexpectedly [broke the CI](https://github.com/Igalia/cog/pull/314/checks?check_run_id=2704208437)… sometimes it's hard to predict whether a change to GitHub Actions will work after merging 😞 

I have reverted this in commit 61cbb2211adc0205ca924319c5abd1e5f908ce0a and I will try to come up with some alternative way in the next days. Probably the best approach would be to look harder into making sure that the Git clone done by the CI run contains both `master` and the branch with the changes so we can just run `data/check-style master` instead of relying on two clones and diffing them externally.

PR #317, opened 2021-05-31
CI: Fetch Git history to use data/check-style

aperezdc:
Arrange to fetch the Git history needed to make `git diff` work from a pull request checkout, so the data/check-style can be invoked directly instead of relying on an external diff tool. This makes the CI code style checks run the same command that a developer would use locally, thus avoiding the situation in which the CI would complain that there are code style issues that were not reported in a local run.


aperezdc:
Ping reviewers 🏓

philn (approved):

PR #320, opened 2021-06-22
Make platform plug-ins implement a GIO extension point

aperezdc:
Instead of using `dlopen`/`dlsym` manually, turn existing platform modules into implementations of a GIO “platform” extension point. The first commit adds the definition of the `CogPlatform` base class, the extension point, and utility functions to deal with modules which follow the structure of GLib's own.

The second commit modifies the platform plug-ins and the launcher to make use of the new interface. This is mostly mechanical work to apply the following changes to each platform plug-in:

- Add a subclass of `CogPlatform` e.g. `CogFdoPlatform` for the platform plug-in.
- Make callback functions private to the platform plug-ins and install them as vfuncs of the `CogPlatform` class.
- Support automatic selection of platform plug-ins by adding a new `is_suported` static vfunc, which is used to determine at run time which plug-in to use.
    
Note than other than the `is_supported` bit, no attempt is done at changing the architecture of the platform plug-ins. Further cleanups will be done later on separate PRs.

This is one feature extracted from the `devel/multiple-views` branch (see #77).


cog.c
```diff
@@ -289,31 +289,23 @@ platform_setup (CogShell *shell)
...
-    g_debug ("%s: Platform plugin: %s", __func__, platform_soname);
-
-    g_autoptr(CogPlatform) platform = cog_platform_new ();
-    if (!cog_platform_try_load (platform, platform_soname)) {
-        g_warning ("Could not load: %s (possible cause: %s).\n",
-                   platform_soname, strerror (errno));
+    g_autoptr(GError) error = NULL;
+    CogPlatform *platform = cog_platform_new(s_options.platform_name, &error);
+    if (!platform) {
+        g_warning("Cannot create platform: %s).\n", error->message);
```
TingPing, line 295: Maybe on purpose, but you don't need a trailing `\n`.

core/cog-modules.c
```diff
@@ -0,0 +1,156 @@
...
+ * vtable at which there is a function which takes no arguments and returns
+ * a boolean. This function will be called on each candidate implementation
+ * to check whether it is usable or not.
+ *
+ * Returns: (transfer none): A type that can be instantiated to implement
+ *    @extension_point, or %G_TYPE_INVALID if there are o usable
+ *    implementations.
+ */
+GType
+cog_modules_get_preferred(GIOExtensionPoint *extension_point, const char *preferred_module, size_t is_supported_offset)
```
TingPing, line 83: Convention for GIO modules in GLib is that they have an env var to override this.

See `_g_io_module_get_default_type()`.

TingPing:
Everything looks good to me. I'm a fan of using extension points :)

core/cog-modules.c
TingPing: GIO also adds `GIO_EXTRA_MODULES` so users can specify an extra directory to add to the search path. Similar could be useful.

aperezdc: Yes, that's the idea here.

A program using the `libcogcore` library can still choose to read an environment variable, if desired.

cog.c
aperezdc: Oh, and also there is a stray `)`, let's fix this.

TingPing (approved):

platform/drm/cog-platform-drm.c
```diff
@@ -1499,11 +1516,27 @@ on_export_shm_buffer (void* data, struct wpe_fdo_shm_exported_buffer *exported_b
...
 
-gboolean
-cog_platform_plugin_setup (CogPlatform *platform,
-                           CogShell    *shell,
-                           const char  *params,
-                           GError     **error)
+static void *
+check_supported(void *data G_GNUC_UNUSED)
+{
+    if (init_drm()) {
```
zdobersek, line 1522: `init_drm()` is way too heavy to act as a check here.

platform/fdo/cog-platform-fdo.c
```diff
@@ -1981,6 +1998,24 @@ clear_wayland (void)
     wl_display_disconnect (wl_data.display);
 }
 
+static void *
+check_supported(void *data G_GNUC_UNUSED)
+{
+    if (init_wayland(NULL)) {
```
zdobersek, line 2004: Same here, `init_wayland` does protocol roundtrips and loads cursor data. That's not viable for a simple check as intended here.

platform/x11/cog-platform-x11.c
```diff
@@ -887,3 +904,58 @@ cog_platform_plugin_get_view_backend (CogPlatform   *platform,
...
+    /*
+     * TODO: This could do more than only trying to connect to the X server.
+     */
+    Display *d = XOpenDisplay(NULL);
+    if (d) {
+        XCloseDisplay(d);
+        return GINT_TO_POINTER(TRUE);
+    } else {
+        return GINT_TO_POINTER(FALSE);
+    }
```
zdobersek, line 920: This approach seems just fine. If the X protocol can work, then it's probably usable for presenting. Same would be done for the FDO module and the Wayland protocol that's used there, but a higher priority. This would avoid preferring the X11 module being picked to connect to Xwayland over the FDO module connecting to the Wayland display.

platform/drm/cog-platform-drm.c
aperezdc: True, some of the code could be split and use it for the checks, instead of using the whole `init_drm()` function in the `is_supported` implementation. I've created issue #330 for that.

platform/fdo/cog-platform-fdo.c
aperezdc: Indeed, I have created #331 to take care of this.

platform/x11/cog-platform-x11.c
aperezdc: OTOH, it would be good to at least check whether the X server can do EGL + OpenGLES, though it may not be reasonably simple to do for the `is_supported` check.

Regarding priorities, if you set `G_MESSAGES_DEBUG=Cog` in the environment before running Cog it will print the list of modules found and their priorities. Currently it shows:

```
Cog-INFO: 22:07:59.456: com.igalia.Cog.Platform:
Cog-INFO: 22:07:59.456:   fdo - 500/CogFdoPlatform
Cog-INFO: 22:07:59.456:   gtk4 - 400/CogGtk4Platform
Cog-INFO: 22:07:59.456:   x11 - 300/CogX11Platform
Cog-INFO: 22:07:59.456:   drm - 200/CogDrmPlatform
Cog-INFO: 22:07:59.456:   headless - 100/CogHeadlessPlatform
```

As you can see, the `fdo` module already has more priority than `x11`, which means native Wayland is already preferred over XWayland 😉

PR #324, opened 2021-06-23
core: Add cog-host-routes-handler

psaavedra:
Handler for custom URI scheme requests that can route requests to different handlers depending on the *host* component of the requested URI.

Optionally, if a “fallback” handler has been specified, it will be used to serve requests which do not match any of the routed hosts.

The set of available host map entries can be configured using [method@Cog.HostMapHandler.add] and
[method@Cog.HostMapHandler.delete]. For each request, map entries  are checked and the one that matches the URI *host* component will handle the request.

This handler is typically used in tandem with [class@Cog.DirectoryFilesHandler], the latter being typically a
fallback, or as the handler for a routed host.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,6 +101,7 @@ set(COGCORE_API_HEADERS
     core/cog-launcher.h
     core/cog-request-handler.h
     core/cog-directory-files-handler.h
+    core/cog-host-routes-handler.h
     core/cog-prefix-routes-handler.h
     core/cog-shell.h
     core/cog-utils.h
@@ -114,6 +115,7 @@ set(COGCORE_SOURCES
     core/cog-launcher.c
     core/cog-request-handler.c
     core/cog-directory-files-handler.c
+    core/cog-host-routes-handler.c
     core/cog-prefix-routes-handler.c
     core/cog-utils.c
     core/cog-shell.c
--- a/core/cog-host-routes-handler.c
+++ b/core/cog-host-routes-handler.c
@@ -1,0 +1,296 @@
+/*
+ * cog-host-routes-handler.c
+ * Copyright (C) 2021 Igalia S.L.
+ *
+ * Distributed under terms of the MIT license.
+ */
+
+#include "cog-host-routes-handler.h"
+#include "cog-directory-files-handler.h"
+
+/**
+ * CogHostRoutesHandler:
+ *
+ * Direct custom URI scheme requests to different handlers.
+ *
+ * Handler for custom URI scheme requests that can route requests to
+ * different handlers depending on the *host* component of the requested
+ * URI.
+ *
+ * Optionally, if a “fallback” handler has been specified, it will be
+ * used to serve requests which do not match any of the routed hosts.
+ *
+ * The set of available host routes entries can be configured using
+ * [method@Cog.HostRoutesHandler.add] and
+ * [method@Cog.HostRoutesHandler.remove]. For each request, route entries
+ * are checked and the one that matches the URI *host* component
+ * will handle the request.
+ *
+ * This handler is typically used in tandem with
+ * [class@Cog.DirectoryFilesHandler], the latter being typically a
+ * fallback, or as the handler for a routed host.
+ *
+ * The following configures a handler to route requests with their host
+ * matching `example.host` to a certain directory, and all the rest of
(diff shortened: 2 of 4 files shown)


aperezdc:
I'll take a look today at this. In the meantime, if you want to fix the style checker, you download a patch stored as artifact [from the CI run](https://github.com/Igalia/cog/pull/324/checks?check_run_id=2892488037), or run the following from your work branch:

```sh
./data/check-style master | git apply -
```

core/cog-host-map-handler.c
```diff
@@ -0,0 +1,299 @@
...
+ *
+ * Adds a route to the handler.
+ *
+ * Configures a route which matches @host in URI, and dispatches
+ * requests to a given @handler.
+ *
+ * Returns: Whether the route was successfully added.
+ */
+gboolean
+cog_host_map_handler_mount(CogHostMapHandler *self, const char *host, CogRequestHandler *handler)
```
aperezdc, line 235: I would call this `cog_host_map_handler_add()` because the “mount” verb indicates that something is being attached to a location, and this handler is _not_ using the URI “location” (the path) at all, only the *host* part.

core/cog-host-map-handler.c
```diff
@@ -0,0 +1,299 @@
...
+ * Removes a previously configured route.
+ *
+ * Removes a route that matches @host previously configured
+ * using cog_host_map_handler_mount() or
+ * cog_host_map_handler_mount_path().
+ *
+ * Returns: Whether the route was found and removed.
+ */
+gboolean
+cog_host_map_handler_unmount(CogHostMapHandler *self, const char *host)
```
aperezdc, line 262: Same here, instead of using the “unmount” verb I think it would be better to call this `cog_host_map_handler_remove()`.

core/cog-host-map-handler.c
```diff
@@ -0,0 +1,299 @@
...
+ *
+ * This is a convenience method which configures a route
+ * matching @host in URI, and creates a new
+ * [class@Cog.DirectoryFilesHandler] for @base_path to handle requests
+ * for the route.
+ *
+ * Returns: Whether the route was successfully added.
+ */
+gboolean
+cog_host_map_handler_mount_path(CogHostMapHandler *self, const char *host, const char *base_path)
```
aperezdc, line 286: Ditto, `cog_host_map_handler_add_path()`.

core/cog-host-map-handler.c
```diff
@@ -0,0 +1,299 @@
...
+{
+    g_return_val_if_fail(COG_IS_HOST_MAP_HANDLER(self), FALSE);
+    g_return_val_if_fail(host != NULL, FALSE);
+    g_return_val_if_fail(base_path != NULL, FALSE);
+
+    g_autoptr(GFile) path_file = g_file_new_for_path(base_path);
+    g_return_val_if_fail(cog_directory_files_handler_is_suitable_path(path_file, NULL), FALSE);
+
+    g_autoptr(CogRequestHandler) handler =
+        g_object_new(COG_TYPE_DIRECTORY_FILES_HANDLER, "base-path", path_file, "strip-components", 0);
```
aperezdc, line 296: You could use `cog_directly_files_handler_new()` here for the sake of clarity. The `strip-components` property is zero by default so there is no need to use `g_object_new()` to set it during construction:

```suggestion
        cog_directory_files_handler_new(path_file);
```

aperezdc (changes requested):
I like the direction in which this is going and this is definitely reusable functionality that belongs in libcogcore 👌🏼 

There are a few suggestions and comments which I would like seeing addressed before merging, though.

Oh, one final note: I love it that the new class comes with API documentation, well done!

psaavedra:
All the previous comments addressed in the latest version of the changes. Additionally I've renamed the CogHostMapHandler to CogHostRoutesHandler. A more mimetic naming in comparison with other handlers already included in the code.

aperezdc:
Somehow I would still prefer a shorter name like `CogHostMapHandler` but let's leave it as-is because at some point I may want to rework the API (including the names) of all the classes used to support URI scheme request handlers—which would be a different PR 😉

aperezdc (approved):
FWIW, I did a rebase myself and fixed a small bit of the API docs, but otherwise I think this is ready for merging. Thanks @psaavedra!

PR #326, opened 2021-07-07
Update CONTRIBUTING.md

Nasah-Kuma:
Documented build/run process for cog paying more attention to ubuntu 20.04

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,6 +44,33 @@ make the code easier to navigate.
 
 *(TODO: Provide some notes on how to prepare a development environment which
 includes the needed dependencies.)*
+### Note
+This project requires libwpewebkit-1.0-dev, libwpe-1.0-dev, libwpebackend-fdo-1.0-dev 
+and GTK4 for the build/run process
+
+After cloning the project from [https://github.com/Igalia/cog] do the following 
+to build and get the project running
+-cd into project
+-create a build directory
+-cd into this directory. This is where your build files will be stored.
+-run cmake -DCOG_PLATFORM_X11=ON -DCOG_PLATFORM_GTK4=ON .. to generate the build files.
+-Then run the make command to generate the binaries
+You just finished building cog. To run it, run COG_MODULEDIR=$PWD/modules ./cog --platform=gtk4 [https://www.igalia.com/] 
+while in the build directory. If you are running Wayland, this will startup fine.
+
+### Building and Running Cog on Ubuntu 20.04
+It might generally be a bit of a problem getting and installing the dependencies 
+on this version.
+-In your project's root, run "sudo apt-get install libwpewebkit-1.0-dev libwpe-1.0-dev" 
+to install libwpewebkit-1.0-dev and libwpe-1.0-dev.
+-Next, cog requires a version of libwpebackend-fdo-1.0-dev >= 1.8.0
+To get this installed, checkout the repository at [vhttps://github.com/Igalia/WPEBackend-fdo].
+Switch to the wpebackend-fdo-1.10 branch(git checkout -t origin/wpebackend-fdo-1.10)
+where you can find the build directory for this version.
+-cd into this directory
+-run meson ..
+-run ninja
+
 
 Pull requests should be also prepared to be merged onto the `master` branch,
 except when the changes specifically apply to a release branch (like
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -44,6 +44,33 @@ make the code easier to navigate.
...
+to build and get the project running
+-cd into project
+-create a build directory
+-cd into this directory. This is where your build files will be stored.
+-run cmake -DCOG_PLATFORM_X11=ON -DCOG_PLATFORM_GTK4=ON .. to generate the build files.
+-Then run the make command to generate the binaries
+You just finished building cog. To run it, run COG_MODULEDIR=$PWD/modules ./cog --platform=gtk4 [https://www.igalia.com/] 
+while in the build directory. If you are running Wayland, this will startup fine.
+
+### Building and Running Cog on Ubuntu 20.04
```
Cwiiis, line 61: I would omit this section, given Ubuntu 20.04 doesn't have Gtk4 or a new enough GLib to build Gtk4. Alternatively, I expect one could use a Flatpak sandbox, or just build without the GTK backend, but that's probably beyond the scope of this documentation.

CONTRIBUTING.md
```diff
@@ -44,6 +44,33 @@ make the code easier to navigate.
 
 *(TODO: Provide some notes on how to prepare a development environment which
 includes the needed dependencies.)*
+### Note
+This project requires libwpewebkit-1.0-dev, libwpe-1.0-dev, libwpebackend-fdo-1.0-dev 
+and GTK4 for the build/run process
+
+After cloning the project from [https://github.com/Igalia/cog] do the following 
+to build and get the project running
+-cd into project
```
Cwiiis, line 53: Rather than using a list here (although I'm not sure if dashes transform to lists?), it would be nice to use the code markdown. See the documentation here: https://guides.github.com/features/mastering-markdown/

CONTRIBUTING.md
```diff
@@ -44,6 +44,33 @@ make the code easier to navigate.
...
+-create a build directory
+-cd into this directory. This is where your build files will be stored.
+-run cmake -DCOG_PLATFORM_X11=ON -DCOG_PLATFORM_GTK4=ON .. to generate the build files.
+-Then run the make command to generate the binaries
+You just finished building cog. To run it, run COG_MODULEDIR=$PWD/modules ./cog --platform=gtk4 [https://www.igalia.com/]
```
Cwiiis, line 58: I think if you actually ran this command with the square brackets it wouldn't work? Best to just leave out the square brackets, given it's a quoted command - similarly, code markdown would help here (i.e. surrounding the whole command with backticks).

CONTRIBUTING.md
```diff
@@ -44,6 +44,33 @@ make the code easier to navigate.
 
 *(TODO: Provide some notes on how to prepare a development environment which
 includes the needed dependencies.)*
+### Note
+This project requires libwpewebkit-1.0-dev, libwpe-1.0-dev, libwpebackend-fdo-1.0-dev
```
Cwiiis, line 48: It would be good to link to the respective GitHub repositories of these projects - perhaps present this as a bullet-point list. Maybe this could be split into separate 'Dependencies' and 'Building' sections?

Cwiiis (changes requested):
Good start, once these comments are addressed I think this will be good to go!

Nasah-Kuma:
Great thanks

CONTRIBUTING.md
Nasah-Kuma: Is there some document I can read to help me better explain how someone can use the flatpak sandbox and also explain building without GTK backend?

Cwiiis: The Flatpak documentation is at https://docs.flatpak.org/en/latest/ and there's a simple guide on building a Flatpak package from a source tarball here: https://opensource.com/article/19/10/how-build-flatpak-packaging but I don't know how much effort that requires, my previous attempts at using Flatpak for development never went that well (it has been a long time since then though). As for building Cog without GTK, you just need to omit the `-DCOG_PLATFORM_GTK4=ON` option (or explicitly set it to `OFF`, though it's off by default last I checked).

Cwiiis (changes requested):
Definitely almost there, thanks for the changes!

Cwiiis (approved):
This looks good to me.

aperezdc (approved):
Patch looks good, thanks!

One idea: we may want to include this as part of the generated documentation so the contents will also appear at https://igalia.github.io/cog — but let's do that in a separate patch.

PR #329, opened 2021-07-14
Implement DOM fullscreen requests in fdo backend

MortimerGoro:
- Enter fullscreen via DOM requests or keyboard key
- Exit fullscreen via keyboard key

### DIFF
--- a/platform/fdo/cog-platform-fdo.c
+++ b/platform/fdo/cog-platform-fdo.c
@@ -242,17 +242,25 @@ static struct {
 
     uint32_t width;
     uint32_t height;
+    uint32_t width_before_fullscreen;
+    uint32_t height_before_fullscreen;
 
     bool is_fullscreen;
+#if WPE_CHECK_VERSION(1, 11, 1)
     bool was_fullscreen_requested_from_dom;
+#endif
     bool is_resizing_fullscreen;
     bool is_maximized;
     bool should_resize_to_largest_output;
 } win_data = {
     .width = DEFAULT_WIDTH,
     .height = DEFAULT_HEIGHT,
+    .width_before_fullscreen = DEFAULT_WIDTH,
+    .height_before_fullscreen = DEFAULT_HEIGHT,
     .is_fullscreen = false,
+#if WPE_CHECK_VERSION(1, 11, 1)
     .was_fullscreen_requested_from_dom = false,
+#endif
     .is_resizing_fullscreen = false,
     .is_maximized = false,
     .should_resize_to_largest_output = false,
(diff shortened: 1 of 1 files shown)


MortimerGoro:
Depends on https://github.com/MortimerGoro/libwpe

platform/fdo/cog-platform-fdo.c
```diff
@@ -645,6 +649,64 @@ output_handle_scale (void *data,
...
+cog_fdo_does_image_match_win_size(struct wpe_fdo_egl_exported_image* image)
+{
+    return image && wpe_fdo_egl_exported_image_get_width(image) == win_data.width &&
+        wpe_fdo_egl_exported_image_get_height(image) == win_data.height;
+}
+
+static void
+cog_fdo_fullscreen_image_ready()
+{
+    xdg_toplevel_set_fullscreen(win_data.xdg_toplevel, NULL);
```
aperezdc, line 662: This only covers the case in which we are using the `xdg_wm_base` Wayland protocol. The `xdg_toplevel` member will be `NULL` in case we are using the `wl_shell` or `zwp_fullscreen_shell_v1`.

The order in which we prefer them is always this:

1. `xdg_wm_base`
2. `zwp_fullscreen_shell_v1`
3. `wl_shell`

You will need something like this:

```suggestion
    if (wl_data.xdg_shell) {
        xdg_toplevel_set_fullscreen(win_data.xdg_toplevel, NULL);
    } else if (wl_data.fshell) {
        return;  // Always fullscreen, nothing to do (?)
    } else if (wl_data.shell) {
        wl_shell_surface_set_fullscreen (win_data.shell_surface,
                                         WL_SHELL_SURFACE_FULLSCREEN_METHOD_SCALE,
                                         0,
                                         NULL);
    } else {
         g_assert_not_reached();
    }
```

...and similarly to un-fullscreen.

aperezdc (changes requested):
This is looking promising; the only remaining part being that it only works with the `xdg_wm_base` protocol, but the code is written to support also `wl_shell`, and `zwp_fullscreen_shell_v1`... Could you please take a look into adding support for them?

Thanks a ton for working on this 😃

platform/fdo/cog-platform-fdo.c
aperezdc: (If you ever wondered: yes, having to do `if`-`else` chains in every place where a Wayland “shell” operation is needed gets old and boring soon; I have ideas to improve that, see #336)

platform/fdo/cog-platform-fdo.c
```diff
@@ -2469,6 +2487,8 @@ cog_fdo_platform_get_view_backend(CogPlatform *platform, WebKitWebView *related_
                                      wpe_host_data.exportable);
     g_assert (wk_view_backend);
 
+    wpe_view_backend_set_fullscreen_handler(wpe_view_data.backend, cog_fdo_set_fullscreen, 0);
```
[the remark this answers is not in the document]
MortimerGoro: Done

platform/fdo/cog-platform-fdo.c
MortimerGoro: I added support for wl_shell and zwp_fullscreen_shell_v1 in the latest commit. I had to save the size before fullscreen to be able to get the previous state after exiting fullscreen. It seems that xdg_shell handles that automatically but not the others.

BTW I also noticed that the code uses a space before the `()`of each function call or declaration. I updated the new code to follow the same coding style.

aperezdc: We are slowly migrating to _not_ have a space before the function call parentheses, but only for new and modified code. That is why the code style checker failed.

If you have `clang-format` installed locally, you can run the following in your branch to apply the changes suggested by the style checker:

```sh
data/check-style master | git apply -p0
```

aperezdc (changes requested):
Other than the style checker issues, I think this is good to go—as long as others agree on the API addition from https://github.com/WebPlatformForEmbedded/libwpe/pull/86 😄

platform/fdo/cog-platform-fdo.c
MortimerGoro: Thanks for sharing that command!

I updated the code with the fixed styles and the function renames suggested by @carlosgcampos in https://github.com/WebPlatformForEmbedded/libwpe/pull/86

aperezdc (changes requested):
I will do the same comment as in #341: let's guard on `WPE_CHECK_VERSION(1, 11, 1)` to allow building Cog with an older libwpe. With that and making the style checker happy I think we can merge this 👍🏼

MortimerGoro:
> I will do the same comment as in #341: let's guard on `WPE_CHECK_VERSION(1, 11, 1)` to allow building Cog with an older libwpe. With that and making the style checker happy I think we can merge this 👍🏼

Done!

aperezdc (approved):
Thanks for doing one more last update to this PR, I think it's now ready 👌🏼

PR #339, opened 2021-08-03
fdo: Simplify checking for a running Wayland compsitor

### LINKED ISSUE #331: fdo: Simplify cog_fdo_platform_is_supported() function
Same here, `init_wayland` does protocol roundtrips and loads cursor data. That's not viable for a simple check as intended here.

_Originally posted by @zdobersek in https://github.com/Igalia/cog/pull/320#discussion_r657958209_

aperezdc:
Make the `cog_fdo_platform_is_supported()` check only whether it is possible to connect to a running compositor, without doing additional busy work.

Closes #331

### DIFF
--- a/platform/fdo/cog-platform-fdo.c
+++ b/platform/fdo/cog-platform-fdo.c
@@ -2001,8 +2001,14 @@ clear_wayland (void)
 static void *
 check_supported(void *data G_GNUC_UNUSED)
 {
-    if (init_wayland(NULL)) {
-        clear_wayland();
+    /*
+     * XXX: It would be neat to be able to determine if the compositor
+     *      supports the few protocols needed for operation, but that
+     *      would require protocol roundtrips.
+     */
+    struct wl_display *display = wl_display_connect(NULL);
+    if (display) {
+        wl_display_destroy(display);
         return GINT_TO_POINTER(TRUE);
     } else {
         return GINT_TO_POINTER(FALSE);


carlosgcampos (approved):

platform/fdo/cog-platform-fdo.c
```diff
@@ -2001,8 +2001,14 @@ clear_wayland (void)
 static void *
 check_supported(void *data G_GNUC_UNUSED)
 {
-    if (init_wayland(NULL)) {
-        clear_wayland();
+    /*
+     * XXX: It would be neat to be able to determine if the compositor
```
Cwiiis, line 2005: So it seems like the old code did this? What's the rationale behind this change, just to reduce round-trips?

Cwiiis (commented):
This new code is now, I suppose, less reliable at error reporting than the old code, but presumably faster? It'd be good to hear a bit about the rationale for the change.

platform/fdo/cog-platform-fdo.c
aperezdc: This was prompted by @zdobersek in [this](https://github.com/Igalia/cog/pull/320#discussion_r657957659) and [this comment](https://github.com/Igalia/cog/pull/320#discussion_r657958209) which introduced the `is_supported` vfunc to allow automatic detection of which plug-in to use. I will quote them here:

> init_drm() is way too heavy to act as a check here.

> Same here, init_wayland does protocol roundtrips and loads cursor data. That's not viable for a simple check as intended here.

Indeed, both `init_drm` and `init_wayland` do more work than strictly needed. In the case of the Wayland connection check I think it may be acceptable to do only one roundtrip to determine if at least one protocol among `xdg_wm_base`, `wl_shell`, and `zwp_fullscreen_shell_v1` are supported. The rest of protocols are either always available as per the Wayland spec (this is the case of `wl_shm` and `wl_seat`) or can be used optionally by Cog.

How about I redo this PR doing only the check for the shell protocols, with only one roundtrip?

aperezdc: I could even measure how long it takes to do the connection and a single roundtrip. I estimate it will be in the ballpark of tens of milliseconds, hundreds at most. If that's the case, to me that sounds acceptable.

aperezdc: > I could even measure how long it takes to do the connection and a single roundtrip. I estimate it will be in the ballpark of tens of milliseconds, hundreds at most. If that's the case, to me that sounds acceptable.

Well, I [did it](https://github.com/aperezdc/wl-connect-bench), and got this on my laptop:

```
 -> Benchmarking connect...
 -> 1 runs, 100000 iterations each run, finished in 1.449525 seconds
 -> 68988.12 i/sec

 -> Benchmarking connect_and_roundtrip...
 -> 1 runs, 100000 iterations each run, finished in 6.300510 seconds
 -> 15871.73 i/sec
```

So, while doing one roundtrip and checking for a few registry globals is 4x slower, it takes *0.63 ms* on average here (peanuts 🥜!)

Now, assuming that on an embedded device on the worst case this can be easily one order of magnitude slower, that would still be ~6.3 ms, which to me sounds very much acceptable.

zdobersek: A separate registry listener can be used to detect availability of specific protocols. Until then, only testing the connection to a wayland server is just fine for the purposes of probing the viability of running this module.

zdobersek (approved):

platform/fdo/cog-platform-fdo.c
aperezdc: It is fine by me to have merged this—I'll do a follow-up MR at some point to add only the checks for the minimum set of required Wayland protocols 👌🏼

PR #340, opened 2021-08-05
Nasahdev format style

Nasah-Kuma:
Added pre-commit hook

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ set(COG_VERSION_EXTRA "")
 if (IS_DIRECTORY "${CMAKE_SOURCE_DIR}/.git")
     set(COG_VERSION_EXTRA "+git")
     find_package(Git)
+    file(COPY ${CMAKE_SOURCE_DIR}/hooks/pre-commit DESTINATION ${CMAKE_SOURCE_DIR}/.git/hooks FILE_PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ)
     if (GIT_FOUND)
         execute_process(
             COMMAND "${GIT_EXECUTABLE}" rev-list --max-count=1 --abbrev-commit HEAD
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,1 @@
+git diff --staged | data/check-style --diff


CMakeLists.txt
```diff
@@ -204,6 +204,7 @@ install(FILES ${COGCORE_API_HEADERS}
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cog
     COMPONENT "development"
 )
+file(COPY hooks/pre-commit DESTINATION ../.git/hooks)
```
Cwiiis, line 207: The relative path here is a bit concerning, I'm guessing this is relative to the build directory. We need this to be relative to the source directory - further up in the file, you can see `"$(CMAKE_SOURCE_DIR}/.git" referenced, we need to do the same here. In fact, this copy should probably go in that same block, we only ever want to do this for git checkouts and not source tarballs.

hooks/pre-commit
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env bash
+chmod +x
```
Cwiiis, line 2: I don't think this line does anything on its own?

hooks/pre-commit.hook
```diff
@@ -0,0 +1 @@
+git diff -U0 HEAD~2 HEAD | data/check-style --diff
```
Cwiiis, line 1: I'm confused as to why there's both hooks/pre-commit and hooks/pre-commit.hook - it seems the latter is redundant?

hooks/pre-commit
```diff
@@ -0,0 +1,3 @@
...
+git diff -U0 HEAD~2 HEAD | data/check-style --diff
```
Cwiiis, line 3: Should this definitely be HEAD~2 and not HEAD~1? I don't know much about pre-commit hooks, but if this is run before the commit, won't this be comparing the last 2 commits to the most recent and ignoring current changes? (I may well be wrong here, but would like to hear the explanation if so)

Cwiiis (changes requested):
Almost there! As well as the comments I've added, this should also be just one commit.

hooks/pre-commit
Nasah-Kuma: Maybe git diff | data/check-style --diff will solve the problem given that we want to check the style of uncommitted changes

Cwiiis: I guess you'd need to diff the staging area, so I think `git diff --staged` rather than just `git diff` - I'm not sure if that's exactly right, but I'm sure some experimentation will reveal the answer :)

Nasah-Kuma: I think` git diff --staged` works. I noticed that the style checking is however not consistent

Nasah-Kuma: check-style gives different outputs with the same change when tried at different times

hooks/pre-commit
```diff
@@ -1 +1 @@
-git diff -U0 HEAD~2 HEAD | data/check-style --diff
+git diff | data/check-style --diff
```
Cwiiis, line 1: This is missing --staged?

hooks/pre-commit
Cwiiis: Could you give an example of these different outputs vs the input and what you expect? Feel free to use https://gist.github.com/ if there's a lot of information you need to share.

CMakeLists.txt
Cwiiis: It looks like squashing commits has brought this line back - the correct line is above, so this just needs removing

core/cog-launcher.h
```diff
@@ -21,21 +21,20 @@ G_BEGIN_DECLS
 
 typedef struct _CogRequestHandler CogRequestHandler;
 
-#define COG_TYPE_LAUNCHER (cog_launcher_get_type ())
+#define COG_TYPE_LAUNCHER (cog_launcher_get_type())
```
Cwiiis, line 24: This file shouldn't be changed at all, so this needs to be removed from this commit.

.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
Cwiiis, line 1: This change shouldn't be in this commit.

Cwiiis (changes requested):
It looks like squashing commits brought back some earlier commits, possibly the tree was in a weird state but it should be easy to resolve from this point. This commit just needs some lines removed and it's good to go I think.

aperezdc (approved):
Looks good to me, too!

Thanks @Nasah-Kuma!

Issue #287, opened 2021-03-21
gtk4: Fullscreen API support


philn:
For this we will need to enable the support for the Fullscreen API in WebKit, which is currently unsupported for the WPE port, and provide a way to listen to fullscreening requests for a web view (likely through libwpe's view backend).

_Originally posted by @aperezdc in https://github.com/Igalia/cog/pull/262#discussion_r595124820_

PR #341, opened 2021-08-05
Implement fullscreen in GTK4 platform

MortimerGoro:
Fixes https://github.com/Igalia/cog/issues/287

### DIFF
--- a/platform/gtk4/cog-platform-gtk4.c
+++ b/platform/gtk4/cog-platform-gtk4.c
@@ -36,8 +36,6 @@ G_DEFINE_DYNAMIC_TYPE_EXTENDED(
 
 /*
  * TODO
- * - fullscreen: if a video element switches to fullscreen it would be nice to also fullscreen the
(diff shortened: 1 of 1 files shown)


aperezdc (changes requested):
Could we have the usage of the new libwpe API guarded with `#if WPE_CHECK_VERSION(1, 11, 1)`? We would like to allow building Cog with the last two or three stable series of the dependency libraries.

Other than that, I like this patch and it goes to show how much GTK does for us under the hood 😄

MortimerGoro:
> Could we have the usage of the new libwpe API guarded with `#if WPE_CHECK_VERSION(1, 11, 1)`? We would like to allow building Cog with the last two or three stable series of the dependency libraries.
> 
> Other than that, I like this patch and it goes to show how much GTK does for us under the hood smile

Thanks for the review. I already added the WPE version check guards, and fixed the style checks

aperezdc (approved):
Thanks for the update! 🎉

PR #345, opened 2021-08-09
fdo: Reintroduce checks on Wayland protocols

aperezdc:
Add code to do a cursory check on the Wayland protocols handled by the Wayland compositor in the `check_supported()` function. This makes sure that the FDO platform will be chosen if at least one of the shell protocols supported by the code are handled by the compositor.

Note that it is not needed to check for protocols from the base Wayland specification which every compositor is expected to implement (`wl_compositor`, `wl_shm`, `wl_seat`, etc.) because these will always be available.

As a related note, it would be neat to check that the EGL will be usable with the Wayland compositor, but there is no 100% reliable way without going through actual EGL initialization, so no check is added for this at the moment.

### DIFF
--- a/platform/fdo/cog-platform-fdo.c
+++ b/platform/fdo/cog-platform-fdo.c
@@ -2092,18 +2092,71 @@ clear_wayland (void)
     wl_display_disconnect (wl_data.display);
 }
 
+// clang-format off
+#define SHELL_PROTOCOLS(m) \
+    m(xdg_wm_base)         \
+    m(wl_shell)            \
+    m(zwp_fullscreen_shell_v1)
+// clang-format on
+
+#define DECLARE_PROTOCOL_ENTRY(proto) gboolean found_##proto;
+struct check_supported_protocols {
+    SHELL_PROTOCOLS(DECLARE_PROTOCOL_ENTRY)
+};
+#undef DECLARE_PROTOCOL_ENTRY
+
+static void
+on_registry_global_is_supported_check(void *data,
+                                      struct wl_registry *registry,
+                                      uint32_t name,
+                                      const char *interface,
+                                      uint32_t version)
+{
+    struct check_supported_protocols *protocols = data;
+
+#define TRY_MATCH_PROTOCOL_ENTRY(proto)                   \
+    if (strcmp(interface, proto##_interface.name) == 0) { \
+        protocols->found_##proto = TRUE;                  \
+        return;                                           \
+    }
+
+    SHELL_PROTOCOLS(TRY_MATCH_PROTOCOL_ENTRY)
+
+#undef TRY_MATCH_PROTOCOL_ENTRY
+}
+
 static void *
 check_supported(void *data G_GNUC_UNUSED)
 {
     /*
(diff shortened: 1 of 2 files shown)


platform/fdo/cog-platform-fdo.c
```diff
@@ -2092,18 +2092,68 @@ clear_wayland (void)
...
+     *      drivers might expose other protocols (w.g. wl_viv for Vivante)
+     *      so that could result in needing to maintain the list updated.
+     *
+     *      For now the check assumes that EGL will work if the compositor
+     *      handles at least one of the shell protocols supported by Cog.
      */
     struct wl_display *display = wl_display_connect(NULL);
     if (display) {
+        struct check_supported_protocols protocols = {};
+        wl_registry_add_listener(wl_display_get_registry(display),
```
zdobersek, line 2138: wl_registry is leaked

platform/fdo/cog-platform-fdo.c
```diff
@@ -2092,18 +2092,68 @@ clear_wayland (void)
...
+static void
+on_registry_global_is_supported_check(void *data, struct wl_registry *registry,
+                                      uint32_t name, const char *interface,
+                                      uint32_t version)
+{
+    struct check_supported_protocols *protocols = data;
+
+#define TRY_BIND_PROTOCOL_ENTRY(proto) \
+    if (strcmp(interface, proto ## _interface.name) == 0) { \
+        protocols->proto = wl_registry_bind(registry, name, &proto ## _interface, version); \
```
zdobersek, line 2113: Binding shouldn't be needed. But whatever is bound has to be cleaned up.

platform/fdo/cog-platform-fdo.c
```diff
@@ -2092,18 +2092,68 @@ clear_wayland (void)
...
+                                   .global = on_registry_global_is_supported_check,
+                                   }),
+                                 &protocols);
+        wl_display_roundtrip(display);
+
+        gboolean ok = FALSE;
+#define CHECK_SHELL_PROTOCOL(proto) \
+        if (protocols.proto) { \
+            ok = TRUE; \
+            goto check_done; \
```
zdobersek, line 2149: Taking this jump leaks the wl_display.

zdobersek (changes requested):

platform/fdo/cog-platform-fdo.c
aperezdc: You are right, it's enough to know that the protocol is supported by the compositor. We can have a boolean flag and toggle it when the interface name is matched.

platform/fdo/cog-platform-fdo.c
aperezdc: Fixed.

platform/fdo/cog-platform-fdo.c
aperezdc: Fixed.

platform/fdo/cog-platform-fdo.c
aperezdc: ...and done!

zdobersek (approved):

PR #346, opened 2021-08-12
Nasahdev deprecation

Nasah-Kuma:
This update fixes the only deprecation warning I saw when the application was built. Please let me know if there are others.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -263,10 +263,9 @@ on_handle_local_options (GApplication *application,
                                                          s_options.web_extensions_dir);
     }
 
-    webkit_web_context_set_tls_errors_policy (cog_shell_get_web_context (shell),
-                                              s_options.ignore_tls_errors
-                                              ? WEBKIT_TLS_ERRORS_POLICY_IGNORE
-                                              : WEBKIT_TLS_ERRORS_POLICY_FAIL);
+    webkit_website_data_manager_set_tls_errors_policy(
+        webkit_web_context_get_website_data_manager(cog_shell_get_web_context(shell)),
+        s_options.ignore_tls_errors ? WEBKIT_TLS_ERRORS_POLICY_IGNORE : WEBKIT_TLS_ERRORS_POLICY_FAIL);
 
     if (s_options.filter_path) {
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
philn, line 1: This is specific to your setup so I think it should be mentioned in your `core.excludesFile` variable in your `~/.gitconfig` file. We can't list here all combinations of build directory names :)

cog.c
```diff
@@ -253,20 +253,18 @@ on_handle_local_options (GApplication *application,
             return EXIT_FAILURE;
         }
 
-        g_object_set (shell, "config-file", g_key_file_ref (key_file), NULL);
```
philn, line 256: Can you split this PR in 2 commits please? One with the style fixes and one with the deprecation warning fixes.

philn (changes requested):

cog.c
Nasah-Kuma: The pre-commit hook prevents you from committing anything if your code format isn't what is required. When I made the modifications on cog.c, I couldn't commit unless the code-style format in this file matched what is required so it might be complicated separating the changes into two commits since one change caused the other.

Nasah-Kuma: The style fixes on cog-handler.h can be put in another commit but the style fixes on cog.c were made because of the change on the file. The pre-commit hook prevents committing of any change not in the right code format. After I made the change on cog.c, I hard to fix the code-style before the commit was accepted.

philn: You can skip the pre-commit hooks with `git commit -n ...`

philn (approved):

aperezdc:
Thanks, @Nasah-Kuma!

philn:
Broke the build, see https://github.com/Igalia/cog/pull/348
I wonder why CI didn't detect this?

aperezdc:
> Broke the build, see #348
> I wonder why CI didn't detect this?

It was because of #349 — thanks for figuring out how to make the CI run for PRs for forked repos 😄

PR #348, opened 2021-08-18
cog: Fix build for WebKit versions inferior to 2.32

philn:
webkit_web_context_set_tls_errors_policy() was introduced in 2.32,
can't be used for older versions...


Cwiiis (approved):
LGTM

PR #353, opened 2021-08-24
drm: simplify the check-if-supported operation

zdobersek:
Don't perform the complete DRM initialization to test whether the DRM module can
be used, instead only iterate through the retrieved DRM devices and indicate as
supported if a primary DRM node was detected.


aperezdc (approved):
Thanks!

PR #356, opened 2021-08-30
docs: webdriver: add missing imports

Bastian-Krause:
This makes the Python example work out of the box.

Fixes #305

@lauromoura Thanks for you work!


aperezdc (approved):
Thanks @Bastian-Krause!

PR #359, opened 2021-09-14
Factor out switcheroos for the different wayland shell protocols

Nasah-Kuma:


philn:
The `rename fdo to wl in the source and build files` commit was merged in master. Can you rebase please?

Nasah-Kuma:
> The `rename fdo to wl in the source and build files` commit was merged in master. Can you rebase please?

Ok I have

code-style.diff
```diff
@@ -0,0 +1,54 @@
+--- cog.c	(before formatting)
```
Cwiiis, line 1: I assume this file isn't intended to be in this commit.

platform/wayland/cog-platform-wl.c
```diff
@@ -757,14 +961,30 @@ registry_global (void               *data,
         wl_data.subcompositor = wl_registry_bind(registry, name, &wl_subcompositor_interface, version);
     } else if (strcmp(interface, wl_shell_interface.name) == 0) {
         wl_data.shell = wl_registry_bind(registry, name, &wl_shell_interface, version);
+        win_data.shell_functions->enter_fullscreen = &wl_shell_enter_fullscreen;
```
Cwiiis, line 964: Rather than setting each function, it might be worthwhile just having each set of functions pre-defined, so something like:
```
   static const shell_operations wayland_shell_operations = {
     .enter_fullscreen = &wl_shell_enter_fullscreen,
     .exit_fullscreen = &wl_shell_exit_fullscreen,
     ...
   };
   
   ...
   
     win_data.shell_functions = &wayland_shell_operations
```
As it is, `win_data.shell_functions` is a pointer and this is referencing invalid memory as I don't see anywhere where `shell_functions` is set.

platform/wayland/cog-platform-wl.c
```diff
@@ -1968,24 +2188,18 @@ init_wayland (GError **error)
 }
 
 static void
-clear_wayland (void)
+clear_wayland(void)
 {
     g_clear_pointer(&wl_data.event_src, g_source_destroy);
+    win_data.shell_functions->destroy_shell();
```
Cwiiis, line 2194: Perhaps it's worth clearing the `shell_functions` pointer after calling `destroy_shell` - this would make the situation where a shell function is erroneously called after this function easier to debug.

platform/wayland/cog-platform-wl.c
```diff
@@ -132,16 +177,24 @@ struct shm_buffer {
 #endif
 
 #ifndef EGL_WL_create_wayland_buffer_from_image
-typedef struct wl_buffer * (EGLAPIENTRYP PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL) (EGLDisplay dpy, EGLImageKHR image);
+typedef struct wl_buffer *(EGLAPIENTRYP PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL)(EGLDisplay dpy, EGLImageKHR image);
 #endif
 
+typedef struct shell_operations {
```
Cwiiis, line 183: I think it might be good to be more consistent with naming - we have `shell_operations * shell_functions` just below here. Also, I notice that two extra xdg-related pointers are added to the `win_data` struct. Maybe it would be better to have a structure "shell_context" that encapsulates both the functions and any implementation-specific data necessary, and calling this "shell_protocol". Use of unions here might save some memory and make organisation a bit easier, for example:

```
typedef struct {
  void (*enter_fullscreen)();
  void (*exit_fullscreen)();
} ShellFunctions;

struct xdg_shell_protocol = {
  ShellFunctions functions;
  struct xdg_surface * xdg_surface;
  struct xdg_toplevel * xdg_toplevel;
};

struct f_shell_protocol {
  ShellFunctions functions;
  // ... Any data that fshell needs specifically ...
};

typedef union {
  ShellFunctions functions;
  struct xdg_shell_protocol xdg_context;
} ShellContext;

static ShellContext xdg_shell = {
  .functions = {
    &xdg_enter_fullscreen,
    &xdg_exit_fullscreen
  }
};
```

(excuse any incorrect syntax here, let's call this pseudo-code :))
So after this, you could have a `ShellContext* ctx` in win_data, and accessing functions would work much the same, like `win_data.ctx->functions.enter_fullscreen()` and individual data could be accessed like `win_data.ctx->xdg_context.xdg_surface`.

Cwiiis (commented):
This is a good start. The two biggest correctness issues here are that the shell functions pointer is accessed without allocating it (null pointer dereference), as mentioned below, and also that in registry_global, if none of the cases that set shell functions are hit, the shell functions will be undefined (rather than leading to a default set of functions that would print an error message, or some other appropriate action). In practice, both of these issues would result in crashing in the best case.

Once these issues are tackled, the next big thing will be figuring out how to get these into separate files. Separating the data necessary for these functions from win_data will play a large part in that, but let's tackle that when we get there!

platform/wayland/cog-platform-wl.c
```diff
@@ -85,6 +85,51 @@ struct _CogWlPlatformClass {
 struct _CogWlPlatform {
     CogPlatform parent;
 };
+static void resize_to_largest_output();
+static void resize_window();
+static void configure_surface_geometry();
+static void xdg_surface_on_configure(void *data, struct xdg_surface *surface, uint32_t serial);
```
Cwiiis, line 91: Just for ease of reading, it'd be nice to put an empty line between the forward declarations above and the shell function definitions here.

platform/wayland/cog-platform-wl.c
```diff
@@ -85,6 +85,51 @@ struct _CogWlPlatformClass {
...
+static void xdg_toplevel_on_configure(void *               data,
+                                      struct xdg_toplevel *toplevel,
+                                      int32_t              width,
+                                      int32_t              height,
+                                      struct wl_array *    states);
+static void xdg_toplevel_on_close(void *data, struct xdg_toplevel *xdg_toplevel);
+static void shell_surface_ping(void *data, struct wl_shell_surface *shell_surface, uint32_t serial);
```
Cwiiis, line 98: Having the xdg and shell surface/popup listener declarations mixed together like this is a little confusing, better to keep them together and separated by newlines to make reading easier. I would also consider putting them closer to where they're used, rather than just at the very top like this.

platform/wayland/cog-platform-wl.c
```diff
@@ -373,6 +616,23 @@ wl_src_finalize(GSource *base)
 {
 }
 
+static void
+default_no_shell(shell_functions *shell_functions)
```
Cwiiis, line 620: Not sure what this function is actually intended to do at the moment.

platform/wayland/cog-platform-wl.c
```diff
@@ -757,14 +977,29 @@ registry_global (void               *data,
         wl_data.subcompositor = wl_registry_bind(registry, name, &wl_subcompositor_interface, version);
     } else if (strcmp(interface, wl_shell_interface.name) == 0) {
         wl_data.shell = wl_registry_bind(registry, name, &wl_shell_interface, version);
+        if (wl_data.shell != NULL) {
+            win_data.shell_functions = &wayland_shell_functions;
+        } else {
+            default_no_shell(&wayland_shell_functions);
```
Cwiiis, line 983: I guess default_no_shell should be setting up default shell functions, but currently it checks whether various function pointers exist in the given shell_functions struct and then calls the default implementation of the first one it finds and returns?

I would suggest that much like you now have wayland_shell_functions, xdg_shell_functions and f_shell_functions, you should also have no_shell_functions (or whatever you'd like to call it) and that you just set that up immediately before this big if/else block. That would mean you no longer need all these else clauses.

I may have misinterpreted the intension of default_no_shell though, please let me know if so!

Cwiiis (commented):
Definitely moving in the right direction - some of my comments from before still apply and I've added some new ones for this current revision.

platform/wayland/cog-platform-wl.c
```diff
@@ -132,16 +159,56 @@ struct shm_buffer {
...
+    void (*enter_fullscreen)();
+    void (*exit_fullscreen)();
+    void (*destroy_shell)();
+    void (*maximize_surface)();
+    void (*create_window)();
+    void (*create_popup)();
+} shell_functions;
+
+typedef struct {
+    shell_functions *    functions;
```
Cwiiis, line 175: I don't think there's any reason to make this a pointer.

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +339,248 @@ static struct {
...
+    wl_surface_commit(popup_data.shell_context->wl_shell_data.wl_surface);
+}
+
+static shell_functions xdg_shell_functions = {.enter_fullscreen = &xdg_shell_enter_fullscreen,
+                                              .exit_fullscreen = &xdg_shell_exit_fullscreen,
+                                              .destroy_shell = &destroy_xdg_shell,
+                                              .maximize_surface = &maximize_xdg_shell_surface,
+                                              .create_window = &create_xdg_shell_window,
+                                              .create_popup = &create_xdg_shell_popup};
+xdg_shell_data         xdg_data = {.functions = &xdg_shell_functions};
```
Cwiiis, line 434: If functions wasn't a pointer, you could just declare the functions struct directly here;
```
.functions = { .enter_fullscreen = ...,
                      ... }
```

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +339,248 @@ static struct {
...
+    g_warning("No available shell capable of fullscreening.");
+    win_data.is_fullscreen = false;
+}
+static void
+no_shell_exit_fullscreen()
+{
+    g_assert_not_reached();
+}
+static void
+maximize_no_shell()
```
Cwiiis, line 561: To be consistent with what you've done above, this should be called `no_shell_maximize`.

platform/wayland/cog-platform-wl.c
```diff
@@ -1968,24 +2188,18 @@ init_wayland (GError **error)
 }
 
 static void
-clear_wayland (void)
+clear_wayland(void)
 {
+    win_data.shell_context->functions.destroy_shell();
```
Cwiiis, line 2193: Just echoing my previous comment on this, I think it would be good to clear the shell_context pointer after calling destroy_shell.

platform/wayland/cog-platform-wl.c
```diff
@@ -2125,108 +2339,44 @@ clear_egl (void)
...
 static gboolean
-create_window (GError **error)
+create_window(GError **error)
 {
-    g_debug ("Creating Wayland surface...");
+    g_debug("Creating Wayland surface...");
 
-    win_data.wl_surface = wl_compositor_create_surface (wl_data.compositor);
-    g_assert (win_data.wl_surface);
+    win_data.shell_context->wl_shell_data.wl_surface = wl_compositor_create_surface(wl_data.compositor);
```
Cwiiis, line 2346: Why has wl_surface been moved to wl_shell_data? It doesn't seem to be specific to wl_shell and putting it here means that you're going to corrupt data in some way or another when using any shell that isn't wl_shell.

platform/wayland/cog-platform-wl.c
```diff
@@ -2125,108 +2339,44 @@ clear_egl (void)
...
 
 static void
-destroy_window (void)
+destroy_window(void)
 {
-    g_clear_pointer (&win_data.xdg_toplevel, xdg_toplevel_destroy);
-    g_clear_pointer (&win_data.xdg_surface, xdg_surface_destroy);
-    g_clear_pointer (&win_data.shell_surface, wl_shell_surface_destroy);
-    g_clear_pointer (&win_data.wl_surface, wl_surface_destroy);
+    g_clear_pointer(&win_data.shell_context->xdg_shell_data.xdg_toplevel, xdg_toplevel_destroy);
```
Cwiiis, line 2373: `shell_context` is only going to be one type of shell, accessing it as multiple types like this almost certainly means you're calling one or more of these functions with invalid data. This should be factored out into a surface_destroy function in shell_functions.

platform/wayland/cog-platform-wl.c
```diff
@@ -2236,124 +2386,76 @@ create_popup (WebKitOptionMenu *option_menu)
...
+    webkit_option_menu_close(popup_data.option_menu);
+    g_clear_pointer(&popup_data.popup_menu, cog_popup_menu_destroy);
+    g_clear_object(&popup_data.option_menu);
 
-    g_clear_pointer (&popup_data.xdg_popup, xdg_popup_destroy);
-    g_clear_pointer (&popup_data.xdg_surface, xdg_surface_destroy);
-    g_clear_pointer (&popup_data.xdg_positioner, xdg_positioner_destroy);
-    g_clear_pointer (&popup_data.shell_surface, wl_shell_surface_destroy);
-    g_clear_pointer (&popup_data.wl_surface, wl_surface_destroy);
+    g_clear_pointer(&popup_data.shell_context->xdg_shell_data.xdg_popup, xdg_popup_destroy);
```
Cwiiis, line 2410: Same with this block, you can't access multiple members of `shell_context` like this, only one is going to be valid.

Cwiiis (commented):
I think there's been a misunderstand of how unions work here, as there are multiple parts of the code that access multiple members of it. Only one member of the shell_context union is going to be valid for the lifecycle of the application - only the shell_functions member should be accessed directly by code outside of shell functions (inside of which you can know which member of the union is valid).

Are you testing cog after making these changes? It seems unlikely that you wouldn't see crashing - make sure to test at least a couple of backends as you work, most of these issues would very likely be flagged by basic testing.

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+    configure_surface_geometry(win_data.width_before_fullscreen, win_data.height_before_fullscreen);
+    resize_window();
+}
+static void
+destroy_wl_shell()
+{
+    wl_shell_destroy(wl_data.shell);
+}
+static void
+maximize_wl_shell_surface()
```
Cwiiis, line 468: wl_shell_maximize_surface

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+    wl_shell_surface_set_maximized(win_data.shell_context->wl_shell_data.shell_surface, NULL);
+}
+
+static const struct wl_shell_surface_listener shell_surface_listener = {
+    .ping = shell_surface_ping,
+    .configure = shell_surface_configure,
+};
+
+static void
+create_wl_shell_window()
```
Cwiiis, line 479: wl_shell_create_window

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+}
+
+static const struct wl_shell_surface_listener shell_popup_surface_listener = {
+    .ping = shell_popup_surface_ping,
+    .configure = shell_popup_surface_configure,
+    .popup_done = shell_popup_surface_popup_done,
+};
+
+static void
+create_wl_shell_popup()
```
Cwiiis, line 499: wl_shell_create_popup

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+
+    wl_shell_surface_add_listener(popup_data.shell_surface, &shell_popup_surface_listener, NULL);
+    wl_shell_surface_set_popup(popup_data.shell_surface, wl_data.seat, wl_data.event_serial, win_data.wl_surface, 0,
+                               (win_data.height - popup_data.height), 0);
+
+    display_popup();
+}
+
+static void
+destroy_shell_surface()
```
Cwiiis, line 512: wl_shell_surface_destroy

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+    resize_to_largest_output();
+}
+static void
+f_shell_exit_fullscreen()
+{
+    configure_surface_geometry(win_data.width_before_fullscreen, win_data.height_before_fullscreen);
+    resize_window();
+}
+static void
+destroy_f_shell()
```
Cwiiis, line 539: f_shell_destroy_shell

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+{
+    zwp_fullscreen_shell_v1_destroy(wl_data.fshell);
+}
+static void
+create_f_shell_window()
```
Cwiiis, line 544: f_shell_create_window

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+                                            win_data.wl_surface,
+                                            ZWP_FULLSCREEN_SHELL_V1_PRESENT_METHOD_DEFAULT,
+                                            NULL);
+
+    /* Configure the surface so that it respects the width and height
+        * environment variables */
+    configure_surface_geometry(0, 0);
+}
+
+f_shell_data f_data = {.functions = {.exit_fullscreen = &f_shell_exit_fullscreen,
```
Cwiiis, line 556: There are a bunch of missing functions here.

platform/wayland/cog-platform-wl.c
```diff
@@ -281,22 +342,258 @@ static struct {
...
+                                     .destroy_shell = &destroy_f_shell,
+                                     .create_window = &create_f_shell_window,
+                                     .enter_fullscreen = &f_shell_enter_fullscreen}};
+
+//default executed when no shell exist
+static void
```
Cwiiis, line 562: I would declare these before all the shell implementations, so that when shells are missing capabilities (like with f_shell above), you can reference these default implementations. I also note that there doesn't seem to be a full set of functions or an accompanying shell_functions struct for these.

platform/wayland/cog-platform-wl.c
```diff
@@ -1794,12 +2026,12 @@ on_export_shm_buffer (void* data, struct wpe_fdo_shm_exported_buffer* exported_b
...
 
     buffer->exported_buffer = exported_buffer;
-    shm_buffer_copy_contents (buffer, exported_shm_buffer);
+    shm_buffer_copy_contents(buffer, exported_shm_buffer);
 
-    wl_surface_attach (win_data.wl_surface, buffer->buffer, 0, 0);
-    wl_surface_damage(win_data.wl_surface, 0, 0, INT32_MAX, INT32_MAX);
-    request_frame ();
-    wl_surface_commit (win_data.wl_surface);
+    wl_surface_attach(win_data.shell_context->wl_shell_data.wl_surface, buffer->buffer, 0, 0);
```
Cwiiis, line 2031: These should be win_data.wl_surface, right? This wouldn't build if HAVE_SHM_EXPORTED_BUFFER is defined.

platform/wayland/cog-platform-wl.c
```diff
@@ -757,14 +1014,17 @@ registry_global (void               *data,
...
     } else if (strcmp(interface, wl_shm_interface.name) == 0) {
         wl_data.shm = wl_registry_bind(registry, name, &wl_shm_interface, version);
     } else if (strcmp(interface, xdg_wm_base_interface.name) == 0) {
         wl_data.xdg_shell = wl_registry_bind(registry, name, &xdg_wm_base_interface, version);
         g_assert(wl_data.xdg_shell);
         xdg_wm_base_add_listener(wl_data.xdg_shell, &xdg_shell_listener, NULL);
+        win_data.shell_context = (shell_context *) (&xdg_data);
     } else if (strcmp(interface, zwp_fullscreen_shell_v1_interface.name) == 0) {
         wl_data.fshell = wl_registry_bind(registry, name, &zwp_fullscreen_shell_v1_interface, version);
+        win_data.shell_context = (shell_context *) (&f_data);
```
Cwiiis, line 1027: These look good, but we're still missing a default setting for win_data.shell_context - if none of these clauses that set it are hit, the first call will likely result in a crash.

platform/wayland/cog-platform-wl.c
```diff
@@ -2236,124 +2398,75 @@ create_popup (WebKitOptionMenu *option_menu)
...
+    webkit_option_menu_close(popup_data.option_menu);
+    g_clear_pointer(&popup_data.popup_menu, cog_popup_menu_destroy);
+    g_clear_object(&popup_data.option_menu);
 
-    g_clear_pointer (&popup_data.xdg_popup, xdg_popup_destroy);
-    g_clear_pointer (&popup_data.xdg_surface, xdg_surface_destroy);
-    g_clear_pointer (&popup_data.xdg_positioner, xdg_positioner_destroy);
-    g_clear_pointer (&popup_data.shell_surface, wl_shell_surface_destroy);
-    g_clear_pointer (&popup_data.wl_surface, wl_surface_destroy);
+    win_data.shell_context->functions.surface_destroy();
```
Cwiiis, line 2422: This is a little confusing to me - perhaps this function should be called popup_surface_destroy to be clearer?

Cwiiis (changes requested):
Definitely looking better, but lots of small things to address. I wouldn't mind seeing another commit that addresses these smaller details, but feel free to go ahead with splitting this out into separate file(s) if it makes sense for you.

aperezdc:
I have taken a brief look at this, and it looks like there is not much left to do before it can be merged—thanks everybody involved!

Nevertheless, I would like to make today the 0.12 stable branch, so I am going to defer to tag it for 0.14 😸

aperezdc:
As agreed with @Nasah-Kuma, I have done a rebase and squashed some of the commits together, and added one more commit with a small cleanup—I will push it momentarily.

While testing, I have noticed that with the changes applied Cog segfaults when running under wlroots-based compositors (which used to work fine). The traceback is as follows (log edited to show the interesting bits):

```
% WAYLAND_DISPLAY=wayland-1 G_MESSAGES_DEBUG=Cog,Cog-Core,Cog-FDO COG_MODULEDIR=b/modules \
    gdb --args b/cog -P wl ddg.gg
[...]
(cog:755159): Cog-DEBUG: 20:11:18.071: platform_setup: Platform name: wl
[...]
(cog:755159): Cog-FDO-DEBUG: 20:11:18.083: Using 'wl_compositor' interface obtained from the Wayland registry.
(cog:755159): Cog-FDO-DEBUG: 20:11:18.083: Using 'wl_subcompositor' interface obtained from the Wayland registry.
[...]
(cog:755159): Cog-FDO-DEBUG: 20:11:18.083: Using 'xdg_wm_base' interface obtained from the Wayland registry.
[...]
(cog:755159): Cog-FDO-DEBUG: 20:11:18.391: New XDG toplevel configuration: (0, 0)
(cog:755159): Cog-FDO-DEBUG: 20:11:18.391: Resized EGL buffer to: (1024, 768) @1x


Thread 1 "cog" received signal SIGSEGV, Segmentation fault.
0x00007ffff7f96928 in cog_popup_menu_get_buffer (popup_menu=0x0) at /home/aperez/devel/wpe/cog-master/platform/wayland/cog-popup-menu-wl.c:318
318         if (popup_menu->pending_changes) {
(gdb) bt
#0  0x00007ffff7f96928 in cog_popup_menu_get_buffer (popup_menu=0x0) at /home/aperez/devel/wpe/cog-master/platform/wayland/cog-popup-menu-wl.c:318
#1  0x00007ffff7f93a32 in display_popup () at /home/aperez/devel/wpe/cog-master/platform/wayland/cog-platform-wl.c:2515
#2  0x00007ffff0f8cd4a in  () at /usr/lib/libffi.so.8
#3  0x00007ffff0f8c267 in  () at /usr/lib/libffi.so.8
#4  0x00007fffeec6f323 in  () at /usr/lib/libwayland-client.so.0
#5  0x00007fffeec6b5ab in  () at /usr/lib/libwayland-client.so.0
#6  0x00007fffeec6cf4c in wl_display_dispatch_queue_pending () at /usr/lib/libwayland-client.so.0
#7  0x00007ffff7f92531 in wl_src_dispatch (base=0x5555556b50c0, callback=<optimized out>, user_data=<optimized out>)
    at /home/aperez/devel/wpe/cog-master/platform/wayland/cog-platform-wl.c:704
#8  0x00007ffff2c34fd3 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#9  0x00007ffff2c8b049 in  () at /usr/lib/libglib-2.0.so.0
#10 0x00007ffff2c32545 in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#11 0x00007ffff2e526ee in g_application_run () at /usr/lib/libgio-2.0.so.0
#12 0x0000555555559b9f in main (argc=3, argv=0x7fffffffdb68) at /home/aperez/devel/wpe/cog-master/launcher/cog.c:41
(gdb)
```

I have not yet investigated why this happens, but will try to debug it in the next days.

Running under Weston works as expected, so I suppose that was the compositor used for testing 😄

aperezdc:
Ah, I just realized that of course I cannot push to @Nasah-Kuma's branch in their fork, so I have posted the rebase to the [aperezdc/pr359-rebased](https://github.com/Igalia/cog/tree/aperezdc/pr359-rebased) branch for now.

PR #361, opened 2021-09-23
Add gamepad implementation.

ceyusa:
The current implementation is based on libmanette. If libmanette is available a noop implementation is used.

Depends on https://github.com/WebPlatformForEmbedded/libwpe/pull/88

### DIFF
--- a/core/cog-gamepad-manette.c
+++ b/core/cog-gamepad-manette.c
@@ -1,0 +1,321 @@
+/*
+ * cog-gamepad-manette.c
+ * Copyright (C) 2022 Igalia S.L.
+ *
+ * Distributed under terms of the MIT license.
+ */
+
+#include "cog.h"
+
+#include <libmanette.h>
+#include <linux/input-event-codes.h>
+#include <wpe/wpe.h>
+
+typedef struct _Provider Provider;
+typedef struct _Gamepad  Gamepad;
+
+struct _Provider {
+    struct wpe_gamepad_provider *wk_provider;
+    ManetteMonitor              *monitor;
+};
+
+struct _Gamepad {
+    struct wpe_gamepad *wk_gamepad;
+    ManetteDevice      *device;
+};
+
+static void *
+provider_create(struct wpe_gamepad_provider *wk_provider)
+{
+    Provider *provider = g_new0(Provider, 1);
+
+    provider->wk_provider = wk_provider;
+    return provider;
+}
+
+static void
+provider_destroy(void *data)
+{
+    Provider *provider = data;
+
+    if (!provider)
+        return;
+
+    g_clear_object(&provider->monitor);
+    g_free(provider);
+}
+
+static void
+add_device(Provider *provider, ManetteDevice *device)
+{
+    g_debug("gamepad: %s - connected", manette_device_get_name(device));
+
+    uintptr_t gamepad_id = (uintptr_t) device;
+    wpe_gamepad_provider_dispatch_gamepad_connected(provider->wk_provider, gamepad_id);
+}
+
+static gboolean
+add_connected_devices(void *data)
+{
+    Provider                     *provider = data;
+    g_autoptr(ManetteMonitorIter) iter = manette_monitor_iterate(provider->monitor);
+    ManetteDevice                *device;
+
+    while (manette_monitor_iter_next(iter, &device))
+        add_device(provider, device);
+
+    return G_SOURCE_REMOVE;
+}
+
+static void
+on_device_connected(ManetteMonitor *monitor, ManetteDevice *device, void *data)
+{
+    add_device(data, device);
+}
+
+static void
+on_device_disconnected(ManetteMonitor *monitor, ManetteDevice *device, void *data)
+{
+    Provider *provider = data;
+
+    g_debug("gamepad: %s - disconnected", manette_device_get_name(device));
+
+    uintptr_t gamepad_id = (uintptr_t) device;
+    wpe_gamepad_provider_dispatch_gamepad_disconnected(provider->wk_provider, gamepad_id);
+}
+
+static void
+provider_start(void *data)
+{
+    Provider *provider = data;
+
+    /* double call to start? */
+    if (provider->monitor)
+        return;
+
+    g_debug("gamepad: starting monitor");
+
+    provider->monitor = manette_monitor_new();
+    g_signal_connect(provider->monitor, "device-connected", G_CALLBACK(on_device_connected), provider);
+    g_signal_connect(provider->monitor, "device-disconnected", G_CALLBACK(on_device_disconnected), provider);
+
+    g_idle_add(add_connected_devices, provider);
+}
+
+static void
+provider_stop(void *data)
+{
+    Provider *provider = data;
+
+    if (!provider)
+        return;
+
+    g_debug("gamepad: stoping monitor");
+
+    g_signal_handlers_disconnect_by_data(provider->monitor, provider);
+}
+
+static bool
+to_standard_gamepad_axis(uint16_t axis, enum wpe_gamepad_axis *result)
+{
+    switch (axis) {
+    case ABS_X:
+        *result = WPE_GAMEPAD_AXIS_LEFT_STICK_X;
+        break;
+    case ABS_Y:
+        *result = WPE_GAMEPAD_AXIS_LEFT_STICK_Y;
+        break;
+    case ABS_RX:
+        *result = WPE_GAMEPAD_AXIS_RIGHT_STICK_X;
+        break;
+    case ABS_RY:
+        *result = WPE_GAMEPAD_AXIS_RIGHT_STICK_Y;
+        break;
+    default:
+        return false;
+    }
+
+    return true;
+}
+
+static void
+on_absolute_axis(ManetteDevice *device, ManetteEvent *event, Gamepad *gamepad)
+{
+    uint16_t              axis;
+    double                value;
+    enum wpe_gamepad_axis std_axis;
+
+    if (!manette_event_get_absolute(event, &axis, &value))
+        return;
+
+    if (to_standard_gamepad_axis(axis, &std_axis))
+        wpe_gamepad_dispatch_axis_changed(gamepad->wk_gamepad, std_axis, value);
+}
+
+static bool
+to_standard_gamepad_button(uint16_t manetteButton, enum wpe_gamepad_button *result)
+{
+    switch (manetteButton) {
+    case BTN_SOUTH:
+        *result = WPE_GAMEPAD_BUTTON_BOTTOM;
+        break;
+    case BTN_EAST:
+        *result = WPE_GAMEPAD_BUTTON_RIGHT;
+        break;
+    case BTN_NORTH:
+        *result = WPE_GAMEPAD_BUTTON_TOP;
+        break;
+    case BTN_WEST:
+        *result = WPE_GAMEPAD_BUTTON_LEFT;
+        break;
+    case BTN_TL:
+        *result = WPE_GAMEPAD_BUTTON_LEFT_SHOULDER;
+        break;
+    case BTN_TR:
+        *result = WPE_GAMEPAD_BUTTON_RIGHT_SHOULDER;
+        break;
+    case BTN_TL2:
+        *result = WPE_GAMEPAD_BUTTON_LEFT_TRIGGER;
+        break;
+    case BTN_TR2:
+        *result = WPE_GAMEPAD_BUTTON_RIGHT_TRIGGER;
+        break;
+    case BTN_SELECT:
+        *result = WPE_GAMEPAD_BUTTON_SELECT;
+        break;
+    case BTN_START:
+        *result = WPE_GAMEPAD_BUTTON_START;
+        break;
+    case BTN_THUMBL:
+        *result = WPE_GAMEPAD_BUTTON_LEFT_STICK;
+        break;
+    case BTN_THUMBR:
+        *result = WPE_GAMEPAD_BUTTON_RIGHT_STICK;
+        break;
+    case BTN_DPAD_UP:
+        *result = WPE_GAMEPAD_BUTTON_D_PAD_TOP;
+        break;
+    case BTN_DPAD_DOWN:
+        *result = WPE_GAMEPAD_BUTTON_D_PAD_BOTTOM;
+        break;
+    case BTN_DPAD_LEFT:
+        *result = WPE_GAMEPAD_BUTTON_D_PAD_LEFT;
+        break;
+    case BTN_DPAD_RIGHT:
+        *result = WPE_GAMEPAD_BUTTON_D_PAD_RIGHT;
+        break;
+    case BTN_MODE:
+        *result = WPE_GAMEPAD_BUTTON_CENTER;
+        break;
+    default:
+        return false;
+    }
+
+    return true;
+}
+
+static inline void
+dispatch_button_event(Gamepad *gamepad, ManetteEvent *event, bool pressed)
+{
+    uint16_t                button;
+    enum wpe_gamepad_button std_button;
+
+    if (!manette_event_get_button(event, &button))
+        return;
+
+    if (to_standard_gamepad_button(button, &std_button))
+        wpe_gamepad_dispatch_button_changed(gamepad->wk_gamepad, std_button, pressed);
+}
+
+static void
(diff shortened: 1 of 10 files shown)


ceyusa:
rebased and implemented method to get the view backend

ceyusa:
sync'ed with latest review of libwpe

ceyusa:
Downgraded libmanette version to 0.2.4 because is the one used in WebKit SDK, sadly the CI can install  0.2.3, which I cannot test locally easily.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+    default:
+        return -1;
+    }
+}
+
+static void
+on_button_press(ManetteDevice *device, ManetteEvent *event, Gamepad *gamepad)
+{
+    uint16_t button;
+    int      std_button;
```
aperezdc, line 212: Let's use `enum wpe_gamepad_button` as the type here.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+}
+
+static void
+on_button_release(ManetteDevice *device, ManetteEvent *event, Gamepad *gamepad)
+{
+    uint16_t button;
+    if (!manette_event_get_button(event, &button))
+        return;
+
+    wpe_gamepad_dispatch_button_changed(gamepad->server, toStandardGamepadButton(button), false);
```
aperezdc, line 229: This is missing to check whether the button can be mapped to a `enum wpe_gamepad_button`, it should use the same logic as above.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+
+typedef struct _Provider Provider;
+typedef struct _Gamepad  Gamepad;
+
+struct _Provider {
+    struct wpe_gamepad_provider *server;
+    ManetteMonitor              *monitor;
+
+    gulong connected;
+    gulong disconnected;
```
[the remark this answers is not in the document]
aperezdc: Actually, we could even use `g_signal_handlers_disconnect_by_data(provider->monitor, provider)` and skip saving the handler identifiers in the `Provider` 🤔

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+};
+
+struct _Gamepad {
+    struct wpe_gamepad *server;
```
aperezdc, line 26: The name “server” here is odd. There is no server, because the code uses libmanette in-process. This could be just `struct wpe_gamepad *gamepad`.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+}
+
+static gpointer
+gamepad_create(struct wpe_gamepad *server, unsigned gamepad_id)
+{
+    Gamepad       *gamepad;
+    ManetteDevice *device;
+
+    device = GUINT_TO_POINTER(gamepad_id);
+    if (!MANETTE_IS_DEVICE(device))
```
aperezdc, line 266: This should never happen, because we control how identifiers are generated and they are always valid pointers. If you feel like adding some safeguard here, use `g_assert(MANETTE_IS_DEVICE(device))` instead, right?

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+
+    return gamepad;
+}
+
+static void
+gamepad_destroy(gpointer data)
+{
+    Gamepad *gamepad = data;
+
+    if (!(gamepad && gamepad->device))
```
aperezdc, line 289: It looks to me that `gamepad_destroy()` is always called with a valid pointer and that there are unneeded checks here. Remember that `g_clear_object()` will only unref if the pointer is non-null so the check on `gamepad->device` is definitely unneeded.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+    .create = gamepad_create,
+    .destroy = gamepad_destroy,
+    .get_id = gamepad_get_id,
+};
+
+void
+cog_register_gamepad_backend(GamepadProviderGetViewBackend *provider_get_view_for_gamepad)
+{
+    provider_interface.get_view_backend = provider_get_view_for_gamepad;
+    wpe_gamepad_set_handler(&provider_interface, &device_interface);
```
aperezdc, line 324: I'm not a fan of this, because it means that there can only ever be one gamepad backend, and it has to be the Manette one. What if we wanted to build more than one and allow choosing at runtime? I think we can leave it as-is for now, but knowing that this may need changes later on.

aperezdc (changes requested):
First things first, let me say that I think the overall structure of the code looks fine and worked fine for me here. Most of the comments are mostly style related, but please do read through them. I think we could merge an updated version of this PR once those have been addressed.

Thanks for the work on this, and please do let me know if anything is unclear! 🎮

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+    Provider *provider = data;
+
+    g_clear_object(&provider->monitor);
+    g_free(provider);
+}
+
+static void
+add_device(Provider *provider, ManetteDevice *device)
+{
+    g_message("gamepad: %s - connected", manette_device_get_name(device));
```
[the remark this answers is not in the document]
ceyusa: I have turn all messages into debugs

core/cog-gamepad-manette.c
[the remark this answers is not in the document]
ceyusa: done

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+}
+
+static gboolean
+add_connected_devices(gpointer data)
+{
+    Provider                     *provider = data;
+    g_autoptr(ManetteMonitorIter) iter;
+    ManetteDevice                *device;
+
+    iter = manette_monitor_iterate(provider->monitor);
```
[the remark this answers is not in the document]
ceyusa: done

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+provider_stop(void *data)
+{
+    Provider *provider = data;
+
+    if (!provider)
+        return;
+
+    g_message("gamepad: stoping monitor");
+
+    if (provider->connected > 0) {
```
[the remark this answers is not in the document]
ceyusa: using `g_signal_handlers_disconnect_by_data`

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+    if (!manette_event_get_absolute(event, &axis, &value))
+        return;
+
+    std_axis = toStandardGamepadAxis(axis);
+    if (std_axis != -1)
+        wpe_gamepad_dispatch_axis_changed(gamepad->server, (enum wpe_gamepad_axis) std_axis, value);
+}
+
+static int
+toStandardGamepadButton(uint16_t manetteButton)
```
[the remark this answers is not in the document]
ceyusa: done.

Just I break the lines. It looks more readable fmpov.

core/cog-gamepad-manette.c
ceyusa: done

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+{
+    uint16_t button;
+    int      std_button;
+
+    if (!manette_event_get_button(event, &button))
+        return;
+
+    std_button = toStandardGamepadButton(button);
+    if (std_button != -1)
+        wpe_gamepad_dispatch_button_changed(gamepad->server, (enum wpe_gamepad_button) std_button, true);
```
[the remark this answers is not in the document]
ceyusa: done... it's nicer, indeed

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,325 @@
...
+    wpe_gamepad_dispatch_button_changed(gamepad->server, toStandardGamepadButton(button), false);
+}
+
+static void
+device_disconnect_cb(ManetteDevice *device, Gamepad *gamepad)
+{
+    if (!gamepad)
+        return;
+
+    if (gamepad->button_pressed > 0) {
```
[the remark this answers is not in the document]
ceyusa: `g_signal_handlers_disconnect_by_data`

core/cog-gamepad-manette.c
ceyusa: I renamed `gamepad` and `provider` from WebKit as `wk_gamepad` and `wk_provider`. Otherwise there're symbols collisions.

core/cog-gamepad-manette.c
ceyusa: done

core/cog-gamepad-manette.c
ceyusa: It's a nice feature to have, but perhaps, as you mention, not a priority.

core/cog-gamepad-manette.c
ceyusa: done as you suggested.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,319 @@
...
+static gboolean
+add_connected_devices(gpointer data)
+{
+    Provider                     *provider = data;
+    g_autoptr(ManetteMonitorIter) iter = manette_monitor_iterate(provider->monitor);
+    ManetteDevice                *device;
+
+    while (manette_monitor_iter_next(iter, &device)) {
+        add_device(provider, device);
+    };
```
aperezdc, line 65: Please remove this stray semicolon before merging. The braces could be removed because there is only one statement, but I have no strong preference in this regard.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,319 @@
...
+    Provider *provider = data;
+    int       n;
+
+    if (!provider)
+        return;
+
+    g_debug("gamepad: stoping monitor");
+
+    n = g_signal_handlers_disconnect_by_data(provider->monitor, provider);
+    g_assert(n == 2);
```
aperezdc, line 116: I would remove this assertion and trust GLib to be able to do the right thing 😉 

Also this is going to cause warnings when building with assertions disabled, due to `n` being assigned but never read back again.

aperezdc (changes requested):
This is almost there! Sorry that I am asking for another couple of changes again 😇. This is _almost_ ready to go!

manuelafm:
Hi @aperezdc and @ceyusa ,

I happened to stumble upon this PR and think that the issues have been addressed or mostly, but it's still pending, so decided to nudge because it looks as if this crept through a crevice but it's basically ready to merge.  Sorry if I'm adding noise.

ceyusa:
Changes:

* Meson support
* Support aarch64, where the memory addresses are bigger than the unsigned int store size.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,343 @@
...
+{
+    Gamepad *gamepad = data;
+
+    if (!(gamepad && gamepad->device))
+        return "Unknown";
+
+    return manette_device_get_name(gamepad->device);
+}
+
+static struct wpe_gamepad_provider_interface provider_interface = {
```
aperezdc, line 317: Use `static const` here:

```suggestion
static const struct wpe_gamepad_provider_interface provider_interface = {
```

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,343 @@
...
+    .create = provider_create,
+    .destroy = provider_destroy,
+    .start = provider_start,
+    .stop = provider_stop,
+};
+
+static struct wpe_gamepad_interface device_interface = {
```
aperezdc, line 324: Same here, this can be `static const`:

```suggestion
static const struct wpe_gamepad_interface device_interface = {
```

aperezdc (changes requested):
I'd like a few things tweaked before merging, please take a look at the comments.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,343 @@
...
+static void
+add_device(Provider *provider, ManetteDevice *device)
+{
+    unsigned gamepad_id;
+
+    g_debug("gamepad: %s - connected", manette_device_get_name(device));
+
+    /* XXX(victor): this hack is for certain architectures, such as aarch64,
+     * where the memory address is bigger than unsigned int */
+    gamepad_id = GPOINTER_TO_UINT(device);
```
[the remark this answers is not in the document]
ceyusa: that would break the libwpe API, since the gamepad identifiers  are set as uints. I think that a data type which size is arch dependent would be error prone.

core/cog-gamepad-manette.c
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * cog_register_gamepad_backend:
+ * @provider_get_view_for_gamepad: A function that * returns the view backend
+ *    associated with the gamepad.
+ *
+ * Installs a gamepad backend, which is in charge of the physical device
+ * events.
+ */
+void
+cog_register_gamepad_backend(GamepadProviderGetViewBackend *provider_get_view_for_gamepad)
```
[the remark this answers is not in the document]
ceyusa: assuming that users will have only one gamepad implementation it will choose at configuration/compilation time, thus only one `cog_register_gamepad_backend` will be resolved by the linker. Right now there are two implementations: noop and manette.

core/cog-gamepad-manette.c
ceyusa: this cannot be const since it's configured by each platform with `.get_view_backed`.

core/cog-gamepad-manette.c
ceyusa: this one might, but I'm not sure because of aesthetics symmetry.

core/cog-gamepad-manette.c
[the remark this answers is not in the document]
aperezdc: We already have that issue, becaus `unsigned int` is architecture-dependant. Maybe we should switch here to `uint64_t`. Anyway, it does not matter because we don't need things like a 32-bit build of libwpe talking to a 64-bit one of WebKit or Cog. At any rate, for the sake of convenience, we should choose a type that can fit a pointer in most (if not all) architectures, and if we don't want to use `uintptr_t`, then `size_t` is the next best choice. WDYT?

core/cog-gamepad-manette.c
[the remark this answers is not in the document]
aperezdc: I don't see why we wouldn't allow chosing an implementation at runtime, the same way we allow choosing Wayland/X11/etc. at runtime. It's not like it would be much more work anyway 🤔

core/cog-gamepad-manette.c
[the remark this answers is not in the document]
ceyusa: fixed. It needs libwpe api change in https://github.com/WebPlatformForEmbedded/libwpe/pull/118

core/cog-gamepad-manette.c
ceyusa: all the callback structures are now const. It needs https://github.com/WebPlatformForEmbedded/libwpe/pull/118

core/cog-gamepad-manette.c
ceyusa: ditto

core/cog-gamepad-manette.c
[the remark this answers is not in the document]
ceyusa: I would leave this change for a next iteration, since it's local to cog.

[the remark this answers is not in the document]
aperezdc: That would be fine 👌🏼

core/meson.build
```diff
@@ -47,6 +48,13 @@ cogcore_dependencies = [
 
 install_headers(cogcore_headers, cogcore_config_h, subdir: 'cog')
 
+if (wpe_dep.version().version_compare('>= 1.13.3') and manette_dep.found())
```
aperezdc, line 51: Probably we want to check for version `>=1.13.90` here, to be sure the latest modifications to the API are available.

aperezdc (changes requested):
This is almost ready to merge, I just made a couple of very small change requests. Thanks!

core/meson.build
ceyusa: done

aperezdc (approved):
r=me, let's merge it 👍🏼

PR #363, opened 2021-09-24
[RFC] Use object properties to configure platform implementations

aperezdc:
This is an initial RFC to steer towards improving how platform modules are configured. The main idea is to turn options into properties of the `CogPlatform` subclass implemented by the platform module.

The information about object properties/settings can then be inspected e.g. at run time and used in different ways:

- Programmatically read property values from the configuration file (implemented in this PR as an usage example).
- Allowing passing options at creation time, e.g. adding a new `cog_platform_new_with_options()` function.
- Reading options from the command line based on the available properties.
- Using their documentation comments to generate pages of the reference manual.

…and probably others.

### DIFF
--- a/core/cog-utils.c
+++ b/core/cog-utils.c
@@ -382,3 +382,110 @@ cog_option_entries_from_class (GObjectClass *klass)
 
     return g_steal_pointer (&entries);
 }
+
+/**
+ * cog_apply_properties_from_key_file:
+ * @object: An object to set properties on
+ * @key_file: A key file
+ * @group_name: Name of a key file group
+ * @error: Where to store an error, if any
+ *
+ * Set properties on an object from values stored in a #GKeyFile.
+ *
+ * For each writable property of the object's class, looks up key with
+ * the same name as the property in the specified group of the key file.
+ * If present, it will be set as the new value of the property.
+ *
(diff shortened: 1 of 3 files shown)


tobias-grasse:
From a purely user perspective, this sounds very desirable (e.g. for our WPE snap package).

core/cog-utils.c
```diff
@@ -383,3 +383,156 @@ cog_option_entries_from_class (GObjectClass *klass)
...
+        g_auto(GStrv) kv = g_strsplit(params[i], "=", 2);
+        if (g_strv_length(kv) != 2) {
+            g_warning("%s: Invalid parameter syntax '%s'.", __func__, params[i]);
+            continue;
+        }
+
+        const char *k = g_strstrip(kv[0]);
+        const char *v = g_strstrip(kv[1]);
+
+        g_key_file_set_value(key_file, group_name, k, v);
```
TingPing, line 416: `params_string` can contain invalid characters, `g_key_file_set_string()` escapes them, `g_key_file_set_value()` does not.

You use `g_key_file_get_string()` below which unescapes.

TingPing: I guess they key also needs to be validated, at least for newlines.

PR #360, opened 2021-09-17
drm: Implement output rotation

aperezdc:
Does what it says on the tin. There are a few things which are needed:

- [x] Fetching the values for the `rotate-<N>` DRM plane properties.
- [x] Configuring the `rotation` property when mode setting.
- [ ] Transforming input coordinates, [libinput_device_config_rotation_set_angle()](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga136b2f7753235c5d2a9fb2762a040eef) should be enough.
- [x] Exposing a setting somewhere, to allow configuring at run time.
- [x] Exposing an option in the `cog` launcher.

Closes #205 

----

Currently this is posted as a draft for early feedback, as this needs still some work before I am completely satisfied with it 😄

PR #371, opened 2021-11-06
drm: Implement an OpenGL ES renderer with rotation support

aperezdc:
This is an alternative take on #360 — as it turns out using DRM/KMS exclusively is a can of works:

- In many cases graphics hardware imposes stringent restrictions on how to use hardware-assisted rotation support for DRM/KMS planes. For example on Intel GPUs typically one needs framebuffers with the `I915_FORMAT_MOD_Y_TILED` or `I915_FORMAT_MOD_Yf_TILED` modifiers for 90º/270º rotations, and on OMAP hardware one needs to manually use `ioctl()` and set one of the `OMAP_BO_TILED_*` flags in BOs which will be used for rotated scanout. Ad-hoc support for each hardware would be needed in `cog-platform-drm.c`.
- In some cases using one more hardware block results in more power consumption than using OpenGL because the GPU would be active and running anyway, and if unused the additional hardware needed to do the rotation (i.e the “tiler” block in OMAP devices) can stay powered off.
- Some hardware plainly does not provide a way to rotate the output on scanout.

For example  Jonas Ådahl mentioned to me in IRC that GNOME Shell [blacklists some configurations](https://gitlab.gnome.org/GNOME/mutter/-/blob/69fcb3a1464716ea8a61e311fa60a0cb6a358fe3/src/backends/native/meta-kms-plane.c#L134) for some of the reasons above.

In order to avoid hitting the issues and complexities above, this patch set:

- Introduces the concept of “renderers“, which implement the mechanism that actually puts the content on the screen.
- Splits a good chunk of the existing code into a new “modeset” renderer (c.f. `cog-drm-modeset-renderer.c`)
- Adds a new “gles“ renderer (c.f. `cog-drm-gles-renderer.c`).
- Adds support for 90º/180º/270º rotations into the “gles“ renderer.
- Adds the needed calls for `libinput` to report events rotated for devices that need it (basically those which are ”touchable”, so not only touch screens, but also tablets and touchpads).
- Allows changing settings from the config file (renderer), on `CogDrmPlatform` (renderer as construct-only, rotation at any time), and passing them as parameters to `cog_platform_setup()` (both renderer and rotation e.g. with a string of comma-separated items `"renderer=gles,rotation=1"`).

There are a few things which are intentionally left out of this PR to avoid making it too big: further refactoring the code from both renderers (there is some duplication), splitting the DRM/KMS handling code into helper functions in their own source file, painting the cursor in the “gles“ renderer when KMS cursor planes are not available, etc. — those I would rather do as follow-ups later on.

### DIFF
--- a/platform/common/cog-gl-utils.c
+++ b/platform/common/cog-gl-utils.c
@@ -1,0 +1,80 @@
+/*
+ * cog-gl-utils.c
+ * Copyright (C) 2021 Igalia S.L.
+ *
+ * Distributed under terms of the MIT license.
+ */
+
+#include "cog-gl-utils.h"
+
+#include "../../core/cog.h"
+
+void
+cog_gl_shader_id_destroy(CogGLShaderId *shader_id)
+{
+    if (shader_id && *shader_id) {
+        glDeleteShader(*shader_id);
+        *shader_id = 0;
+    }
+}
+
+CogGLShaderId
+cog_gl_shader_id_steal(CogGLShaderId *shader_id)
+{
+    g_assert(shader_id);
+    CogGLShaderId result = *shader_id;
+    *shader_id = 0;
+    return result;
+}
+
+CogGLShaderId
+cog_gl_load_shader(const char *source, GLenum kind, GError **error)
+{
+    g_assert(source != NULL);
+    g_assert(kind == GL_VERTEX_SHADER || kind == GL_FRAGMENT_SHADER);
+
+    g_auto(CogGLShaderId) shader = glCreateShader(kind);
+    glShaderSource(shader, 1, &source, NULL);
+
+    GLenum err;
+    if ((err = glGetError()) != GL_NO_ERROR) {
+        g_set_error_literal(error, COG_PLATFORM_EGL_ERROR, err, "Cannot set shader source");
+        return 0;
+    }
+
+    glCompileShader(shader);
+    if ((err = glGetError()) != GL_NO_ERROR) {
+        g_set_error_literal(error, COG_PLATFORM_EGL_ERROR, err, "Cannot compile shader");
+        return 0;
+    }
+
+    GLint shaderCompiled = GL_FALSE;
+    glGetShaderiv(shader, GL_COMPILE_STATUS, &shaderCompiled);
+    if (shaderCompiled == GL_TRUE)
+        return cog_gl_shader_id_steal(&shader);
+
+    GLint log_length = 0;
+    glGetShaderiv(shader, GL_INFO_LOG_LENGTH, &log_length);
+    g_autofree char *log = g_new0(char, log_length + 1);
+    glGetShaderInfoLog(shader, log_length, NULL, log);
+    g_set_error(error, COG_PLATFORM_EGL_ERROR, 0, "Shader compilation: %s", log);
+    return 0;
+}
+
+bool
+cog_gl_link_program(GLuint program, GError **error)
+{
+    glLinkProgram(program);
+
+    GLint status = 0;
+    glGetProgramiv(program, GL_LINK_STATUS, &status);
+    if (status)
+        return true;
+
+    GLint log_length = 0;
+    glGetProgramiv(program, GL_INFO_LOG_LENGTH, &log_length);
+    g_autofree char *log = g_new0(char, log_length + 1);
+    glGetProgramInfoLog(program, log_length, NULL, log);
+    g_set_error(error, COG_PLATFORM_EGL_ERROR, 0, "Shader linking: %s", log);
+    return false;
+}
--- a/platform/common/cog-gl-utils.h
+++ b/platform/common/cog-gl-utils.h
@@ -1,0 +1,26 @@
+/*
+ * cog-gl-utils.h
+ * Copyright (C) 2021 Igalia S.L.
+ *
+ * Distributed under terms of the MIT license.
+ */
+
+#pragma once
+
+#include <epoxy/gl.h>
+#include <glib.h>
+
+G_BEGIN_DECLS
+
+typedef GLuint CogGLShaderId;
+
+void cog_gl_shader_id_destroy(CogGLShaderId *shader_id);
+G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC(CogGLShaderId, cog_gl_shader_id_destroy)
+
+CogGLShaderId cog_gl_shader_id_steal(CogGLShaderId *shader_id) G_GNUC_WARN_UNUSED_RESULT;
+
+CogGLShaderId cog_gl_load_shader(const char *source, GLenum kind, GError **) G_GNUC_WARN_UNUSED_RESULT;
+
+bool cog_gl_link_program(GLuint program, GError **);
+
+G_END_DECLS
--- a/platform/drm/cog-drm-gles-renderer.c
+++ b/platform/drm/cog-drm-gles-renderer.c
@@ -5,16 +5,77 @@
  * Distributed under terms of the MIT license.
  */
 
+#include "../../core/cog.h"
+#include "../common/cog-gl-utils.h"
 #include "cog-drm-renderer.h"
+#include <drm_fourcc.h>
+#include <drm_mode.h>
 #include <epoxy/gl.h>
+#include <errno.h>
+#include <gbm.h>
+#include <glib-unix.h>
 #include <wayland-util.h>
 #include <wpe/fdo-egl.h>
 #include <wpe/fdo.h>
+#include <xf86drm.h>
+#include <xf86drmMode.h>
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC(drmModePlane, drmModeFreePlane)
 
 typedef struct {
     CogDrmRenderer base;
 
+    struct gbm_device  *gbm_device;
+    struct gbm_surface *gbm_surface;
+    struct gbm_bo      *current_bo;
+    struct gbm_bo      *next_bo;
+    uint32_t            gbm_format;
+
+    /*
+     * Logical view size without transformations applied, which is needed to
+     * change the transformed size (i.e. rotated) of the view after the view
+     * backend has been instantiated. Note that physical output size is
+     * always determinted from the chosen DRM/KMS mode.
+     */
+    uint32_t width, height;
+
+    CogDrmRendererRotation rotation;
+
+    EGLDisplay egl_display;
+    EGLConfig  egl_config;
+    EGLContext egl_context;
+    EGLSurface egl_surface;
+
+    /*
+     * The renderer wraps frames provided by wpebackend-fdo into a texture,
+     * then paints a quad which samples said texture. A simple GLSL shader
+     * program uses the "position" attribute to fetch the coordinates of
+     * the quad, and the "texture" attribute the UV mapping coordinates for
+     * the texture. Rotation is achieved by changing the UV mapping. The
+     * "texture" uniform is used to reference the texture unit where the
+     * frame textures get loaded.
+     */
+    GLuint gl_program;
+    GLuint gl_texture;
+    GLint  gl_attrib_position;
+    GLint  gl_attrib_texture;
+    GLint  gl_uniform_texture;
+
     struct wpe_view_backend_exportable_fdo *exportable;
+
+    drmEventContext drm_context;
+    unsigned        drm_fd_source;
+    uint32_t        crtc_id;
+    uint32_t        connector_id;
+    uint32_t        plane_id;
+    drmModeModeInfo mode;
+    bool            mode_set;
+    bool            atomic_modesetting;
+
+    struct {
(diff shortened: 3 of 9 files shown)


aperezdc:
Another thought: The “x11” and “gtk4” platform plug-ins have in them code that is eerily similar to that in `cog-drm-gles-renderer.c`, and we may even be able to make all of them use the same code at some point. Wouldn't that be neat? 😄

aperezdc:
This needed a rebase after #376 got merged, now it's again good to go.

aperezdc:
Ping reviewers 🏓

crzwdjk:
The approach to GL rendering looks correct to me. I am not sure that the name "renderer" is the best term here, but I can't think of anything better. I don't know what the cog review policy is like, but I can go through the commits in more detail to give a more complete review.

platform/drm/cog-drm-renderer.c
```diff
@@ -0,0 +1,19 @@
...
+#include "cog-drm-renderer.h"
+
+void
+cog_drm_renderer_destroy(CogDrmRenderer *self)
+{
+    if (self) {
+        if (self->destroy)
+            self->destroy(self);
+        else
+            g_free(self);
```
Cwiiis, line 17: While it may make sense if I knew the codebase better, this looks like weird behaviour to me... asserting self->destroy and calling it feels like the more natural thing to me, surely every renderer should have a destructor?

platform/drm/cog-platform-drm.c
```diff
@@ -214,119 +204,83 @@ static struct {
...
-        gdouble value = g_key_file_get_double (key_file,
-                                               "drm", "device-scale-factor",
-                                               &lookup_error);
-        if (!lookup_error) {
-            drm_data.device_scale = value;
-            g_debug ("init_config:overriding device_scale value, using %.2f from config",
-                     drm_data.device_scale);
+        {
+            g_autoptr(GError) lookup_error = NULL;
+            gdouble           value = g_key_file_get_double(key_file, "drm", "device-scale-factor", &lookup_error);
```
Cwiiis, line 229: This seems like weird spacing to me, is this part of the style...?

platform/drm/cog-platform-drm.c
```diff
@@ -214,119 +204,83 @@ static struct {
...
-clear_buffers (void)
-{
-    drm_data.committed_buffer = NULL;
-
-    struct buffer_object *buffer, *tmp;
-    wl_list_for_each_safe (buffer, tmp, &drm_data.buffer_list, link) {
-        wl_list_remove (&buffer->link);
-        wl_list_remove (&buffer->destroy_listener.link);
+    if (params_string) {
+        g_auto(GStrv) params = g_strsplit(params_string, ",", 0);
```
Cwiiis, line 248: I guess if you wanted to save some memory here you could use g_strdelimit and transform commas into NUL bytes... But maybe that's being too smart :)

platform/drm/cog-platform-drm.c
```diff
@@ -214,119 +204,83 @@ static struct {
...
+        for (unsigned i = 0; params[i]; i++) {
+            g_auto(GStrv) kv = g_strsplit(params[i], "=", 2);
```
Cwiiis, line 250: And same here.

Cwiiis (approved):
I have some comments, but this looks really good :) I was less thorough in the DRM renderer, but as I read it, that's essentially just moved code. I think the term "renderer" is fine and fits well, but if you were looking for alternatives, "presenter" would make sense, I guess?

zdobersek (approved):

platform/common/cog-gl-utils.c
```diff
@@ -0,0 +1,83 @@
...
+
+    GLint status = 0;
+    glGetProgramiv(program, GL_LINK_STATUS, &status);
+    if (status)
+        return true;
+
+    GLint log_length = 0;
+    glGetProgramiv(program, GL_INFO_LOG_LENGTH, &log_length);
+    {
+        char log[log_length + 1];
```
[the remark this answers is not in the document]
aperezdc: Yes, variable length arrays (VLAs) have [made it to C99](https://en.wikipedia.org/wiki/Variable-length_array#C99). They are quite handy IMO, but I am open to change this if people think it's a bad idea to use them.

platform/common/cog-gl-utils.c
```diff
@@ -0,0 +1,83 @@
...
+        return 0;
+    }
+
+    glCompileShader(shader);
+    if ((err = glGetError()) != GL_NO_ERROR) {
+        g_set_error_literal(error, COG_PLATFORM_EGL_ERROR, err, "Cannot compile shader");
+        return 0;
+    }
+
+    GLint ok = GL_FALSE;
```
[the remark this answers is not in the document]
aperezdc: Sure, will do. The name you suggest is indeed more descriptive.

platform/common/cog-gl-utils.h
```diff
@@ -0,0 +1,26 @@
...
+#include <glib.h>
+
+G_BEGIN_DECLS
+
+typedef GLuint CogGLShaderId;
+
+void cog_gl_shader_id_destroy(CogGLShaderId *shader_id);
+G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC(CogGLShaderId, cog_gl_shader_id_destroy)
+
+CogGLShaderId cog_gl_shader_id_steal(CogGLShaderId *shader_id) G_GNUC_WARN_UNUSED_RESULT;
```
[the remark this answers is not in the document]
aperezdc: Good point, I think we could tell CMake to pass `-fvisibility=hidden` when building plug-in modules; the only symbols that need to be public are already marked with `G_MODULE_EXPORT` which arranges for the symbol to have `__attribute__((visibility("default")))`. I think we can do this separately, so I have filed #380 to track it.

platform/drm/cog-drm-gles-renderer.c
```diff
@@ -0,0 +1,567 @@
...
+    EGLDisplay egl_display;
+    EGLConfig  egl_config;
+    EGLContext egl_context;
+    EGLSurface egl_surface;
+
+    GLuint gl_program;
+    GLuint gl_texture;
+    GLint  gl_attrib_position;
+    GLint  gl_attrib_texture;
+    GLint  gl_uniform_texture;
```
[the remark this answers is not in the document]
aperezdc: I am writing the following next to the variables for my updated patch set:

```c
    /*
     * The renderer wraps frames provided by wpebackend-fdo into a texture,
     * then paints a quad which samples said texture. A simple GLSL shader
     * program uses the "position" attribute to fetch the coordinates of
     * the quad, and the "texture" attribute the UV mapping coordinates for
     * the texture. Rotation is achieved by changing the UV mapping. The
     * "texture" uniform is used to reference the texture unit where the
     * frame textures get loaded.
     */
```

Please let me know if this may need further clarification.

platform/drm/cog-drm-gles-renderer.c
```diff
@@ -0,0 +1,567 @@
...
+                                         strides, offsets, modifiers, &fb_id, flags);
+    if (ret) {
+        handles[0] = gbm_bo_get_handle(bo).u32;
+        handles[1] = handles[2] = handles[3] = 0;
+        strides[0] = gbm_bo_get_stride(bo);
+        strides[1] = strides[2] = strides[3] = 0;
+        offsets[0] = offsets[1] = offsets[2] = offsets[3] = 0;
+        ret = drmModeAddFB2(drm_fd, self->mode.hdisplay, self->mode.vdisplay, self->gbm_format, handles, strides,
+                            offsets, &fb_id, 0);
+    }
```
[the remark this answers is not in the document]
aperezdc: Okay, I have been reading around and the “retry without modifiers on failure” dance (which I cargo-culted from the existing DRM code, maybe itself cargo-culted from elsewhere) is needed because BO creation typically does _not_ query the KMS subsystem for appropriate/needed modifiers, but at some point the ultimate responsible for allocating the BO could have picked some modifiers which may or may not be supported by the KMS plane used for presenting the framebuffer. Retrying without the modifiers can get things working in some cases and I have seen it as a recommendation [e.g. in the LKML](https://lkml.org/lkml/2020/7/1/1118)—to be done _even_ if `DRM_CAP_ADDFB2_MODIFIERS` is supported by the DRM/KMS output.

I'll add a comment here about this.

platform/drm/cog-drm-modeset-renderer.c
```diff
@@ -0,0 +1,694 @@
...
+        return;
+
+    wl_shm_buffer_begin_access(shm_buffer);
+
+    uint8_t *src = wl_shm_buffer_get_data(shm_buffer);
+    uint8_t *dst = map_data;
+
+    uint32_t bo_width = gbm_bo_get_width(bo);
+    uint32_t bo_height = gbm_bo_get_height(bo);
+    if (!(width == bo_width && height == bo_height && stride == bo_stride)) {
```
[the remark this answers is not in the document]
aperezdc: If the dimensions match, including the stride, that means the allocated size for both buffers is the same, and that the disposition of pixels inside them match because pixel format is the same for both. That is the reason why there are no bounds check in the `else` part.

The BO is created with the same dimensions and pixel format as SHM buffers the first time it is needed, so I would expect that at least width and height always match, given that `wpe_view_backend_dispatch_set_size()` is never being called in the DRM/KMS platform plug-in.

OTOH, when the stride does not match, I do wonder if it's faster to let the compiler derive the memory copy with a double loop, or whether it would be better to do a row-by-row copy, or maybe even offload to Pixman, which is likely to have hand-optimized assembler pixel copy routines for most common platforms 🤔 

FWIW, this was existing code that got moved around, so I would rather leave it as-is for now. Originally this was written by @zdobersek, so if he wants to comment here about possible improvements, I will be happy to do a follow-up patch.

platform/common/cog-gl-utils.c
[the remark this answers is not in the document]
crzwdjk: VLAs are considered somewhat harmful for a number of reasons, given that they're basically the equivalent of alloca() and could use an unbounded amount of stack memory without any way to detect that. In this case, it's an error path so I don't think it would hurt anything to do a heap allocation.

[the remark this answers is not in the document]
aperezdc: Sure, I can turn this into a `g_new()` and allocate the array on the heap. Will do!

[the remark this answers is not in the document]
aperezdc: Actually, I am also going to replace the static 4096-byte buffer used to get the shader compilation log in `cog_gl_load_shader()` to retrieve the log length with `glGetShaderiv()` and allocate a buffer dynamiically.

platform/drm/cog-drm-renderer.c
aperezdc: So far both renderers added by this patch set have one, so let's `g_assert(self->destroy)` for the sake of simplicity. If the need to have a renderer without a destruction function comes up later, we can always rethink this.

aperezdc: Also I think it's somewhat silly to have a `cog-drm-renderer.c` file with a single two-line function so I will remove it and put the `cog_drm_renderer_destroy()` in the header as a `static inline`.

platform/drm/cog-platform-drm.c
aperezdc: Yes, this is `clang-format` being funny about aligning variable names; the `g_autoptr()` macro makes it look even weirder than it would normally be. I am not 100% convinced about that formatting rule. Even funnier: there is a similar instance right above which is in theory badly formatted, and `clang-format` does not complain about it 🙃

platform/drm/cog-platform-drm.c
aperezdc: I thought of that but it would still need to `g_strdup(params_string)` to have a mutable copy, plus this is done only once during initialization so it is definitely not memory or performance critical. In the end I went with code clarity instead of saving a few allocations.

platform/drm/cog-platform-drm.c
aperezdc: Same reason as above, I went for readability/clarity.

platform/drm/cog-platform-drm.c
aperezdc: Oh well, I went ahead and tried locally disabling the `AlignConsecutiveDeclarations` option locally and it indeed makes this ugly instance go away, but at the cost of making other parts of the code not as neat and readable. I'll workaround adding an empty line between the `GError` declaration and the next line, to avoid the formatter trying to align them in this particular case.

aperezdc:
I have addressed all the comments, and cannot come up with a name better than “renderer”. Given that the patch set got only minor changes after review I'll go ahead and land it once CI passes.

Thanks @Cwiiis, @crzwdjk, and @zdobersek for reviewing!

PR #372, opened 2021-11-06
docs: Fix building documentation with libsoup3

aperezdc:
Pass the correct `pkg-config` module name to `gi-docgen` depending on the version of libsoup in use: WPE WebKit builds with libsoup2 have API version 1.0, and builds with libsoup3 have API version 1.1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,11 +126,13 @@ set(COGCORE_SOURCES
 )
 
 if (USE_SOUP2)
+    set(WPEWEBKIT_PC_NAME "wpe-webkit-1.0")
     set(REQUIRED_GIO_VERSION "gio-2.0>=2.44")
     set(REQUIRED_SOUP_VERSION "libsoup-2.4")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -126,11 +126,13 @@ set(COGCORE_SOURCES
 )
 
 if (USE_SOUP2)
+    set(WPEWEBKIT_PC_NAME "wpe-webkit-1.0")
```
psaavedra, line 129: AFAIK this assume using libsoup2 only with old versions of WebKit (<2.34) but ot is possible setup libsoup2 as well with the latest version of WebKit.

aperezdc: Note that this is only the name of the `pkg-config` module. The suffix (`1.0` or `1.1`) is the **API** version—not the WPE WebKit version. When built with libsoup2 the API version of WPE WebKit is always `1.0`, and when built with libsoup3 it is always `1.1`.

psaavedra (approved):

PR #373, opened 2021-11-08
core: Remove unneeded CogPlatformClass.teardown vfunc

aperezdc:
Use `GObjectClass.finalize` instead of `CogPlatformClass.teardown`. Now that `CogPlatform` derives from `GObject`, it is possible to reuse its functionality instead of having an ad-hoc vfunc that gets called immediately before destroying an instance. As a bonus, not only the implementations follow a more idiomatic GObject approach, but also it is not possible anymore to forget a call to `cog_platform_teardown()` before destruction.

### DIFF
--- a/cog.c
+++ b/cog.c
@@ -342,13 +342,7 @@ platform_setup (CogShell *shell)
 static void
 on_shutdown (CogLauncher *launcher G_GNUC_UNUSED, void *user_data G_GNUC_UNUSED)
 {
-    g_debug ("%s: Platform = %p", __func__, s_options.platform);
-
-    if (s_options.platform) {
-        cog_platform_teardown (s_options.platform);
-        g_clear_object(&s_options.platform);
-        g_debug ("%s: Platform teardown completed.", __func__);
-    }
+    g_clear_object(&s_options.platform);
 }
 
 static void*
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -93,16 +93,6 @@ cog_platform_new(const char *name, GError **error)
     return g_steal_pointer(&self);
 }
 
-void
-cog_platform_teardown(CogPlatform *platform)
-{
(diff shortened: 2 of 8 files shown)


aperezdc:
Heh, it seems even `clang-format` having access to a full blown C language parser gets confused sometimes 🤣 — this is clearly a bogus report:

```diff
--- platform/headless/cog-platform-headless.c	(before formatting)
+++ platform/headless/cog-platform-headless.c	(after formatting)
@@ -31,7 +31,7 @@
     guint tick_source;
     gboolean frame_complete;
     struct wpe_view_backend_exportable_fdo *exportable;
-    WebKitWebViewBackend                   *view_backend;
+    WebKitWebViewBackend *                  view_backend;
 };
 
 static struct platform_window win = {
```

cog.c
```diff
@@ -344,11 +344,9 @@ on_shutdown (CogLauncher *launcher G_GNUC_UNUSED, void *user_data G_GNUC_UNUSED)
...
     g_debug ("%s: Platform = %p", __func__, s_options.platform);
 
-    if (s_options.platform) {
-        cog_platform_teardown (s_options.platform);
-        g_clear_object(&s_options.platform);
-        g_debug ("%s: Platform teardown completed.", __func__);
-    }
+    g_clear_object(&s_options.platform);
+
+    g_debug("%s: Platform destroyed.", __func__);
```
Cwiiis, line 349: Does it still make sense to have this debug log here? The platform can be destroyed elsewhere, maybe this should be "on_shutdown called" instead, or this statement moved to the finalize function?

Cwiiis (approved):
Much nicer :)

cog.c
aperezdc: Let's remove this debugging statement, I think it's not particularly useful.

aperezdc:
The cross-builds are failing due to #374 at the moment, but I think it's safe to land this anyway.

Issue #374, opened 2021-11-08
CI: Cross-build images for ARM do not detect wayland-scanner


aperezdc:
For example this can be seen in this [build log](https://github.com/Igalia/cog/runs/4140268316), apparently `wayland-scanner` exists but then some check fails:

```
-- Checking for module 'wayland-scanner'
-- Could NOT find WaylandScanner (missing: WAYLAND_SCANNER WAYLAND_SCANNER_VERSION WAYLAND_SCANNER_CODE_ARG) 
-- Checking for module 'wayland-protocols'
--   Found wayland-protocols, version 1.21
CMake Error at /home/runner/toolchain/sysroots/x86_64-pokysdk-linux/usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find WaylandProtocols (missing: WAYLAND_SCANNER)
Call Stack (most recent call first):
  /home/runner/toolchain/sysroots/x86_64-pokysdk-linux/usr/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindWaylandProtocols.cmake:202 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  platform/wayland/CMakeLists.txt:16 (find_package)
```

I wonder if this might be related to the wildcard I added in c8756bd074fdc936a722052e4775424aaecb32a0 😇

psaavedra:
it could useful to verify the wayland-scanner binary is provided or not by the toolkit. If not I should fix that.

aperezdc:
> it could useful to verify the wayland-scanner binary is provided or not by the toolkit. If not I should fix that.

It turns out that `wayland-scanner` is in the toolchain, both the ARM and x86_64 versions:

```
% fd -tx -F wayland-scanner ~/tc -X file -b '{}'       
ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-armhf.so.3, BuildID[sha1]=d2eef2931c8a381c1f14961848273aab75905ff2, for GNU/Linux 3.2.0, stripped
[...]

PR #375, opened 2021-11-08
CI: Switch to clang-format version 13

aperezdc:
Update the `check-style` script and the CI job to use `clang-format` 13, which is the first version that supports combining `PointerAlignment: Right` with `AlignConsecutiveDeclarations: true`. This was unimplemented until recently, see https://reviews.llvm.org/D103245

While at it, do not let clang-format choose a different way of doing pointer alignment by setting `DerivePointerAlignment: false`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -61,7 +61,7 @@ ConstructorInitializerIndentWidth: 4
 ContinuationIndentWidth: 4
 Cpp11BracedListStyle: true
 DeriveLineEnding: true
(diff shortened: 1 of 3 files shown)


Cwiiis (approved):

aperezdc:
As with another PR from earlier, the cross build is failing on #374 ­— unrelated to this.

Thanks @Cwiiis!

PR #376, opened 2021-11-10
cog: Add command line options for memory limits

aperezdc:
Add a new `memory-limits` option group with options to control WebKit's internal memory pressure handler settings:

```
% cog --help-memory-limits
Usage:
  cog [OPTION?] [URL]

These options allow configuring WebKit's memory pressure handling mechanism.

  In particular, a limit for the maximum amount of memory to use can be set,
  and thresholds relative to the limit which determine at which points memory
  will be reclaimed. The conservative threshold is typically lower when reached
  memory will be reclaimed; the strict threshold works in the same way but the
  process is more aggresive. The kill threshold configures when worker processes
  will be forcibly killed. Note that if there is no memory limit set, the other
  settings are ignored.

  --web-mem-limit=SIZE                                Maximum amount of memory to use, in MiB.
  --web-check-interval=SECONDS                        Interval of time between memory usage checks.
  --web-conservative-threshold=(0..1)                 Conservative threshold (default: 0.33).
  --web-strict-threshold=(0..1)                       Strict threshold (default: 0.5).
  --web-kill-threshold=(0..                           Kill threshold (default: 0).
  --net-mem-limit=SIZE                                Maximum amount of memory to use, in MiB.
  --net-check-interval=SECONDS                        Interval of time between memory usage checks.
  --net-conservative-threshold=(0..1)                 Conservative threshold (default: 0.33).
  --net-strict-threshold=(0..1)                       Strict threshold (default: 0.5).
  --net-kill-threshold=(0..                           Kill threshold (default: 0).

%
```

In the end the main complication in this PR is moving everything around to make `CogLauncher` an implementation detail of the `cog` program. The reason for this is twofold:

- `CogLauncher` does not fit particularly well as public API in `libcogcore` because it's rigid and third-party launcher that use `libcogcore` reach out for `CogShell`, the platform modules, and their loader.
- `CogLauncher` created early a `CogShell` and in `cog.c` we were relying on that to directly configure things from the command line options, which is okay until the moment we need to process all the configuration options to prepare a `WebKitMemoryPressureSettings` instance _before_ the `WebKitWebContext` is created during the initialization of `CogShell`: the memory settings are a construct-only property of the web context—we have a chicken-and-egg problem! 🥚🐔

By conflating most of `cog.c` into `cog-launcher.c` not only evreything lives in the same place and is a bit more simplified, but also it makes it possible parse all the configuration in one step, then instantiate the `CogShell` later.

There are a number of possibilities for further improvements, but I would rather leave those for follow-up PRs.


aperezdc:
Probably @magomez is the most suited to review this, as he implemented the API in WebKit. While at it, a question: should I also add handling of `WEBKIT_WEB_PROCESS_TERMINATED_BY_API` to cover the case in which the memory limits results in a process being killed, or is that unrelated? (The API reference docs are not clear on this!)

magomez:
> Probably @magomez is the most suited to review this, as he implemented the API in WebKit. While at it, a question: should I also add handling of `WEBKIT_WEB_PROCESS_TERMINATED_BY_API` to cover the case in which the memory limits results in a process being killed, or is that unrelated? (The API reference docs are not clear on this!)

WEBKIT_WEB_PROCESS_TERMINATED_BY_API is emitted when the web process is killed because the app has called webkit_web_view_terminate_web_process, so it's unrelated to this case.

When the web process is killed because it exceeded the memory limit, the web-process-terminated signal will be emitted with the value WEBKIT_WEB_PROCESS_EXCEEDED_MEMORY_LIMIT.

aperezdc:
@magomez Thanks a lot for your comments, clearly I need to rework this a bit and have a second set of CLI options for the web process!

aperezdc:
Gah, I have a good yak shave going on: the `WebKitWebContext` is created by `CogShell`, and that in turn is created inside `CogLauncher`, but we need the `CogLauncher` instance pre-created in order to dynamically add the `GOptionGroup` for all the web settings and a couple of other things...

Given my previous experiences, my impression is that `CogLauncher` is quite specific to the implementation of our `cog` program, and quite likely not worth it having in `libcogcore`. Most third party launchers which use `libcogcore` ditch `CogLauncher` and start at `CogShell` instead Therefore I am thinking that it would probably be better to remove `CogLauncher` from the library and use it only as an implementation detail of the `cog` program, and rewrite parts of it to accomodate things better.

aperezdc:
Probably it would be a good idea to review this commit by commit, to make better sense of things 😉

TingPing (approved):
I looked through this in detail, it looked good to me.

aperezdc:
@TingPing: Thanks!

PR #377, opened 2021-11-11
CI: Improve cross-toolchain reliability

### LINKED ISSUE #374: CI: Cross-build images for ARM do not detect wayland-scanner

aperezdc:
This does two things:

- Switches the ARM cross-toolchain builds to use the system-provided `wayland-scanner` binary, as it seems that newer versions of the toolchains either do not include it, or it is an ARM binary and thus CMake cannot run it to determine its version. Eventually this should be solved in the toolchain.
- Make toolchain download retries more reliable by also inspecting the exit code from the `curl` program and acting accordingly.

Closes #374

### DIFF
--- a/.github/toolchain.sh
+++ b/.github/toolchain.sh
@@ -3,7 +3,7 @@ set -eu -o pipefail
 
 # Info on available toolchains: https://wk-contrib.igalia.com/
 INSTALL_DIR=${1:-${HOME}/toolchain}
-BASEURL=https://wk-contrib.igalia.com/yocto/meta-perf-browser/browsers/nightly/sdk
+BASEURL=https://wk-contrib.igalia.com/yocto/meta-perf-browser/browsers/stable/sdk
 FILE=wandboard-mesa/browsers-glibc-x86_64-core-image-weston-browsers-cortexa9t2hf-neon-wandboard-mesa-toolchain-1.0.sh
 
 declare -r INSTALL_DIR BASEURL FILE
--- a/.github/toolchain.sh
+++ b/.github/toolchain.sh
@@ -1,5 +1,5 @@
 #! /bin/bash
-set -e
+set -eu -o pipefail
 
 # Info on available toolchains: https://wk-contrib.igalia.com/
 INSTALL_DIR=${1:-${HOME}/toolchain}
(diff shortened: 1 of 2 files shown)


psaavedra:
El jue, 11 nov 2021 a las 10:11, Adrian Perez ***@***.***>)
escribió:

>
> Also, I have no idea if it would be better to be on the “candidate“
> toolchain, I would need some input from @psaavedra
> <https://github.com/psaavedra> to know what the differences are. I
> suppose the difference is the version of WPE WebKit included? If that is
> the case, I think for now it should be okay to use the stable one, and if
> some new feature we want to add needs newer versions we can consider
> switching to “candidate”.
>

In general, `stable` version should be more than enough. "candidate" is
only a transitional state between "nightly" -> "stable" when there are new
changes considered to be added on "stable"  so please don't use
"candidate".

I suppose, you set 'nightly' at some point because some missing recent
dependency only available in meta-webkit "main" and not incorporated to the
stable image.

psaavedra (approved):

aperezdc:
Yay, it's working! I'll remove the `set -x` that I added in the `toolchain.sh` script for debugging and then merge this. Thanks @psaavedra!

aperezdc:
> In general, `stable` version should be more than enough. "candidate" is only a transitional state between "nightly" -> "stable" when there are new changes considered to be added on "stable" so please don't use "candidate". I suppose, you set 'nightly' at some point because some missing recent dependency only available in meta-webkit "main" and not incorporated to the stable image.

Okay, I'll also set it back to “stable” before landing 👌🏼

PR #378, opened 2021-11-24
cmake: Set default module location from CMAKE_INSTALL_LIBDIR

aperezdc:
Derive the location where to install platform modules from the `CMAKE_INSTALL_LIBDIR` variable (instead of `CMAKE_INSTALL_PREFIX`), instead of assuming that the directory is always called `lib/`
inside the prefix. This provides a saner default for distributors which use other kinds of locations, like multilib/multiarch systems.

----

(Thanks to @bertogg for spotting the issue!)


bertogg (approved):
Yep, I applied this downstream in Debian already.

aperezdc:
Cherry-picked as commit d6eee034102ef50dbb2c4c6a427ad6d6443d265e in the `cog-0.12` branch.

PR #384, opened 2021-12-11
gtk4: Fix mouse events handling

philn:
cog_gtk4_platform_setup() is called before the CogShell had a chance to set its device scale factor
property, so we need to watch for property update notifications.

### DIFF
--- a/platform/gtk4/cog-platform-gtk4.c
+++ b/platform/gtk4/cog-platform-gtk4.c
@@ -747,8 +747,16 @@ setup_fdo_exportable(struct platform_window* window)
     g_assert_nonnull(window->view_backend);
 }
 
+static void
+shell_device_factor_changed(CogShell *shell, GParamSpec *param_spec, gpointer data)
+{
+
+    struct platform_window *window = (struct platform_window *) data;
+    window->device_scale_factor = cog_shell_get_device_scale_factor(shell);
+}
+
 static gboolean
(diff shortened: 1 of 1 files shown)


philn:
Not sure this the right fix, it's weird the platform backend gets initialized before the shell is fully configured?

philn:
Also, shouldn't `wpe_view_backend_dispatch_set_device_scale_factor()` return an error or something when it receives a 0 factor? Because that's wrecking havoc :)

philn:
Maybe @aperezdc can review this? 🙏🏻

aperezdc (approved):
The breakage was unintentional, and I think at some point we want to define how options are defined in som more formal way, and apply the configuration logic consistently to all of them. There is a proposal in this regard in https://github.com/Igalia/cog/pull/363, currently open for discussion, which takes some bits from the `devel/multiple-views` branch and adds on top sample code to read the config file (in fact one of the things it does was fixing setting `CogDrmPlatform.device-scale-factor`!)

Anyhoo, I digress—let's have this merged for now. Thanks!

aperezdc:
> Also, shouldn't `wpe_view_backend_dispatch_set_device_scale_factor()` return an error or something when it receives a 0 factor? Because that's wrecking havoc :)

Probably we want it to check that values are in some reasonable range. Setting it to big values, let's say 1000, will also cause amusing things to happen. Reported in https://github.com/WebPlatformForEmbedded/libwpe/issues/89 to fix in libwpe.

PR #387, opened 2021-12-22
cmake: Turn COG_MODULEDIR into an absolute path

aperezdc:
Make the `COG_MODULEDIR` macro expand to an absolute path at runtime to ensure that modules can always be found correctly. Otherwise, it is possible to configure the build e.g. with `COG_MODULEDIR=lib/cog/modules` (basically what is done after a18626b42c58c57208a5c17bd25d09e5d2aa5a16) and the relative path would end up in the built programs, resulting in failure to find the modules (unless, by chance, they are executed while the current directory matches the configured installation prefix.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,19 @@ set(COG_APPID "" CACHE STRING "Default GApplication unique identifier")
 set(COG_HOME_URI "" CACHE STRING "Default home URI")
 
 set(COG_MODULEDIR "${CMAKE_INSTALL_LIBDIR}/cog/modules"
-    CACHE STRING "Default search path for loadable modules")
+    CACHE PATH "Default search path for loadable modules")
+
+# Make sure that the module path is always absolute. If the specified
+# value is relative, this makes it a child of CMAKE_INSTALL_PREFIX.
+if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.20)
+    cmake_path(ABSOLUTE_PATH COG_MODULEDIR BASE_DIRECTORY ${CMAKE_INSTALL_PREFIX} NORMALIZE)
+elseif (NOT IS_ABSOLUTE ${COG_MODULEDIR})
+    set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/${COG_MODULEDIR}")
+endif ()
+if (NOT IS_ABSOLUTE ${COG_MODULEDIR})
+    message(FATAL_ERROR "Cog module location '${COG_MODULEDIR}' is not an absolute path")
+endif ()
+message(STATUS "Cog module path: ${COG_MODULEDIR}")
 
 if (NOT COG_APPID OR COG_APPID STREQUAL "")
     set(COG_DEFAULT_APPID com.igalia.Cog)


clopez (approved):

aperezdc:
This patch is going to need a fallback for older versions of CMake because the [cmake_path()](https://cmake.org/cmake/help/v3.22/command/cmake_path.html) was added in 3.20 and some distributions we still want to support are shipping older versions.

clopez:
> This patch is going to need a fallback for older versions of CMake because the [cmake_path()](https://cmake.org/cmake/help/v3.22/command/cmake_path.html) was added in 3.20 and some distributions we still want to support are shipping older versions.

Oh yes... 3.20 is pretty new
Ideally we should no require a CMake newer than WPE one which is currently setting a minimum of 3.12

clopez:
I have tested this with CMake 3.13 and the message printed on stdout when configuring had the correct absolute path, however in CMakeCache.txt the value wrote was the relative path.
I think is because the value `COG_MODULEDIR` is declared as `CACHE` so further updates to it don't affect the value that is stored in Cache, a bit weird this behavior. See [doc about set()](https://cmake.org/cmake/help/v3.12/command/set.html#set-cache-entry)
I ended making it work with this patch over yours:

```patch
 # Make sure that the module path is always absolute. If the specified
 # value is relative, this makes it a child of CMAKE_INSTALL_PREFIX.
 if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.20)
     cmake_path(ABSOLUTE_PATH COG_MODULEDIR BASE_DIRECTORY ${CMAKE_INSTALL_PREFIX} NORMALIZE)
 elseif (NOT IS_ABSOLUTE ${COG_MODULEDIR})
-    set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/${COG_MODULEDIR}")
+    set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/${COG_MODULEDIR}"
+        CACHE PATH "Default search path for loadable modules" FORCE)
 endif ()

```
I didn't tested it with CMake >= 2.20

clopez (approved):

aperezdc:
> I have tested this with CMake 3.13 and the message printed on stdout when configuring had the correct absolute path, however in CMakeCache.txt the value wrote was the relative path. I think is because the value `COG_MODULEDIR` is declared as `CACHE` so further updates to it don't affect the value that is stored in Cache, a bit weird this behavior. See [doc about set()](https://cmake.org/cmake/help/v3.12/command/set.html#set-cache-entry) I ended making it work with this patch over yours:
> 
> ```diff
>  # Make sure that the module path is always absolute. If the specified
>  # value is relative, this makes it a child of CMAKE_INSTALL_PREFIX.
>  if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.20)
>      cmake_path(ABSOLUTE_PATH COG_MODULEDIR BASE_DIRECTORY ${CMAKE_INSTALL_PREFIX} NORMALIZE)
>  elseif (NOT IS_ABSOLUTE ${COG_MODULEDIR})
> -    set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/${COG_MODULEDIR}")
> +    set(COG_MODULEDIR "${CMAKE_INSTALL_PREFIX}/${COG_MODULEDIR}"
> +        CACHE PATH "Default search path for loadable modules" FORCE)
>  endif ()
> ```
> 
> I didn't tested it with CMake >= 2.20

We don't need to do this extra work: the value saved in the cache is the one given by the user in the command line with `-DCOG_MODULEDIR=…` or the default value. Then when CMake runs over the script it will check the value from the cache and calculate the absolute path if needed. Note how `<builddir>/cog-config.h` gets the absolute path correctly, because expansion of variables in configured files uses the values of _live_ variables and not the cached ones (as it should):

```
% cmake -S. -Bbuild1
...
-- Cog module path: /usr/local/lib/cog/modules
...
% grep MODULE build1/cog-config.h                            
#define COG_MODULEDIR "/usr/local/lib/cog/modules"
%
```

Or specifying some path:

```
% cmake -S. -Bbuild2 -DCOG_MODULEDIR=/usr/libexec/cog
...
% grep MODULE build2/cog-config.h       
#define COG_MODULEDIR "/usr/libexec/cog"
%
```

PR #388, opened 2022-01-11
gtk4: Remove the requirement of a shader language version.

alexgcastro:
The shader does not need this version.

### DIFF
--- a/platform/gtk4/cog-platform-gtk4.c
+++ b/platform/gtk4/cog-platform-gtk4.c
@@ -84,8 +84,7 @@ static struct platform_window win = {
     .device_scale_factor = 1,
 };
 
-static const char s_vertex_shader[] = "#version 330\n"
-                                      "attribute vec2 pos;\n"
+static const char s_vertex_shader[] = "attribute vec2 pos;\n"
                                       "attribute vec2 texture;\n"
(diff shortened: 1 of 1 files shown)


aperezdc (approved):
Nice, thanks for the patch!

Cwiiis:
Looks like this is causing cog to fail to start for me with

`Cog-Gtk4-WARNING **: 07:47:46.258: Shader setup failed: Shader compilation: 0:1(1): error: syntax error, unexpected NEW_IDENTIFIER`

Perhaps a version is required?

aperezdc:
> Looks like this is causing cog to fail to start for me with
> 
> `Cog-Gtk4-WARNING **: 07:47:46.258: Shader setup failed: Shader compilation: 0:1(1): error: syntax error, unexpected NEW_IDENTIFIER`
> 
> Perhaps a version is required?

Ah, interesting! Which hardware/driver are you using? The X11 and DRM/KMS platforms use very similar shaders (if not the same) and there they do not have any `#version` directive, and those work for me with the `amdgpu` and `i915` Mesa drivers 🤔 

I should probably read a bit about GLSL and what `#version` exactly does and figure out what exactly we need to do here.

PR #389, opened 2022-01-11
Reinstate shader version in GTK4 backend

Cwiiis:
Reinstate shader version declarations, but use #version 100 (1.00 ES)
instead of #version 330 (3.3).

### DIFF
--- a/platform/gtk4/cog-platform-gtk4.c
+++ b/platform/gtk4/cog-platform-gtk4.c
@@ -84,7 +84,8 @@ static struct platform_window win = {
     .device_scale_factor = 1,
 };
 
-static const char s_vertex_shader[] = "attribute vec2 pos;\n"
+static const char s_vertex_shader[] = "#version 100\n"
+                                      "attribute vec2 pos;\n"
                                       "attribute vec2 texture;\n"
                                       "varying vec2 v_texture;\n"
                                       "void main() {\n"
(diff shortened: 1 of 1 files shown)


alexgcastro:
FTR this option works for me too, I've just tested it.

platform/gtk4/cog-platform-gtk4.c
```diff
@@ -84,15 +84,17 @@ static struct platform_window win = {
     .device_scale_factor = 1,
 };
 
-static const char s_vertex_shader[] = "attribute vec2 pos;\n"
+static const char s_vertex_shader[] = "#version 100\n"
```
aperezdc, line 87: As discussed in a chat conversation, all ES contexts are expected to support GLSL 1.0, so this is expected to always work regardless of the driver/hardware combination. These shaders are simple enough that they do not use any feature outside of version 1.0—all good!

OTOH, for the sake of correctness we should add the same to the similar (if not same) shaders in `cog-platform-x11.c` and `cog-drm-gles-renderer.c`. It would be fine to do so in a follow-up patch, so we unblock people wanting to use the GTK4 plugin ASAP.

aperezdc (approved):

platform/gtk4/cog-platform-gtk4.c
aperezdc: I have made all the platform plug-ins use exactly the same shaders and textured quad painting code in #392 😸

PR #392, opened 2022-01-12
drm, x11, gtk4: Refactor GL textured quad painting code

aperezdc:
This patch set moves the code that paints textured quads out from the DRM platform into the shared GL utility code, and makes the DRM, X11, and GTK4 platform use it—instead of each of them having slightly different copies of the code.

1. First, code that will be shared is moved from `cog-drm-gles-renderer.c` and into `cog-gl-utils.c`, and the DRM plug-in modified accordingly. This adds a new `CogGLRenderer` struct, functions to (de)initialize it, and for painting, accompanied by a comment explaining how to use the shared code.
2. The X11 plug-in is changed to link `libepoxy` in preparation for the next step.
3. The X11 and GTK4 plug-ins get their copy of the code removed and changed to use `CogGLRenderer` instead.

### DIFF
--- a/platform/common/cog-gl-utils.c
+++ b/platform/common/cog-gl-utils.c
@@ -1,6 +1,6 @@
 /*
  * cog-gl-utils.c
- * Copyright (C) 2021 Igalia S.L.
+ * Copyright (C) 2021-2022 Igalia S.L.
  *
  * Distributed under terms of the MIT license.
  */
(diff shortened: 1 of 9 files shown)


platform/gtk4/cog-platform-gtk4.c
```diff
@@ -41,8 +41,12 @@ G_DEFINE_DYNAMIC_TYPE_EXTENDED(
     g_io_extension_point_implement(COG_MODULES_PLATFORM_EXTENSION_POINT, g_define_type_id, "gtk4", 400);)
 
 /*
- * TODO
- * - multi-views
+ * TODO:
+ * - Multiple views.
+ * - Call cog_gl_renderer_finalize() when GL area is being destroyed.
+ * - Implement an actual widget that wraps exported the EGLImages into a
```
Cwiiis, line 47: extraneous 'the' after 'exported'

platform/x11/cog-platform-x11.c
```diff
@@ -643,16 +591,16 @@ init_egl (void)
...
-        s_window->egl.surface = s_display->egl.create_platform_window_surface (s_display->egl.display,
-                                                                               s_display->egl.config,
-                                                                               &win, NULL);
+    if (epoxy_has_egl_extension(s_display->egl.display, "EGL_EXT_platform_x11")) {
+        s_window->egl.surface =
+            eglCreatePlatformWindowSurfaceEXT(s_display->egl.display, s_display->egl.config, &win, NULL);
+    } else {
+        s_window->egl.surface =
+            eglCreateWindowSurface(s_display->egl.display, s_display->egl.config, (EGLNativeWindowType) &win, NULL);
+    }
```
Cwiiis, line 600: I guess this is if we get given a display without the X11 extension, we still end up with a surface... But do we just not need the X11 extension? We don't look at `create_platform_window_surface` anymore, this seems like a change in behaviour and I don't really understand it - maybe a comment here is warranted?

Cwiiis (approved):
I think my comments could be addressed in follow-up if you want to get this landed quicker - this is a great change, extremely satisfying to see so much code removed for an overall improvement in behaviour :)

alexgcastro (commented):
LGTM

platform/common/cog-gl-utils.c
```diff
@@ -78,3 +78,150 @@ cog_gl_link_program(GLuint program, GError **error)
...
+        },
+        [COG_GL_RENDERER_ROTATION_270] = {
+            { 0, 1 }, { 0, 0 },
+            { 1, 1 }, { 1, 0 },
+        },
+    };
+    /* clang-format on */
+
+    glVertexAttribPointer(self->attrib_position, 2, GL_FLOAT, FALSE, 0, position_coords);
+    glVertexAttribPointer(self->attrib_texture, 2, GL_FLOAT, FALSE, 0, texture_coords[rotation]);
```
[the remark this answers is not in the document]
aperezdc: You are right, good point! We do not need to call `glVertexAttribPointer()` for each frame. I _think_ we still need the calls to `gl{Enable,Disable}VertexAttribArray()` before and after `glDrawArrays()`.

The rotation may change at any time during execution, so at least the second call to set the texture UV-mapping would need to be re-applied when it changes. I think we could add a helper `cog_gl_renderer_set_rotation()` function that takes care of it 🤔 

For now I prefer to leave the code as-is and only do the refactor here, and do the cleanup in a follow-up PR.

platform/common/cog-gl-utils.h
```diff
@@ -23,4 +24,61 @@ CogGLShaderId cog_gl_load_shader(const char *source, GLenum kind, GError **) G_G
...
+ *   - Embed a CogGLRenderer somewhere. Typically this will be a Cog
+ *     platform implementation.
+ *   - Initialize EGL, ensure eglBindAPI(EGL_OPENGL_ES_API) is called,
+ *     create an EGLContext suitable for rendering.
+ *   - With the EGLContext active, call cog_gl_renderer_initialize().
+ *
+ * - Painting:
+ *   - Activate the EGLContext.
+ *   - Optionally, paint before using the renderer (e.g. background below
+ *     the imagefilled background with a solid colour).
```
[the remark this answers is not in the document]
aperezdc: Ah, the phrasing is super awkward here and wrong, it should be:

> Optionally, paint before using the renderer (e.g. some solid color background below the image, shows if the web view and its content have transparency).

I'll change it before merging.

platform/gtk4/cog-platform-gtk4.c
aperezdc: The “the“ shall be removed before merging, indeed.

platform/x11/cog-platform-x11.c
aperezdc: The fallback is intended to be used for drivers which do not provide the extension (typically old-ish ones), assuming that `eglGetDisplay()` would have returned a correct kind of display (an X11 display in this case), then `eglCreateWindowSurface()` will work just fine... but it is now that I notice that we never do in the code a fallback from `eglGetPlatformDisplayEXT()` to `eglGetDisplay()` if the `EGL_EXT_platform_x11` extension is not available, so the fallback here is dead code and will never be reached. Let's remove it.

aperezdc: Also: we may want to `g_error()` inside `init_egl()` if `eglGetPlatformDisplayEXT()` cannot be used instead of continuing without a valid `EGLDisplay`. I'll add this.

platform/common/cog-gl-utils.c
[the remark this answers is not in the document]
aperezdc: Tracked in https://github.com/Igalia/cog/issues/393 🔍

aperezdc:
Thanks a lot @Cwiiis and @alexgcastro for reviewing these patches, and for the suggestions!

PR #397, opened 2022-01-25
wl: Improve handling of Wayland protocol interface versions

### LINKED ISSUE #396: Cog crash in wlroots compositors when built against recent Wayland versions
After this message, Cog crashes:

```
listener function for opcode 4 of wl_output is NULL
```

This is with:

- wayland 1.20.0
- wayland-protocols 1.24
- wlroots 0.15.0

The issue can be reproduced with both Sway and Cage.

aperezdc:
Define bounds to the versions of Wayland protocol interfaces requested from the compositor on calls to `wl_registry_bind()`. This prevents the compositor from using newer features for which there is no client side implementation in Cog.

While at it, improve checks on protocol bits that depend on the version of the protocols being used, mainly for `wl_seat` and `wl_pointer` (which is part of the former) to use the correct checks at runtime, and to feature-check at build time (instead of depending on version of the Wayland library headers present at build time).

Fixes #396

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -753,35 +753,35 @@ registry_global (void               *data,
     gboolean interface_used = TRUE;
 
     if (strcmp(interface, wl_compositor_interface.name) == 0) {
-        wl_data.compositor = wl_registry_bind(registry, name, &wl_compositor_interface, version);
+        wl_data.compositor = wl_registry_bind(registry, name, &wl_compositor_interface, 1);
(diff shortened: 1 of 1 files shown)


alex-voodoo (approved):
Looks good to me.

aperezdc:
Thanks @alex-voodoo for taking a look! 👀

aperezdc:
Cherry picked into the `cog-0.12` branch as commit d5ec9bdb4426f799aba20ba831a52539468d357e

PR #398, opened 2022-01-26
x11: Support setting view activity state flags

aperezdc:
Implement setting view activity state flags, which allows WebKit to perform certain optimizations (like skipping updates while the view is not present in a window, or not visible) and makes certain features work as expected (like the Page Visibility API).

- Events `XCB_FOCUS_{IN,OUT}` modify `wpe_view_activity_state_focused`.
- Events `XCB_{UN,}MAP_NOTIFY` modify `wpe_view_activity_state_in_window`.
- Event `XCB_VISIBILITY_NOTIFY` modifies `wpe_view_activity_state_visible`; both partially obscured and completely unobscured windows are set as visible.

The corresponding event masks are added during window creation.

Interestingly, at least on GNOME Shell in an X11 session, Cog does not get any `XCB_VISIBILITY_NOTIFY` events after the initial one, even if the window is hidden iconified. This is likely caused by the compositor wanting to have windows update so it can have an always a fresh view of the contents e.g. to paint window miniatures in a task switcher or in overview mode.

### DIFF
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -311,6 +311,34 @@ xcb_handle_motion_event (xcb_motion_notify_event_t *event)
 }
 
 static void
+view_backend_modify_activity_state(xcb_window_t window_id, enum wpe_view_activity_state state_flag, bool enable)
+{
+    struct CogX11Window *window = s_window;
+
+    if (window->xcb.window != window_id)
+        return;
+
+    if (enable)
+        wpe_view_backend_add_activity_state(window->wpe.backend, state_flag);
+    else
+        wpe_view_backend_remove_activity_state(window->wpe.backend, state_flag);
+}
+
+static inline void
+xcb_handle_visibility_event(const xcb_visibility_notify_event_t *event)
+{
+    switch ((xcb_visibility_t) event->state) {
+    case XCB_VISIBILITY_UNOBSCURED:
+    case XCB_VISIBILITY_PARTIALLY_OBSCURED:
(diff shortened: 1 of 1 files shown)


aperezdc:
I tried Cog with the X11 platform plug-in in a non-compositing window manager (Openbox), and the following happens (both with and without this PR applied) when moving another window in front of the Cog window:

Cog window with trails of a Xterm window left over it

I suspect this is caused by exposure events not being handled at all, and in compositing window managers it does not matter because they make the application believe they are not obscured. The code already had `XCB_EVENT_MASK_EXPOSURE` set on the window, but it is never handled in `xcb_process_events()`.

As the issue illustrated above happens *also* without the changes proposed in this PR, so I think that should be a new bug and not block merging.

aperezdc:
> I suspect this is caused by exposure events not being handled at all, and in compositing window managers it does not matter because they make the application believe they are not obscured. The code already had `XCB_EVENT_MASK_EXPOSURE` set on the window, but it is never handled in `xcb_process_events()`.

Indeed, this was the issue. I have pushed https://github.com/Igalia/cog/commit/030848c401c0c80d40e1956aab0aa8e6340266e6 to the [aperezdc/x11-expose](https://github.com/Igalia/cog/tree/aperezdc/x11-expose), which handles expose events by keeping the last exported image around and to re-paint it when needed (plus some accounting to avoid doing too many repaints). I will submit a PR with that after this one is merged.

Cwiiis (approved):
LGTM!

PR #409, opened 2022-03-04
x11: Fix underlinking when --as-needed is used

aperezdc:
Add the `x11-xcb` module as dependency as well. The corresponding library is needed for its `XGetXCBConnection` symbol, which will be reported as missing when the platform plug-in was built with the `--as-needed` linker flag.


ceyusa:
lgtm

PR #415, opened 2022-03-24
wl: Fix request of wl_seat version and guard feature usage

aperezdc:
Ensure that version selection for the `wl_seat` interface is inside the range *[3, 7]*; and properly guard usage of `wl_seat_listener::name`.


psaavedra (approved):
look right 👍

aperezdc:
Thanks @psaavedra 🤝

PR #420, opened 2022-04-07
Fix COG_WESTON_DIRECT_DISPLAY build

jameshilliard:
For some reason this option was enabling content protection as well, since that is a separate option that merely depends on direct display move it under COG_WESTON_CONTENT_PROTECTION.

We also need to fix pkg-config search to look for libweston protocols starting with weston version 8(first to introduce direct display), the current weston development branch uses version 11 so lets include that as well.

We also need to [fix the client protocol header names](https://github.com/Igalia/cog/blob/d2720103ff8a5c3c5c056487586ed0b5acdc2b91/cmake/FindWaylandProtocols.cmake#L142) as the current ones are invalid.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ set_project_version(0 13 3)
 calculate_library_versions_from_libtool_triple(COGCORE 9 2 0)
 
 project(cog VERSION "${PROJECT_VERSION}" LANGUAGES C)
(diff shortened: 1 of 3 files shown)


aperezdc (approved):
Changes LGTM. I will wait a few hours in case @philn wants to take a look before merging.

@jameshilliard Thanks for the fix!

philn (approved):

PR #422, opened 2022-04-07
Fix OpenGL4

ceyusa:
A proposal to make work cog (tested only for gtk4 platform) with OpenGL4. In theory they are also compatible with OpenGL3.

### DIFF
--- a/platform/common/cog-gl-utils.c
+++ b/platform/common/cog-gl-utils.c
@@ -151,6 +151,43 @@ cog_gl_renderer_initialize(CogGLRenderer *self, GError **error)
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
     glBindTexture(GL_TEXTURE_2D, 0);
 
+    /* Create vertex buffer */
+    if (epoxy_is_desktop_gl() || epoxy_gl_version() >= 30) {
+        glGenVertexArrays(1, &self->vao);
+        glBindVertexArray(self->vao);
+    } else {
+        self->vao = 0;
+    }
+
+    /* clang-format off */
+    static const GLfloat vertices[] = {
+        /* position */
+        -1.0f,  1.0f, 1.0f,  1.0f,
+        -1.0f, -1.0f, 1.0f, -1.0f,
+        /* texture */
+        /* COG_GL_RENDERER_ROTATION_0 */
+        0.0f, 0.0f, 1.0f, 0.0f,
+        0.0f, 1.0f, 1.0f, 1.0f,
+        /* COG_GL_RENDERER_ROTATION_90 */
+        1.0f, 0.0f, 1.0f, 1.0f,
+        0.0f, 0.0f, 0.0f, 1.0f,
+        /* COG_GL_RENDERER_ROTATION_180 */
+        1.0f, 1.0f, 0.0f, 1.0f,
+        1.0f, 0.0f, 0.0f, 0.0f,
+        /* COG_GL_RENDERER_ROTATION_270 */
+        0.0f, 1.0f, 0.0f, 0.0f,
+        1.0f, 1.0f, 1.0f, 0.0f,
+    };
+    /* clang-format on */
+
+    glGenBuffers(1, &self->buffer_vertex);
+    glBindBuffer(GL_ARRAY_BUFFER, self->buffer_vertex);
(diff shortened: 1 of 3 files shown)


zdobersek:
`glGetStringi` and VAO won't function when running this module on top of GLES2. I don't know if it was supported before, but it for sure won't be after.

zdobersek:
No, if you call `glGetString` you'll  get a call into `glGetString`, what most likely changed is Mesa now rejecting GL_EXTENSIONS for that entrypoint, as per standard.

aperezdc:
> No, if you call `glGetString` you'll get a call into `glGetString`, what most likely changed is Mesa now rejecting GL_EXTENSIONS for that entrypoint, as per standard.

Right, I was thinking of functions like `epoxy_has_gl_extension()` or `epoxy_has_egl_extension()` but you are right that there is no generic function to get the string with the list of extensions. In this case we only fetch the extensions to show a debug message, so we could as well remove it; and if we want to keep it, then we can choose what to call depending on `if (epoxy_is_desktop_gl() && epoxy_gl_version() >= 40)` 🤔 

Thoughs?

The issue with VAOs not working on GLES is more pressing, though. Is there any reason why the existing code didn't work with desktop OpenGL 4.x, or was that trying some cleanup?

zdobersek:
> > No, if you call `glGetString` you'll get a call into `glGetString`, what most likely changed is Mesa now rejecting GL_EXTENSIONS for that entrypoint, as per standard.
> 
> Right, I was thinking of functions like `epoxy_has_gl_extension()` or `epoxy_has_egl_extension()` but you are right that there is no generic function to get the string with the list of extensions. In this case we only fetch the extensions to show a debug message, so we could as well remove it; and if we want to keep it, then we can choose what to call depending on `if (epoxy_is_desktop_gl() && epoxy_gl_version() >= 40)` thinking
> 
> Thoughs?
> 

You also have to cover GLES3.

> @ceyusa The issue with VAOs not working on GLES is more pressing, though. Is there any reason why the existing code didn't work with desktop OpenGL 4.x, or was that trying some cleanup?

On GLES2.

aperezdc:
Can't we tell EGL to create an OpenGL 3.x context instead of having to deal with an OpenGL 4.x context? Or is the problem that GTK crated the context itself? Cannot we use [gtk_gl_area_set_required_version()](https://docs.gtk.org/gtk4/method.GLArea.set_required_version.html) and/or [gtk_gl_area_set_use_es()](https://docs.gtk.org/gtk4/method.GLArea.set_use_es.html) to have some control on how the GL context will be initialized so we can avoid having to maintain two code paths and arrange to choose one at runtime?

ceyusa:
> Can't we tell EGL to create an OpenGL 3.x context instead of having to deal with an OpenGL 4.x context? Or is the problem that GTK crated the context itself? Cannot we use [gtk_gl_area_set_required_version()](https://docs.gtk.org/gtk4/method.GLArea.set_required_version.html) and/or [gtk_gl_area_set_use_es()](https://docs.gtk.org/gtk4/method.GLArea.set_use_es.html) to have some control on how the GL context will be initialized so we can avoid having to maintain two code paths and arrange to choose one at runtime?

That's an interesting approach to try. Next week are my vacations. I can try in two weeks.

ceyusa:
> Can't we tell EGL to create an OpenGL 3.x context instead of having to deal with an OpenGL 4.x context? Or is the problem that GTK crated the context itself? Cannot we use [gtk_gl_area_set_required_version()](https://docs.gtk.org/gtk4/method.GLArea.set_required_version.html) and/or [gtk_gl_area_set_use_es()](https://docs.gtk.org/gtk4/method.GLArea.set_use_es.html) to have some control on how the GL context will be initialized so we can avoid having to maintain two code paths and arrange to choose one at runtime?

I've just tried them and it didn't fix the issue, but still,  GL_VERSION still logs `4.6 (Core Profile) Mesa 21.3.8 (git-813ee839be)`

My changeset is

```
     window->gl_drawing_area = gtk_gl_area_new();
+    gtk_gl_area_set_required_version(GTK_GL_AREA(window->gl_drawing_area), 3, 2);
+    gtk_gl_area_set_use_es(GTK_GL_AREA(window->gl_drawing_area), TRUE);
```

ceyusa:
This new version add checks to validate if the new API usage is valid for running GL version or if GLES 1/2 is used.

philn:
I don't feel confident to review this but I've tested it and it works, from what I can see anyway. Can we merge this soon please?

philn (approved):

PR #431, opened 2022-05-02
x11: Fix glitches in non-composited window managers

aperezdc:
Handle X11 expose events, which was missing. Typically composited environments (running `xcompmgr`, a composited desktop like GNOME Shell, or similar) never produces expose events: the compositor will keep around a buffer with the latest contents of the X11 windows, in order to perform composition, which means that it does not ever need to request application to repaint windows. Given that composited environment are common these days, that is likely the reason why this issue has not been
noticed earlier.

On the other hand, with more “traditional”, non-composited window managers will happily discard pixmaps for obscured windows (in order to save memory, for example) and rely on the fact that X11 applications are expected to repaint themselves when requested via expose events. In order to be able to do this, the X11 platform plug-in now keeps the latest frame exported by WPEBackend-fdo, and performs a repaint when an expose event indicates that the window (or part of it) is visible.

The existing `needs_initial_repaint` flag is renamed as `needs_repaint` and reused also for expose events. Repaints are scheduled by sending a notice event, similarly to how frame completions are postponed queued to the event loop.

This is a follow-up to #398

### DIFF
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -98,8 +98,9 @@ struct CogX11Window {
     struct {
         xcb_window_t window;
 
-        bool needs_initial_paint;
+        bool needs_repaint;
         bool needs_frame_completion;
+
         unsigned width;
         unsigned height;
     } xcb;
@@ -119,10 +120,13 @@ struct CogX11Window {
 static struct CogX11Display *s_display = NULL;
 static struct CogX11Window *s_window = NULL;
 
-
 static void
-xcb_schedule_repaint (void)
+xcb_schedule_notice(void)
 {
+    /* There is a notice message already scheduled, do not queue another. */
+    if (s_window->xcb.needs_repaint || s_window->xcb.needs_frame_completion)
+        return;
+
     xcb_client_message_event_t client_message = {
         .response_type = XCB_CLIENT_MESSAGE,
         .format = 32,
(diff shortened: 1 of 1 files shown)


platform/x11/cog-platform-x11.c
```diff
@@ -380,15 +390,7 @@ xcb_process_events (void)
...
-                if (s_window->xcb.needs_initial_paint) {
-                    xcb_initial_paint ();
-                    s_window->xcb.needs_initial_paint = false;
-                }
-
-                if (s_window->xcb.needs_frame_completion) {
-                    xcb_frame_completion ();
-                    s_window->xcb.needs_frame_completion = false;
-                }
+                xcb_handle_notice();
```
Cwiiis, line 393: It feels just slightly odd to me that there's an `xcb_handle_notice` rather than just putting the code here like the message handlers above. It's not so big that breaking it out helps readability and it's not going to get reused anywhere else(?)

platform/x11/cog-platform-x11.c
```diff
@@ -432,11 +434,19 @@ xcb_process_events (void)
         case XCB_VISIBILITY_NOTIFY:
             xcb_handle_visibility_event((xcb_visibility_notify_event_t *) event);
             break;
+        case XCB_EXPOSE: {
```
Cwiiis, line 437: Is the style incorrect on the previous case events, or should the brace go on the next line here?

Cwiiis (approved):
Took me a little bit to understand the flow here, but looks good - couple of comments, neither of which are important to address before landing, I think.

platform/x11/cog-platform-x11.c
aperezdc: I'm not 100% sure because I wrote this a while ago, but probably I thought the function would end up being longer, but then it didn't. Let's move back the code inline into the `switch`.

platform/x11/cog-platform-x11.c
aperezdc: This is the correct way, the previous ones are wrong; but the style checker runs only on the changed parts of the code, that's why the other ones are not reported as being wrong.

Issue #429, opened 2022-05-01
core: Add a “retry” action to built-in error pages


aperezdc:
Can the error page be changed to include a "try again" button at least?

_Originally posted by @EinarArnason in https://github.com/Igalia/cog/issues/45#issuecomment-1106405147_

aperezdc:
This can be done by changing the template HTML in `core/cog-webkit-utils.c` 📜

PR #433, opened 2022-05-03
#429 add try again button on error page

EinarArnason:


.github/workflows/ci-native.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           python-version: 3.8
       - name: Install Python Packages
         run: |
-          pip install --upgrade pip gi-docgen==2021.6 meson==0.49
+          pip install --upgrade pip gi-docgen==2021.6 meson==0.62.1
```
aperezdc, line 35: It would be better to use `==0.55` here, to make sure we don't accidentally introduce use of newer Meson features.

The reason why we changed libwpe to 0.55 (and not newer) is that people integrating WPE sometimes have slightly old toolchains and/or build systems, and versions newer than Meson 0.55 do not add any feature that we would need at the moment 🤔

core/cog-webkit-utils.c
```diff
@@ -69,61 +60,53 @@ load_error_page (WebKitWebView *web_view,
  * Returns: Whether other signal handlers should be skipped.
  */
 gboolean
-cog_handle_web_view_load_failed (WebKitWebView  *web_view,
```
aperezdc, line 72: It looks like you ran `clang-format` on the whole file, is that so? We do not really encourage doing that and instead we prefer to slowly change only the parts of the code touched by changes towards the coding style checked by `clang-format`.

To do that, you would commit only the changes needed to improve the error page HTML and then run this to apply `clang-format` only on your changes:

```sh
data/check-style origin/master | git apply -p0
```

aperezdc (changes requested):
Thanks @EinarArnason! Changes look good, but we would need to make sure the patch only reformats the part that you needed to modify. I have left a note with a tip on how to do that. Please take a look as well at the comment about the Meson version.

I look forward to merging an updated patch 😄

core/cog-webkit-utils.c
EinarArnason: Got it :). I've addressed these issues and updated the PR.

aperezdc:
@EinarArnason Much nicer having a smaller set of changes, thanks! The style checker is still complaining, could you take a look? If it gets complicated, let me know and I could merge your patch by hand.

aperezdc (approved):

Issue #401, opened 2022-02-03
drm: Both modesetting and GLES renderers broken


aperezdc:
As of bbf64f0eb8a1e01f4a4dcd9d1ff22631d5ceac34:

- `cog -P drm -O renderer=modeset` results in failed page flips.
- `cog -P drm -O renderer=gles` results in a blank screen.

I had 0.12.0 installed in my system, which worked, so I ran a bisect as follows:

```sh
git bisect start master $(git merge-base 0.12.0 master)
```

This did the bisection starting at the branch point for the 0.12 stable branch (where things still worked) and `master`. After a few tries the commit where the fallout started turned out to be 91369c0c1c1a4b83c941f5f90783eea78fba78c1—even before merging the GLES render in #371, but amusingly I am quite sure I had the GLES renderer stuff working, so I don't know what's going on anymore. But clearly the situation needs to be fixed.

PR #435, opened 2022-05-09
launcher: set device scale factor before cog_shell_startup

q66:
A valid scale factor needs to be set before shell startup as
otherwise it breaks the drm platform.

Fixes https://github.com/Igalia/cog/issues/401


q66:
it has to be set to something for the drm platform to correctly come up (it used to be, but then got changed during refactoring)

aperezdc:
Potentially related: https://github.com/WebPlatformForEmbedded/libwpe/issues/89

PR #439, opened 2022-05-16
Add auto retry on interval

EinarArnason:
Retries the failing URL on 5 second interval

### DIFF
--- a/core/cog-webkit-utils.c
+++ b/core/cog-webkit-utils.c
@@ -19,11 +19,13 @@ static const char error_message_template[] =
     "        margin-left: 0.75em; margin-top: 0 }\n"
     ".try-again { text-align: center; font-size: 1em; \n"
     "             height: 100%; margin: 1em; }\n"
-    "</style></head><body>\n"
+    "</style>\n"
+    "<script>\nfunction retry() { window.location.href = '%s' }\n"
+    "setTimeout(retry, 5000);\n</script></head><body>\n"
     "  <h3>%s</h3>\n"
     "  <p class='uri'>%s</p>\n"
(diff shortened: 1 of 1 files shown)


core/cog-webkit-utils.c
```diff
@@ -19,11 +19,13 @@ static const char error_message_template[] =
     "        margin-left: 0.75em; margin-top: 0 }\n"
     ".try-again { text-align: center; font-size: 1em; \n"
     "             height: 100%; margin: 1em; }\n"
-    "</style></head><body>\n"
+    "</style>\n"
+    "<script>\nfunction retry() { window.location.href = '%s' }\n"
+    "setTimeout(retry, 5000);\n</script></head><body>\n"
```
aperezdc: This make the “retry every five seconds” behaviour unconditional, and I can imagine there may be cases in which this is not desired. It would be better if we could have an option to configure (or disable) the retry interval.

aperezdc (changes requested):
@EinarArnason Hi there, thanks for the the patch! I am definitely not against the feature, but I would rather not change the existing behaviour by default. Would you consider making the retry interval a configurable option?

core/cog-webkit-utils.c
EinarArnason: You're right, and option would be best. Should it be disabled by default?

aperezdc: Yes, disabled by default would keep the current behaviour by default, to avoid breaking people's setups. Thanks for considering reworking the patch 😸

limbo127:
hello, can you commit this patch ?
Nicolas

PR #440, opened 2022-05-20
Fix missing headers

dpino:
WPE Non-unified build bot was failing due to missing headers in Cog.

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -18,12 +18,14 @@
 #include <gbm.h>
 #include <libinput.h>
 #include <libudev.h>
+#include <stdio.h>
 #include <string.h>
 #include <wayland-server.h>
 #include <wpe/fdo-egl.h>
 #include <wpe/fdo.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
+#include <xkbcommon/xkbcommon.h>
 
 #include <epoxy/egl.h>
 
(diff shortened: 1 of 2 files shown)


platform/drm/cog-platform-drm.c
```diff
@@ -18,12 +18,14 @@
...
 #include <libinput.h>
 #include <libudev.h>
+#include <stdio.h>
 #include <string.h>
 #include <wayland-server.h>
 #include <wpe/fdo-egl.h>
 #include <wpe/fdo.h>
 #include <xf86drm.h>
 #include <xf86drmMode.h>
+#include <xkbcommon/xkbcommon.h>
```
aperezdc, line 28: This should not be needed. Note that:

- The source file includes `<wpe/fdo.h>`
- `<wpe/fdo.h>` includes `<wpe/wpe.h>`
- `<wpe/wpe.h>` includes `"input-xkb.h"`
- `"input-xkb.h"` includes `<xkbcommon/xkbcommon.h>`

Most likely the build issue you experienced was caused by an old build directory created before https://github.com/WebPlatformForEmbedded/libwpe/pull/94 was merged: that PR makes inclusion of the xkbcommon headers conditional on the `WPE_ENABLE_XKB` macro, which will get defined automatically by `pkg-config` if libwpe was built with xkbcommon support (in that case `-DWPE_ENABLE_XKB=1` gets picked from the `Cflags` field of the `wpe-1.0.pc` file).

What we could add is a check to the Cog `CMakeLists.txt` that verifies whether libwpe is older than version 1.13.2 (which always have xkbcommon support) _or_ that the `WPE_ENABLE_XKB` macro gets defined.

aperezdc (commented):
I think this is wrong, please read the explanation below.

dpino:
This change is indeed no needed.

PR #443, opened 2022-05-20
Add target refresh rate reporting

Cwiiis:
Add target refresh rate reporting to drm, gtk4 and wayland backends. The
latter two should follow refresh rate/monitor changes, drm doesn't as
we're responsible for setting the mode and only ever set it once.


aperezdc (approved):
Well, I am not a fan of `clang-format` making you add some churn here, but that's what we get for preferring to converge towards the code style enforced by it progressively 😅 

The functionality in the patch is great to have, thanks! 👌🏼

PR #445, opened 2022-05-24
x11: handle upper case letter, caps lock, num lock, control and alt

### LINKED ISSUE #391: x11: Cannot type capital letters in input fields
Steps to reproduce:

1. Run `cog -P x11 'data:text/html,<input>'`
2. Try entering text with capital letters in the input field.

Most likely this is caused by the code missing some handling for modifiers keys when inspecting X11 key events, or the shift modifier being handled incorrectly.

ubercomp:
Fixes #391

Steps to test:
1. Run COG_MODULEDIR=$(pwd)/modules ./cog -P x11 'data:text/html,<input>'
2. Enter various key combinations (with shift, control, caps lock, num lock, etc), and verify that they work.

What was missing was updating the xkb state on each key press.
That needs to be done before the call to xkb_state_key_get_one_sym
as xkb reads the state in order to determine which keysym it should
output.

The function is more complex that in need to be because XCB, X11,
and WPE all use different masks for the modifiers.

Future improvements to this code might include:
* unifying xcb_handle_key_press and xcb_handle_key_release
* handle xkb keyboard map changes
* handle AltGr

### DIFF
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -75,12 +75,15 @@ struct CogX11Display {
     } xcb;
 
     struct {
-        int32_t device_id;
+        int32_t             device_id;
         struct xkb_context *context;
-        struct xkb_keymap *keymap;
-        struct xkb_state *state;
-
-        uint8_t modifiers;
+        struct xkb_keymap  *keymap;
+        struct xkb_state   *state;
+        xkb_mod_mask_t      shift;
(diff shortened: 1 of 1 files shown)


aperezdc:
> Future improvements to this code might include:

I have created https://github.com/Igalia/cog/issues/446 to not forget about these.

ubercomp:
@aperezdc , thanks for the kind words and, most of all, thank you and all Igalians for your work on all web technologies.

PR #468, opened 2022-07-14
Remove CMake build system

aperezdc:
In #465 we added a new build system based on Meson. This picks where that PR was left off, and additionally does the following:

- [x] Remove the CMake build system.
- [x] Changes CMake references in the documentation over to Meson.
- [x] Changes the CI jobs to use Meson.


README.md
```diff
@@ -46,8 +45,9 @@ Using Cog
 ---------
 
 **Compiling** Cog follows the usual procedure for projects which use
-[CMake](http://cmake.org): `cmake . && make` should get you started, if your
-system has the needed [dependencies](#dependencies) installed.
+[Meson](https://mesonbuild.com): `meson build && ninja -C build` should get
```
eli-schwartz, line 48: Technically, it is `meson setup builddir/`, not `meson builddir/` -- if the first argument is not recognized as a valid subcommand, we insert "setup" before it.

This runs the risk that the build directory name you use clashes with some meson subcommand added in the future -- this is actually the reason why we didn't use "build" as a subcommand name, funnily enough.

We may or may not, in the future, issue deprecation warnings for implicit "setup", who knows.

aperezdc: I have updated the instructions to use `meson setup …`, thanks for the tip 👌🏼

ceyusa:
given that the CI passed, it looks good to me :)

PR #469, opened 2022-07-20
Clean up the code of some warnings produced by Clang 14

aperezdc:
This includes two commits:

- One removing unneeded forward-declarations of `struct`s which get defined anyway by GLib/GObject/WebKit headers. 
- Removing an unused variable.


ceyusa:
:clap: :clap: :clap:  I was about to tackle those warnings (which I also saw with clang 13 in debian)

lgtm

ceyusa (approved):

PR #477, opened 2022-08-24
gamepad: Select backend at launch time.

ceyusa:
Adds the command line option `--gamepad` where the user can select, at launch
time, the backend of the gamepad to use. By default `disable` is available. If
cog is compiled with libmanette as dependecy, also `manette` backend would be
available.

The initialization of the gamepad is split in two stages:

1. When the launcher selects the backend.
2. When the platform sets the `get_view_backend()` callback.


core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+}
+
+void
+cog_gamepad_setup(GamepadProviderGetViewBackend *gamepad_get_view)
+{
+    if (selected)
+        return;
+
+    if (!backend)
+        return;
```
aperezdc, line 68: Doing as suggested, the `s_backend` pointer will always be valid, and we only need to check that this function is called only once.

```suggestion
    static bool initialized = false;
    g_return_if_fail(!initialized);

    initialized = true;
```

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+
+    for (int i = 0; i < G_N_ELEMENTS(s_gamepad_backends); i++) {
+        if (g_strcmp0(s_gamepad_backends[i].name, backend) == 0)
+            selected = &s_gamepad_backends[i];
+    }
+
+    g_assert(selected);
+    g_debug("gamepad setup: %s", selected->name);
```
aperezdc, line 76: No need for this loop or the check, the `s_backend` pointer will always be valid.

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+    if (g_strcmp0(selected->name, "disabled") == 0)
+        return;
```
aperezdc, line 78: No need for a string comparison here:

```suggestion
    if (!s_backend->provider)
        return;
```

launcher/cog-launcher.c
```diff
@@ -987,6 +988,16 @@ cog_launcher_add_web_settings_option_entries(CogLauncher *launcher)
     g_application_add_option_group(G_APPLICATION(launcher), g_steal_pointer(&option_group));
 }
 
+static gboolean
+option_entry_parse_gamepad(const char *option_name, const char *value, void *data, GError **error)
+{
+    if (!cog_gamepad_parse_backend(value, error))
+        return FALSE;
+
+    s_options.gamepad = g_strdup(value);
```
aperezdc, line 997: This could call `cog_gamepad_set_backend()` directly here, no need to save the passed value.

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+} s_gamepad_backends[] = {
+#if COG_ENABLE_GAMEPAD_MANETTE
+    {
+        "manette",
+        &manette_provider_interface,
+        &manette_device_interface,
+    },
+#endif
+    {
+        "disabled",
```
aperezdc, line 29: I like `"none"` better than `"disabled"`.

aperezdc (changes requested):
Thanks for this, it's a neat improvement. I have made some comments around the code which would be good to address before landing.

core/cog-gamepad-manette.c
```diff
@@ -292,30 +292,15 @@ gamepad_get_id(void *data)
     return manette_device_get_name(gamepad->device);
 }
 
-static struct wpe_gamepad_provider_interface provider_interface = {
+const struct wpe_gamepad_provider_interface manette_provider_interface = {
```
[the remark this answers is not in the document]
ceyusa: done

core/cog-gamepad-manette.c
```diff
@@ -292,30 +292,15 @@ gamepad_get_id(void *data)
...
     .create = provider_create,
     .destroy = provider_destroy,
     .start = provider_start,
     .stop = provider_stop,
 };
 
-static const struct wpe_gamepad_interface device_interface = {
+const struct wpe_gamepad_interface manette_device_interface = {
```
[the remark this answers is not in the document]
ceyusa: done

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+        NULL,
+    },
+};
+
+/* selected gamepad backend name */
+static const char *backend;
+/* selected gamepad backend interfaces */
+static const struct CogGamepadBackend *selected;
+/* selected gamepad backend provider interfaces */
+static struct wpe_gamepad_provider_interface provider_interface;
```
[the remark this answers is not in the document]
ceyusa: I kept the global `static struct wpe_gamepad_provider_interface provider_interface;` otherwise the compiler complains with: `error: initializaer element is not a compile-time constant'

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+
+void
+cog_gamepad_set_backend(const char *name)
+{
+    if (backend)
```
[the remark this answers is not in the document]
ceyusa: done

core/cog-gamepad.c
```diff
@@ -0,0 +1,104 @@
...
+        return;
+
+    if (!name) {
+        backend = s_gamepad_backends[0].name;
+        goto bail;
+    }
+
+    for (int i = 0; i < G_N_ELEMENTS(s_gamepad_backends); i++)
+        if (g_strcmp0(name, s_gamepad_backends[i].name) == 0)
+            backend = s_gamepad_backends[i].name;
```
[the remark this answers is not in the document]
ceyusa: done :)

core/cog-gamepad.c
ceyusa: done

core/cog-gamepad.c
ceyusa: done

launcher/cog-launcher.c
ceyusa: done

ceyusa: I kept both methods, for users, perhaps they don't want to parse arguments, but only set the gamepad backend.

core/cog-gamepad.c
ceyusa: done

aperezdc (approved):

PR #485, opened 2022-09-02
cmake: Write correct wpe-webkit requirement in .pc file

aperezdc:
Arrange for the generated `cogcore.pc` file to include the version of the `wpe-webkit-1.x` dependency, instead of hardcoding `1.0`; this solves issues when Cog was configured against `wpe-webkit-1.1` where the `.pc` file would have the wrong version.

The switch from `Requires.private` to plain `Requires` is needed to let `pkg-config` add the WPE WebKit library in the output for `pkg-config cogcore --libs`, otherwise users need to specify
both `cogcore` and `wpe-webkit-1.x` as dependencies, when using `cogcore` should be enough.

### DIFF
--- a/core/cogcore.pc.in
+++ b/core/cogcore.pc.in
@@ -6,6 +6,6 @@ Name: cogcore
 Description: Cog Core - WPE WebKit base launcher
 Version: @PROJECT_VERSION@
 
-Requires.private: wpe-webkit-1.0
+Requires: @WPEWEBKIT_PC_NAME@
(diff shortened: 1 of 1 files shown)


aperezdc:
Note that this PR targets the `cog-0.14` branch, which still uses CMake. In ToT we use Meson and a different change is needed, see https://github.com/Igalia/cog/pull/486.

psaavedra (approved):

PR #497, opened 2022-09-26
meson: Allow fallback to wpe-webkit-1.1

aperezdc:
Allow using `wpe-webkit-1.1` in place of `wpe-webkit-2.0`, to ease building on systems which provide the former.

Commit be29794 changeg the dependency when using libsoup3 from `wpe-webkit-1.1` to `wpe-webkit-2.0`, but there are at the moment no tangible differences between both.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -79,10 +79,22 @@ with_x11_platform = platform_plugins.contains('x11')
 
 with_programs = get_option('programs')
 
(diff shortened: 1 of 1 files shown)


aperezdc:
cc/ @obyknovenius @lauromoura

psaavedra (approved):

PR #522, opened 2022-12-16
webkit-utils: don't use WebKit plugins API

carlosgcampos:
The API is deprecated and the functionality removed from WebKit.

### DIFF
--- a/core/cog-webkit-utils.c
+++ b/core/cog-webkit-utils.c
@@ -72,13 +72,6 @@ cog_handle_web_view_load_failed (WebKitWebView  *web_view,
                                  GError         *error,
                                  void           *userdata)
 {
-    // If the resource is going to be shown by a plug-in (or a
-    // media engine) just return FALSE and let WebKit handle it.
-    if (g_error_matches (error,
-                         WEBKIT_PLUGIN_ERROR,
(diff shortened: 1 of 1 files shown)


TingPing (approved):

core/cog-webkit-utils.c
```diff
@@ -72,13 +72,6 @@ cog_handle_web_view_load_failed (WebKitWebView  *web_view,
                                  GError         *error,
                                  void           *userdata)
 {
-    // If the resource is going to be shown by a plug-in (or a
-    // media engine) just return FALSE and let WebKit handle it.
-    if (g_error_matches (error,
-                         WEBKIT_PLUGIN_ERROR,
-                         WEBKIT_PLUGIN_ERROR_WILL_HANDLE_LOAD))
-        return FALSE;
-
```
aperezdc, line 81: Couldn't we use conditional compilation here to enable this part of the code only when building against the old API?

aperezdc (changes requested):

core/cog-webkit-utils.c
aperezdc: We know that if API<2.0, then this code needs to be built.

carlosgcampos: It's not just deprecated API, the functionality was removed, so it's dead code.

PR #524, opened 2023-01-06
Drop support for wpe-webkit-1.1, build only for libsoup 3 by default, and don't try to enable sandbox when using new WebKit API

mcatanzaro:
This will allow building cog after https://github.com/WebKit/WebKit/pull/5944/

### DIFF
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -53,7 +53,9 @@ static struct {
     } on_failure;
     char    *web_extensions_dir;
     gboolean ignore_tls_errors;
+#if !COG_USE_WPE2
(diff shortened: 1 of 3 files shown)


mcatanzaro:
Ah, lovely, the CI is not ready for it....

TingPing:
Merging into `mcatanzaro/next` for now so upstream webkit can use it. Getting it ready for master will come later.

TingPing:
And just an FYI, Ubuntu 22.04 contains `libsoup-3.0-0`, so that's probably the easiest way to fix CI. Not sure if targeting 20.04 is an important requirement.

PR #529, opened 2023-01-16
mcatanzaro/next: CI: Build with libsoup2 for now

TingPing:


TingPing:
This fixed native builds but not cross builds. It's an improvement.

TingPing:
The Igalia hosted ARM toolchains would have to be updated there. Either they target wpe 1.0 or the new API.

PR #530, opened 2023-01-16
drm: Document supported environment variables

aperezdc:
Add the missing documentation for the environment variables honored by the DRM/KMS platform plug-in.

### DIFF
--- a/docs/platform-drm.md
+++ b/docs/platform-drm.md
@@ -62,6 +62,30 @@ The `rotation` parameter indicates the initial [output
 rotation](#output-rotation) applied.
 
 
+## Environment Variables
+
+The following environment variables can be set to further tweak how the
+DRM plug-in operates:
+
+| Variable | Type | Default |
+|:---------|:-----|--------:|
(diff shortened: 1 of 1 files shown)


ceyusa (approved):
lgtm

ceyusa:
As a non-actionable comment, I'm a bit wary about environment variables and it potential abuse (such as changing values in run time). All the configuration envvars are also parsed in the configuration file? IMO these documented envvars make more sense in a configuration file rather than envvars.

aperezdc:
> As a non-actionable comment, I'm a bit wary about environment variables and it potential abuse (such as changing values in run time). All the configuration envvars are also parsed in the configuration file? IMO these documented envvars make more sense in a configuration file rather than envvars. I have filed #531 to try and reduce their usage.

While in most (even all?) cases in Cog we only read environment variables once at initialization (which is okay), I do agree that changing them to be parameters and/or config file entries is desirable in most cases. That's why platform plug-in parameters were added, and then the config file... Also I made #363 to explore how to unify configuration parameters as object properties as a first step towards abandoning most environment variables, but there has been no comments there whatsoever 🤷🏼

PR #533, opened 2023-01-25
Port to use new WebKit network session API

carlosgcampos:


launcher/cog-launcher.c
```diff
@@ -388,6 +398,13 @@ cog_launcher_startup(GApplication *application)
...
     }
 
+#if COG_USE_WPE2
+    if (!self->automated) {
+        webkit_network_session_set_tls_errors_policy(self->network_session,
+                                                     s_options.ignore_tls_errors ? WEBKIT_TLS_ERRORS_POLICY_IGNORE
+                                                                                 : WEBKIT_TLS_ERRORS_POLICY_FAIL);
+    }
+#else
 #if WEBKIT_CHECK_VERSION(2, 32, 0)
```
aperezdc, line 408: I would replace both lines with an `#elif`:

```suggestion
#elif WEBKIT_CHECK_VERSION(2, 32, 0)
```

launcher/cog-launcher.c
```diff
@@ -1251,9 +1318,7 @@ cog_launcher_handle_local_options(GApplication *application, GVariantDict *optio
     }
 
     if (s_options.filter_path) {
-        WebKitWebsiteDataManager *data_manager = cog_launcher_get_web_data_manager(launcher);
-        g_autofree char          *filters_path =
-            g_build_filename(webkit_website_data_manager_get_base_cache_directory(data_manager), "filters", NULL);
+        g_autofree char          *filters_path = g_build_filename(g_get_user_cache_dir(), g_get_prgname(), "filters", NULL);
```
aperezdc, line 1321: It makes me a bit sad that we no longer have a “base directories for all cache/user data” that we can query WebKit for.

launcher/cog-launcher.c
```diff
@@ -1271,11 +1336,22 @@ cog_launcher_handle_local_options(GApplication *application, GVariantDict *optio
...
+            return EXIT_FAILURE;
+        }
+
         g_autoptr(WebKitNetworkProxySettings) webkit_proxy_settings =
             webkit_network_proxy_settings_new(s_options.proxy, (const gchar *const *) s_options.ignore_hosts);
-        webkit_website_data_manager_set_network_proxy_settings(data_manager, WEBKIT_NETWORK_PROXY_MODE_CUSTOM,
+#if COG_USE_WPE2
+        webkit_network_session_set_proxy_settings(launcher->network_session,
+                                                  WEBKIT_NETWORK_PROXY_MODE_CUSTOM,
+                                                  webkit_proxy_settings);
```
aperezdc, line 1349: Having some things like the proxy settings in something called “network session” does indeed make sense and the new API feels nicer in general 👍🏼

aperezdc (approved):
I left a couple of comments, but nothing that would be a showstopper.

launcher/cog-launcher.c
carlosgcampos: Well, there's `webkit_website_data_manager_get_base_data_directory()` and `webkit_website_data_manager_get_base_cache_directory()`, but they return what you set, so when set to `NULL` to use the default you get `NULL`. I don't know why we designed it this way, I guess the idea is that either you set it yourself (and then you know it) or use the default (which is known too, but we can't change it in the future if apps rely on it). I thought about changing that in the new API, since you don't really set the values directly to the website data manager. I really prefer if the getters return the paths actually used, so only `NULL` in case of ephemeral session. What do you think @mcatanzaro ?

mcatanzaro: Agree.

PR #535, opened 2023-02-06
wl: fix hardware keycodes

jameshilliard:
The `hardware_key_code` param for `wpe_input_keyboard_event` is not utf32 but rather the raw keycode as it is not intended to change based on keymaps.

Raw hardware keycodes are translated by [keyCodeForHardwareKeyCode](https://github.com/WebKit/WebKit/blob/wpewebkit-2.39.5/Source/WebCore/platform/libwpe/PlatformKeyboardEventLibWPE.cpp#L446-L765).


aperezdc (approved):
Thanks @jameshilliard, nice catch!

aperezdc:
Cherry-picked as commit 0716c4cb90654366a70d00ecab14637da6024ed9 in the `cog-0.16` branch.

Issue #537, opened 2023-02-16
WL MAXIMIZE do not work on MAC M1 QEMU


munoz0raul:
Hi All, 

I'm using QEMU and COG and on my Intel machine, ruing QEMU ARM64 everything works fine. But on a college machine with the MAC M1 (ARM), running QEMU ARM64, maximize does not work, and when I remove maximize and set the QEMU resolution the cog is launched with an offset. All other Wayland Apps I have work fine in both environments.

Here are some examples:

```
COG_PLATFORM_WL_VIEW_MAXIMIZE="0"
COG_PLATFORM_WL_VIEW_WIDTH="1280"
COG_PLATFORM_WL_VIEW_HEIGHT:="768"
```

Intel x86 (Linux) running QEMU ARM: (Works fine)
![image](https://user-images.githubusercontent.com/7702208/219238837-099cee93-d5fe-497a-a9bc-70b95d0404cd.png)

MAC M1 ARM64 (MAC OS) running QEMU ARM: (Offset)
![image](https://user-images.githubusercontent.com/7702208/219239286-48776ec9-e5a0-4a45-884e-23a40418e864.png)

When trying fullscreen:

`COG_PLATFORM_WL_VIEW_MAXIMIZE="1"
`

Intel x86 (Linux) running QEMU ARM: (Works fine)
![image](https://user-images.githubusercontent.com/7702208/219240135-b74f9b86-2147-466e-86c7-cc798aa774a2.png)

MAC M1 ARM64 (MAC OS) running QEMU ARM: (Error)

```
(cog:8): GLib-GIO-WARNING **: 00:36:50.081: Your application does not implement g_application_activate() and has no handlers connected to the 'activate' signal.  It should do one of these.
Cog-Core-Message: 00:36:50.154: <file:///index.html> Load started.
xdg_wm_base@11: error 4: xdg_surface geometry (1024 x 768) does not match the configured maximized state (1280 x 768)
Cog-Core-Message: 00:36:50.157: <file:///index.html> Loading...
Cog-Core-Message: 00:36:50.162: <file:///index.html> Loaded successfully.
```

Funny that FULLSCREEN works fine for both environments.

I'm wondering if there is something implemented for the FULLSCREEN that was not added to the MAXIMIZE.

I'm using COG upstream built with Yocto.

munoz0raul:
I found the problem. When using QEMU on mac, it doesn't have EGL, so Weston starts without EGL:

`EGL Wayland extension: no`

[...]

Issue #489, opened 2022-09-09
xdg_surface geometry is larger than the configured fullscreen state (dmabuf)


jameshilliard:
This error appears to be nearly identical to #419(which was fixed by #421) except that it happens when using `dmabuf` instead of `shm` buffer codepaths. It appears that `zwp_linux_buffer_params_v1@15.create_immed(new id wl_buffer@16, 1024, 768, 875713089, 0)` is called with the default width and height instead of the required full-screen width and height.

I'm seeing this error in `0.12.4` through `master` and the issue only happen under some configured screen sizes(when either the width or height of the default window exceed the fullscreen size):
```c
Started WPE launcher and webapp container.
ensure_extension_points: Extension points registered.
cog_modules_add_directory: Scanning '/usr/lib/cog/modules'
com.igalia.Cog.Platform:
  wl - 500/CogWlPlatform
  headless - 100/CogHeadlessPlatform
_g_io_module_get_default: Found default implementation local (GLocalVfs) for ‘gio-vfs’
platform_setup: Platform name: wl
[1951555.670]  -> wl_display@1.get_registry(new id wl_registry@2)
[1951555.683]  -> wl_display@1.sync(new id wl_callback@3)
[1951555.744] wl_display@1.delete_id(3)
[1951555.750] wl_registry@2.global(1, "wl_compositor", 4)
[1951555.760] wl_registry@2.global(2, "wl_subcompositor", 1)
[1951555.770] wl_registry@2.global(3, "wp_viewporter", 1)
[1951555.780] wl_registry@2.global(4, "zxdg_output_manager_v1", 2)
[1951555.791] wl_registry@2.global(5, "wp_presentation", 1)
[1951555.801] wl_registry@2.global(6, "zwp_relative_pointer_manager_v1", 1)
[1951555.811] wl_registry@2.global(7, "zwp_pointer_constraints_v1", 1)
[1951555.824] wl_registry@2.global(8, "zwp_input_timestamps_manager_v1", 1)
[1951555.834] wl_registry@2.global(9, "wl_data_device_manager", 3)
[1951555.843] wl_registry@2.global(10, "wl_shm", 1)
[1951555.853] wl_registry@2.global(11, "weston_touch_calibration", 1)
[1951555.863] wl_registry@2.global(12, "wl_drm", 2)
[1951555.874] wl_registry@2.global(13, "wl_seat", 7)
[1951555.884] wl_registry@2.global(14, "zwp_linux_dmabuf_v1", 4)
[...]

PR #539, opened 2023-02-23
wl: Do not attach frames with invalid size to the surface

aperezdc:
Skip frames which do not match the size of the output surface, in order to avoid a Wayland protocol error resulting from mismatched sizes. This can happen, for example, after requesting that the surface is maximized, while the view backend has already produced some frames, but the change of size has not been yet noted by WebKit. In these cases we can safely skip using the frames but they  still need to be ACK'd in order to let WebKit continue producing new ones, which eventually will arrive in the desired size.

Related issue #537


Bastian-Krause:
Maybe this also solves #489?

aperezdc:
> Maybe this also solves #489?

Yes, it looks like it could also fix that issue 🤔

carlosgcampos (approved):

PR #544, opened 2023-03-03
Adjust for web process extensions API renaming

mcatanzaro:


mcatanzaro:
> Should we also merge the contents of mcatanzaro/next into the master branch? I understand that the code there should build with with the old and new APIs, right?

Yes, I think so. Maybe there was some problem with CI preventing it from being merged into master? Certainly I agree this should be merged into master.

PR #545, opened 2023-03-07
wl: Set opaque region if possible also when windowed

aperezdc:
Configure an opaque region for the Wayland surface also when not fullscreened, as long as the web view background color is set to be opaque. This should slightly improve performance in compositors which take the opaque region hint into account also for non fullscreen surfaces.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -1605,6 +1605,8 @@ static const struct wl_buffer_listener dmabuf_buffer_listener = {
 static void
 on_export_wl_egl_image(void *data, struct wpe_fdo_egl_exported_image *image)
 {
+    CogWlPlatform *self = data;
+
     const uint32_t surface_pixel_width = wl_data.current_output->scale * win_data.width;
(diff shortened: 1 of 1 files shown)


aperezdc:
Note that there is a small issue: this assumes that the background color won't change after being set, which is true for the Cog launcher, but might not be for others. Unfortunately, we do not have a way of knowing when the background color has been changed: it's not a GObject property, so we cannot connect to the `GObject::notify` signal.

aperezdc:
> Note that there is a small issue: this assumes that the background color won't change after being set, which is true for the Cog launcher, but might not be for others. Unfortunately, we do not have a way of knowing when the background color has been changed: it's not a GObject property, so we cannot connect to the `GObject::notify` signal.

Reported: https://bugs.webkit.org/show_bug.cgi?id=253818

psaavedra (approved):
I've been used this patch on downstream for a long while. The change looks good and this new logic is more likely to what the user could expect: if transparency works in a windowed view then it should work also if the window is maximized.

PR #546, opened 2023-03-08
headless: Cleanup in preparation for multiple view support

aperezdc:
Tidy up the headless backend in preparation for adding multiple view support down the line. In particular, rename `struct platform_window` to `CogHeadlessView`, and embed an instance in the `CogPlatform` struct for now, removing the `win` global.

While at it, remove the `view_backend` member from `CogHeadlessView`, because once it is returned from the `get_view_backend` vfunc the ownership is transferred anyway, and it is not needed elsewhere.

Rename `setup_fdo_exportable` to `cog_headless_view_initialize`, and leave out the wpe-fdo initialization in `cog_headless_platform_setup`. Add a matching `cog_headless_view_finalize`, which takes care of disarming the tick timer. Destroying the view backend is now done using the `destroy_notify` callback from `WebKitWebViewBackend`, which is more correct.

### DIFF
--- a/platform/headless/cog-platform-headless.c
+++ b/platform/headless/cog-platform-headless.c
@@ -11,12 +11,23 @@
 #include <wpe/fdo.h>
 #include <wpe/unstable/fdo-shm.h>
 
+typedef struct {
+    unsigned                                tick_source;
+    bool                                    frame_ack_pending;
+    struct wpe_view_backend_exportable_fdo *exportable;
+} CogHeadlessView;
+
 struct _CogHeadlessPlatformClass {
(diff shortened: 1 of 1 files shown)


aperezdc:
Fix for the failed CI cross builds coming in #548

philn (commented):
Are the `g_debug()` calls supposed to remain here? Or were temporary?

aperezdc:
> Are the `g_debug()` calls supposed to remain here? Or were temporary?

Good catch, thanks. I have removed them and updated the PR 👌🏼

philn (approved):

PR #550, opened 2023-03-10
Add support for the new WPE WebKit 2.0 API

aperezdc:
This brings the contents of the `mcatanzaro/next` branch, which already supported building against the 2.0 API, and re-adds support for the 1.1 API (which we want to keep supporting as long as it is part of WebKit).

Thanks to @mcatanzaro, @carlosgcampos, and @TingPing for making the 2.0 API support possible.


aperezdc:
The commits were already reviewed with the PRs targeting `mcatanzaro/next`, and I only re-added the Meson bits to support API 1.1, so I am going to go ahead and land this in `master` so I can make a development release 💾

PR #552, opened 2023-03-11
CI: Save checksum of toolchain installer insted of timestamp

aperezdc:
Improve the logic around reusing a cached toolchain installation by storing the toolchain installer checksum file instead of the time of installation. This way we can check that the cached installation corresponds to what is expected, or do it afresh otherwise. This is more robust than the existing approach that trusts that any file on the server newer than the installation timestamp should be preferred.


aperezdc:
As a (desirable) side effect, this will also allow branches to point to different toolchains other than “the newest in the server”, which is nice to avoid breaking the CI for PRs targeting a branch other than `master` 😉

PR #553, opened 2023-03-13
x11, wl: Make scroll wheel behaviour snappier

aperezdc:
Adjust the amount of page scrolled when receiving scroll wheel events to make scrolling feel snappier in the Wayland and X11 platforms snappier. The 120 step value comes from the fact that libinput includes the `libinput_event_pointer_get_scroll_value_v120()` function, which in turn is based on a recommendation contained in a design document written by Microsoft.

Also, the table of deltas for X11 has its values reversed, to make the scroll feel more natural, i.e. "pulling" a scroll wheel makes the page scroll downwards. This matches defaults in common desktop environments.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -974,23 +974,28 @@ pointer_uses_frame_event(struct wl_pointer *pointer)
 #endif
 }
 
+/*
+ * Wayland reports axis events as being 15 units per scroll wheel step. Scale
+ * values in order to match the 120 value used in the X11 plug-in and by the
+ * libinput_event_pointer_get_scroll_value_v120() function.
+ */
+enum {
+    SCROLL_WHEEL_STEP_SCALING_FACTOR = 8,
+};
+
 static void
-pointer_on_axis (void* data,
-                 struct wl_pointer *pointer,
-                 uint32_t time,
-                 uint32_t axis,
-                 wl_fixed_t value)
+pointer_on_axis(void *data, struct wl_pointer *pointer, uint32_t time, uint32_t axis, wl_fixed_t value)
 {
     if (axis == WL_POINTER_AXIS_VERTICAL_SCROLL) {
         wl_data.axis.has_delta = true;
         wl_data.axis.time = time;
-        wl_data.axis.y_delta += value;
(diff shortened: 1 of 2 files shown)


aperezdc:
Talking with @manuelafm he was wondering about where the hardcoded values come from. I basically tried a few values in the X11 plugin until it “felt right”, then figured out which amount of scaling in the Wayland plug-in would result in a similar amount of scroll done for each step of the scroll wheel. That's it.

Turns out libinput has a function to [normalize scroll values](https://wayland.freedesktop.org/libinput/doc/1.22.0/api/group__event__pointer.html#ga31d3c555e912f021d3880d1cd7eb8a49) into “v120 units”, regardless of the resolution of the input device. If I am understanding correctly, the intention correctly, 120px has been used extensively and feels “good” for the amount of content to scroll for one scroll wheel event, and comes from Microsoft... which makes sense I guess knowing that the MS IntelliMouse was the one that popularized scroll wheels in PCs. The libinput documentation even links to a document by Microsoft which mentions the 120 value, which for “clicky” low resolution scroll wheels (the kind that produces a button event on X11) results in multiplying each “step” of the wheel by 120.

For the DRM/KMS plug-in Cog is already using `libinput_event_pointer_get_scroll_value_v120()` so I understand we do not need a change there. I will update this to use 120 as step size, and figure out which would be the correct scaling in the Wayland plug-in to end up using 120 for each wheel step.

aperezdc:
I added a log message, and turns out Wayland is reporting each step as `15`, so let's use `8` as multiplier there, to make them both match:

```
Cog-Wayland-Message: 11:29:04.366: pointer_on_axis, value = 15, scaled 75
Cog-Wayland-Message: 11:29:05.406: pointer_on_axis, value = -15, scaled -75
```

manuelafm (approved):
I think that it's a good solution and well researched and documented, great work :clap:

clopez:
I just discovered there is a PR on WebKit trying to fix this in WebCore instead: https://github.com/WebKit/WebKit/pull/11197

aperezdc:
> I just discovered there is a PR on WebKit trying to fix this in WebCore instead: [WebKit/WebKit#11197](https://github.com/WebKit/WebKit/pull/11197)

I am aware of it. When/if a patch like that gets accepted into WebKit we may need to adjust Cog accordingly, guarded with `WEBKIT_CHECK_VERSION()`.

PR #554, opened 2023-03-18
Fix touch limit raw region when  rotate screen

iamdrq:
I have a touchscreen:
```
cog -P drm -O renderer=gles,rotation=1
```
but the touchscreen libinput  report `Rotation` is not available 
```
$ libinput list-devices
Device:           Goodix Capacitive TouchScreen
Kernel:           /dev/input/event2
Group:            2
Seat:             seat0, default
Capabilities:     keyboard touch 
Tap-to-click:     n/a
Tap-and-drag:     n/a
Tap drag lock:    n/a
Left-handed:      n/a
Nat.scrolling:    n/a
Middle emulation: n/a
Calibration:      identity matrix
Scroll methods:   none
Click methods:    none
Disable-w-typing: n/a
Disable-w-trackpointing: n/a
Accel profiles:   n/a
Rotation:         n/a
```
so I rotate touchscreen use device tree,then touch limit raw region when  rotate screen,this patch fix it when rotate angle  on 90 or 270.

By the way,I use  device tree rotate the display pannel,but I use weston or gnome not need rotate again like cog. so maybe some thing not right on cog.

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -1201,6 +1201,11 @@ init_input(CogDrmPlatform *platform)
     input_data.input_width = drm_data.mode->hdisplay;
     input_data.input_height = drm_data.mode->vdisplay;
 
+    if (platform->rotation == COG_GL_RENDERER_ROTATION_90 || platform->rotation == COG_GL_RENDERER_ROTATION_270){
+        input_data.input_width = drm_data.mode->vdisplay;
+        input_data.input_height = drm_data.mode->hdisplay;
+    }
+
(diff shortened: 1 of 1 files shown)


aperezdc:
Hi @iamdrq, my apologies for missing this pending PR until now. Indeed you are right, and the logical sizes of the input area need to be swapped for the 90° and 270° rotations&mdash;this patch makes rotated touch input work as expected.

Thanks for the fix!

aperezdc (approved):

aperezdc:
Cherry-picked into the `cog-0.18` branch as commit 119346266394b41a418f97301bc5681b17490b42.

PR #561, opened 2023-04-27
Allow to pass the values for platform and platform-params via env vars.

clopez:
On some cases it is useful to be able to specify the Cog platform plugin and the platform parameters via environment variables.

For example: bots running CI tests may be executing the same command to start Cog but each bot can be configured to use a different platform plugin via enviroment variables.

Another example: The script `run-benchmark` from WebKit doesn't allow the user to pass custom command-line parameters to the browser, so with this environment variables the user should be able to select the platform parameters for Cog via the environment.

### DIFF
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -222,21 +222,28 @@ platform_setup(CogLauncher *self)
      * a given platform.
      */
 
-    g_debug("%s: Platform name: %s", __func__, s_options.platform_name);
+    /* The documentation of g_getenv() says that the returned string may be overwritten by the next
+     * call to g_getenv(), g_setenv() or g_unsetenv(), so avoid passing it directly as parameter to
+     * other functions that may call any of the *env() functions and then try to use the parameter. */
+    g_autofree const char *platform_name = g_strdup(s_options.platform_name ?: g_getenv("COG_PLATFORM_NAME"));
+    g_clear_pointer(&s_options.platform_name, g_free);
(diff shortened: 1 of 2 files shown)


launcher/cog-launcher.c
```diff
@@ -222,21 +222,28 @@ platform_setup(CogLauncher *self)
      * a given platform.
      */
 
-    g_debug("%s: Platform name: %s", __func__, s_options.platform_name);
+    /* The documentation of g_getenv() says that the returned string may be overwritten by the next
+     * call to g_getenv(), g_setenv() or g_unsetenv(), so avoid passing it directly as parameter to
+     * other functions that may call any of the *env() functions and then try to use the parameter. */
+    g_autofree const char *platform_name = g_strdup(s_options.platform_name ?: g_getenv("COG_PLATFORM_NAME"));
```
TingPing, line 228: I don't understand this.

`cog_platform_new()` does not store the pointer to `platform_name` at all. It never lives beyond the function call. This seems to also be true for `cog_platform_setup()`.

If that is not true for either of them they should call `g_strdup()`.

Every interaction with env vars is not thread safe. But this is all early in init right? There shouldn't be threads modifying these?

TingPing: OK, I see. `setup()` does call `getenv()` in multiple cases. *Potentially* invalidating it there.

TingPing (approved):

clopez:
Thanks!

PR #563, opened 2023-04-28
drm: ignore key event if no xkb context is available

jdknight:
When a key event occurs, Cog will ask WPE from its managed XKB context for an associated key code. If WPE fails to acquire a default XKB context, requests to fetch the state/key-code is undefined. This will cause Cog to fault on a first detected input.

Part of the issue is that `wpe_input_xkb_context_get_default` does not gracefully handle failed attempts to initialize the XKB context \[1\]. Assuming WPE does give Cog proper notice of a fail context, Cog should also limiting attempts to query state information from a context that is not available (assuming Cog desires to handle an inputless environment).

This commit tweaks the key event handling to drop any key event if the XKB context is not available.

\[1\]: https://github.com/WebPlatformForEmbedded/libwpe/pull/126

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -689,7 +689,24 @@ static void
 input_dispatch_key_event (uint32_t time, uint32_t key, enum libinput_key_state state)
 {
     struct wpe_input_xkb_context *default_context = wpe_input_xkb_context_get_default ();
+
+    // if wpe is unable to prepare an xkb context (e.g. environment without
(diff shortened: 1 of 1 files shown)


aperezdc (approved):
Very nice, specially the comments in the code explaining what's going on. Thanks for the patch!

aperezdc:
@jdknight Could you do a quick amend to make the style checker happy? There is only [one small change](https://github.com/Igalia/cog/actions/runs/4825963053/jobs/8690421954?pr=563) to apply.

PR #565, opened 2023-05-04
launcher: Simplify by using global CogPlatform

aperezdc:
Remove the unneeded instance variable that tracks the current platform in `CogLauncher`, instead use the default global `CogPlatform` which has been around since a3815bb.


zdobersek (approved):

aperezdc:
Thanks @zdobersek 😸

PR #566, opened 2023-05-05
Add a CogFallbackPlatform implementation

aperezdc:
This PR includes two commits:

- The first adds a new `CogFallbackPlatform` implementation, which can handle plain WPE backends (like WPEBackend-rdk). This gets built into the core library and thus is always available, instead of building it as a plug-in.
- The second removes the ad-hoc handling of plain WPE backends, which are left to be handled by `CogFallbackPlatform`.

### DIFF
--- a/core/cog-fallback-platform.c
+++ b/core/cog-fallback-platform.c
@@ -1,0 +1,84 @@
+/*
+ * cog-fallback-platform.c
+ * Copyright (C) 2023 Igalia S.L.
+ *
+ * SPDX-License-Identifier: MIT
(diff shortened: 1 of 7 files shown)


aperezdc:
This is something I had in a local branch with improvements to the ease of use of the public API, and could be easile extracted into its own patch set. Tested with WPEBackend-rdk built with its Wayland-EGL support, but I expect all other build modes should work, as well as other WPE backends which support this mode of operation 😉

zdobersek (approved):

PR #568, opened 2023-05-11
headless: Move frame tick source to platform object

aperezdc:
Move the variable for the frame tick source into `CogHeadlessPlatform`; there is no reason for it to be part of the `CogHeadlessView` type. Moreover, once multiple views are supported, it would be better to have a single timer to update all of them at once, instead of a timer per view.


nikolaszimmermann (approved):
r=me.

PR #570, opened 2023-05-11
core: Introduce a new CogView class

aperezdc:
Add a new `CogView` class, which extends `WebKitWebView` to avoid needing to specify the view backend for each instantiated view. Instead, the creation is delegated to the current CogPlatform, in one of the following two ways:

- Overriding a new `get_view_type` vfunc, which returns the `GType` of a `CogView` subclass specific to the platform plug-in. The custom view class needs to implement a `create_backend` vfunc that creates the view backend to use.

- Letting the core library use its fallback `CogView` subclass, which relies on the platform's `get_view_backend` vfunc. This matches the existing behaviour.

This way platform plug-ins can opt-in to the new behaviour by implementing their own `CogView` subclass, and unchanged plug-ins will continue working during the transition.


aperezdc:
Note that this patch, in the same vein as #566, removes the code to handle plain WPE backends. If this gets merged before, plain backends will not work until the other PR is also merged, so merging #566 first would be ideal.

zdobersek (approved):

PR #571, opened 2023-05-15
wl: fix blurry rendering in some compositors

aperezdc:
Improve tracking of output properties and setting output scaling factor to avoid blurry rendering due to using 1x scaling factor on output that use 2x scaling (or higher). The main offender was a missing call to `wpe_view_backend_dispatch_set_device_scale_factor()` in `resize_window()`.

Instead of keeping only the scaling factor for for the Wayland surface, keep a reference to its current output device. The output is initially set to the first one reported by the Wayland compositor (which usually is the primary output), ensuring that the pointer is always valid. This allows setting fallback values for device output scaling and refresh rate in `output_handle_done()`. Combined, both changes allow keeping other code simple, as it can assume the values will always be valid.

Overall, this makes surface output handling a bit more robust, avoiding slight behaviour changes depending on the particular sequence of events produced by the Wayland compositor being used.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -179,15 +179,12 @@ static struct {
 #endif /* COG_USE_WAYLAND_CURSOR */
 
     struct output_metrics metrics[16];
+    struct output_metrics *current_output;
 
     struct zwp_text_input_manager_v3 *text_input_manager;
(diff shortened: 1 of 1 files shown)


aperezdc:
Cherry-picked as commit 31d11ba45162e101bf1847a7845dd227e5c67301 in the `cog-0.18` branch.

PR #572, opened 2023-05-16
core: Move platform selection logic into the library

aperezdc:
Take the logic of `platform_setup_once()` out from the launcher into the core library. This will allow reusing the same implementation in other programs.

### DIFF
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -138,3 +138,69 @@ cog_platform_create_im_context(CogPlatform *platform)
 
     return NULL;
 }
+
+/**
+ * cog_platform_configure: (constructor)
+ * @name: (nullable): Name of the platform implementation to use.
+ * @params: (nullable): Parameters passed to the platform implementation.
+ * @env_prefix: (nullable): Prefix for environment variables to check.
+ * @shell: The shell that the platform will be configured for.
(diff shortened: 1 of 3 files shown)


core/cog-platform.c
```diff
@@ -138,3 +138,69 @@ cog_platform_create_im_context(CogPlatform *platform)
...
+ * Environment variables will *not* be used if %NULL is passed as the
+ * prefix.
+ *
+ * If both @name is %NULL and the `<env_prefix>_PLATFORM_NAME` variable
+ * not defined, then the platform implementation will be chosen automatically
+ * among the available ones.
+ *
+ * Note that [id@cog_modules_add_directory] may be used beforehand to
+ * configure where to search for available platform plug-ins.
+ *
```
TingPing, line 168: It seems this can only be called a single time, maybe that should be documented?

TingPing (approved):

core/cog-platform.c
aperezdc: It should be possible to call `cog_modules_add_directory` more than once just fine.

TingPing: I meant the configure function. It sets a global and warns if it was already set.

PR #574, opened 2023-05-19
core: Move key binding handling into CogView

aperezdc:
Pick the key binding handling from the Wayland platform plug-in, and move it into the core library. A new `cog_view_handle_key_event()` method takes care of detecting supported key bindings or actually sending the events to the view backend otherwise. In order to support cases where key bindings are not desirable (e.g. kiosks) it is possible to disable key binding support.

The only bit missing to move is the <kbd>F11</kbd> key binding to (un)fullscreen as the operation may need platform view specific actions, so for now the handling for this binding is left as it was in the Wayland platform plug-in. This can be improved later on in a follow-up patch.

### DIFF
--- a/core/cog-view.c
+++ b/core/cog-view.c
@@ -24,10 +24,24 @@
  * A number of utility functions are also provided.
  */
 
-G_DEFINE_ABSTRACT_TYPE(CogView, cog_view, WEBKIT_TYPE_WEB_VIEW)
+typedef struct {
+    gboolean use_key_bindings;
+} CogViewPrivate;
+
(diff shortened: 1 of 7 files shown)


aperezdc:
I am working also going to send a follow-up patche to add an option to disable key bindings in the `cog` launcher (for kiosk use).

Additionally, I have ideas to introduce a “key binding controller” which can be assigned to web views, and even when that's not a priority for me at the moment I have a local WIP branch from last year which might be quick to revive ⌨️

TingPing (approved):

aperezdc:
Thanks @TingPing 😄

PR #579, opened 2023-05-28
CI: Switch native builds to Ubuntu 22.04 LTS

aperezdc:
Change the base images for the native CI builds to Ubuntu 22.04 LTS. In turn, this allows to use system packages for gi-docgen, meson, and libportal, thus removing the need for manually-installed Python packages. The later makes the build faster to complete.

### DIFF
--- a/.github/workflows/ci-native.yml
+++ b/.github/workflows/ci-native.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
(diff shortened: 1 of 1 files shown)


TingPing:
```yaml
      - name: Setup Python
        uses: actions/setup-python@v4
        with:
          python-version: 3.8
      - name: Install Python Packages
        run: |
pip install --upgrade pip gi-docgen==2021.6 meson==0.55
```

Just worth noting Ubuntu 22.04 has newer versions of Python, meson, and gi-docgen in its repos.

psaavedra (approved):
Awesome.

PR #581, opened 2023-06-03
launcher: Handle GApplication::activate

TingPing:
GApplication warns if you do not handle activate.

For now this does nothing, typically it would focus the main window of an application, which doesn't apply to many of Cog's platforms.

### DIFF
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -374,6 +374,14 @@ cog_launcher_startup(GApplication *application)
 }
 
 static void
+cog_launcher_activate(GApplication *application)
+{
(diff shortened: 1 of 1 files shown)


launcher/cog-launcher.c
```diff
@@ -373,6 +373,12 @@ cog_launcher_startup(GApplication *application)
 #endif
 }
 
+static void
+cog_launcher_activate(GApplication *application)
+{
+    G_APPLICATION_CLASS(cog_launcher_parent_class)->activate(application);
```
clopez, line 379: I would add a comment here saying that we handle this just to avoid a warning but in practice we do nothing so far

clopez (approved):

PR #585, opened 2023-06-15
wl: Remove wl_surface_frame as part of wpe_view_data

psaavedra:
The wl_surface_frame callback has all that it needs from the context. It also doesn't requires a reference in wpe_view_data since it is auto destroyed during the callback action.

Also formatting fixes.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -9,29 +9,29 @@
 #include "../../core/cog.h"
 
 #include <errno.h>
+#include <glib-object.h>
+#include <glib.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <wpe/webkit.h>
-#include <wpe/fdo.h>
-#include <wpe/fdo-egl.h>
-#include <stdlib.h>
 #include <stdio.h>
-#include <glib.h>
-#include <glib-object.h>
+#include <stdlib.h>
 #include <string.h>
+#include <wpe/fdo-egl.h>
+#include <wpe/fdo.h>
+#include <wpe/webkit.h>
 
-#include <wayland-client.h>
-#include <wayland-server.h>
-#include <wayland-egl.h>
 #include <EGL/egl.h>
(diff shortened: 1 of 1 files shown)


platform/wayland/cog-platform-wl.c
```diff
@@ -1460,10 +1455,8 @@ static const struct wp_presentation_feedback_listener presentation_feedback_list
 static void
 request_frame(void)
 {
-    if (wpe_view_data.frame_callback == NULL) {
-        wpe_view_data.frame_callback = wl_surface_frame(win_data.wl_surface);
-        wl_callback_add_listener(wpe_view_data.frame_callback, &frame_listener, NULL);
-    }
+    struct wl_callback *frame_callback = wl_surface_frame(win_data.wl_surface);
+    wl_callback_add_listener(frame_callback, &frame_listener, NULL);
```
aperezdc: Removing the variable and the null-check can result in more than one frame callback being registered for the same frame. The check was there to allow calling `request_frame()` from different places but still having only one callback registered for each frame.

Are you sure that without the check we never schedule more than one callback per frame?

aperezdc:
I would remove the commit with all the formatting fixes, and only do the minimum amount of formatting required by the changes in the other patch. The only formatting change needed to keep the style checker happy seems to be this one:

```diff
--- platform/wayland/cog-platform-wl.c	(before formatting)
+++ platform/wayland/cog-platform-wl.c	(after formatting)
@@ -305,7 +305,7 @@
 static struct {
     struct wpe_view_backend *backend;
     struct wpe_fdo_egl_exported_image *image;
-    struct wl_buffer *buffer;
+    struct wl_buffer                  *buffer;
     bool should_update_opaque_region;
 } wpe_view_data = {
     .should_update_opaque_region = true, /* Force initial update. */
```

aperezdc:
@psaavedra Ping 🏓, please let me know about the question above when you have time. Thanks!

psaavedra:
This PR is obsolete. Closing it.

Issue #584, opened 2023-06-14
No window shown using non-root user


diorcety:
I'm building a yocto distribution using weston 10.0.2, wpewebkit 2.38.2, cog 0.16.1 and I fall on a strange issue.

If i'm runnin the following command the flower are shown on the screen
`sudo  sh -c 'XDG_RUNTIME_DIR=/run G_MESSAGES_DEBUG=all WAYLAND_DEBUG=1 weston-flower'`

same for
`sudo su browser sh -c 'XDG_RUNTIME_DIR=/run G_MESSAGES_DEBUG=all WAYLAND_DEBUG=1 weston-flower'`

For cog, if i run using sudo, the window is shown fullscreen on the screen
`sudo  sh -c 'XDG_RUNTIME_DIR=/run G_MESSAGES_DEBUG=all WAYLAND_DEBUG=1 cog'`

But using browser user, nothing is displayed and no errors on the output:
sudo su browser sh -c 'XDG_RUNTIME_DIR=/run G_MESSAGES_DEBUG=all WAYLAND_DEBUG=1 cog'

Here the working output:
```
(process:1226): Cog-Core-DEBUG: 07:54:31.843: ensure_extension_points: Extension points registered.
(process:1226): Cog-Core-DEBUG: 07:54:31.855: cog_modules_add_directory: Scanning '/usr/lib/cog/modules'
Cog-INFO: 07:54:31.938: com.igalia.Cog.Platform:
Cog-INFO: 07:54:31.943:   wl - 500/CogWlPlatform
Cog-INFO: 07:54:31.948:   drm - 200/CogDrmPlatform
Cog-INFO: 07:54:31.950:   headless - 100/CogHeadlessPlatform
(cog:1226): GLib-GIO-DEBUG: 07:54:32.087: _g_io_module_get_default: Found default implementation local (GLocalVfs) for ?gio-vfs?
(cog:1226): Cog-DEBUG: 07:54:32.129: platform_setup: Platform name: (null)
[1125729.615]  -> wl_display@1.get_registry(new id wl_registry@2)
[1125740.336]  -> wl_display@1.sync(new id wl_callback@3)
[1125751.183] wl_display@1.delete_id(3)
[1125760.754] wl_registry@2.global(1, "wl_compositor", 4)
[1125766.323] wl_registry@2.global(2, "wl_subcompositor", 1)
[1125771.484] wl_registry@2.global(3, "wp_viewporter", 1)
[...]

PR #586, opened 2023-06-16
Meson: check file sys/mman.h in order to define HAVE_MEMFD_CREATE

diorcety:
Will "fix" #584. Currently os_create_anonymous_file can use memfd_create in order to create an anonymous file. It depends on HAVE_MEMFD_CREATE define, but HAVE_MEMFD_CREATE is never defined.
This PR add a check in meson in order to define HAVE_MEMFD_CREATE  if the symbol memfd_create exists in sys/mman.h


philn:
Thanks! Can you add `Fixes #584` on the 3rd line of the commit message?

diorcety:
Done

PR #588, opened 2023-07-06
Manage CAPS_LOCK and NUM_LOCK keys

dpino:
Depends on https://github.com/WebPlatformForEmbedded/libwpe/pull/128.

Implement handling of CAPS_LOCK and NUM_LOCK for several layers (GTK4, Wayland, X11).

### DIFF
--- a/platform/gtk4/cog-platform-gtk4.c
+++ b/platform/gtk4/cog-platform-gtk4.c
@@ -341,6 +341,18 @@ on_scroll(GtkEventControllerScroll* controller, double dx,
 }
 
 static gboolean
+is_pressed_key(guint keyval)
+{
+    static bool initialized = false;
+
+    if (!initialized) {
+        GUniqueOutPtr<GdkKeymapKey> keys;
+        int entriesCount;
+    }
+    return gdk_display_map_keyval(gdk_display_get_default(), keyval, &keys.outPtr(), &entriesCount) && entriesCount;
+}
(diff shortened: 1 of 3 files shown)


platform/gtk4/cog-platform-gtk4.c
```diff
@@ -340,6 +340,18 @@ on_scroll(GtkEventControllerScroll* controller, double dx,
     return TRUE;
 }
 
+static gboolean
+is_pressed_key(guint keyval)
+{
+    static bool initialized = false;
+
+    if (!initialized) {
+        GUniqueOutPtr<GdkKeymapKey> keys;
```
aperezdc, line 349: This won't work on a plain C source. It looks like you were picking from C++ code from elsewhere, so this will need adapting to compile as C99.

platform/gtk4/cog-platform-gtk4.c
```diff
@@ -340,6 +340,18 @@ on_scroll(GtkEventControllerScroll* controller, double dx,
     return TRUE;
 }
 
+static gboolean
+is_pressed_key(guint keyval)
```
aperezdc, line 344: Small nit: naming this `is_key_pressed()` would read more naturally.

aperezdc (changes requested):
There's an issue in the patch, please see the comment about mixing C and C++ code below.

PR #590, opened 2023-07-12
drm: Fix crash on imx6 when no current encoder+crtc is bound

zhani:
imx6 crashes when trying drm backend. This depends heavily on board used and this fix is generic. Doesn't affect existing configs but just fixed edge case for specific board when no default endoder+crtc was found

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -353,7 +353,8 @@ check_drm(void)
 }
 
 static int32_t
-find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder) {
+find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder)
+{
     for (int i = 0; i < resources->count_crtcs; i++) {
         const uint32_t crtc_mask = 1 << i;
         const uint32_t crtc_id = resources->crtcs[i];
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -354,9 +354,7 @@ check_drm(void)
 
 static int32_t
 find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder) {
-    int i;
-
-    for (i = 0; i < resources->count_crtcs; i++) {
+    for (int i = 0; i < resources->count_crtcs; i++) {
         const uint32_t crtc_mask = 1 << i;
         const uint32_t crtc_id = resources->crtcs[i];
         if (encoder->possible_crtcs & crtc_mask) {
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -352,6 +352,22 @@ check_drm(void)
     return supported;
 }
 
+static int32_t
+find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder) {
+    int i;
+
+    for (i = 0; i < resources->count_crtcs; i++) {
+        const uint32_t crtc_mask = 1 << i;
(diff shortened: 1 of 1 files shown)


psaavedra (approved):
LGTM.

platform/drm/cog-platform-drm.c
```diff
@@ -352,6 +352,22 @@ check_drm(void)
     return supported;
 }
 
+static int32_t
+find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder) {
+    int i;
+
+    for (i = 0; i < resources->count_crtcs; i++) {
```
aperezdc, line 359: Cog is built in C11 mode, which means we can move the declaration of the variable used for loop control inside the first expression of the `for` statement:

```suggestion
    for (int i = 0; i < resources->count_crtcs; i++) {
```

aperezdc (approved):
Good patch, thanks @zhani for submitting it!

platform/drm/cog-platform-drm.c
```diff
@@ -352,6 +352,22 @@ check_drm(void)
     return supported;
 }
 
+static int32_t
+find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder) {
```
aperezdc, line 356: You need to move the opening brace to its own line to make the style checker happy:

```suggestion
find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder)
{
```

aperezdc (changes requested):
@zhani Could you update the patch taking into account the two small suggestions I made? With those in, I'd be happy to merge the changes. Thanks 🙏🏼

platform/drm/cog-platform-drm.c
```diff
@@ -352,6 +352,21 @@ check_drm(void)
...
+    for (int i = 0; i < resources->count_crtcs; i++) {
+        const uint32_t crtc_mask = 1 << i;
+        const uint32_t crtc_id = resources->crtcs[i];
+        if (encoder->possible_crtcs & crtc_mask) {
+            return crtc_id;
+        }
+    }
+
+    /* no match found */
+    return -1;
```
aperezdc, line 367: This should return zero, because later below the code checks for `crtc_id != 0`.

platform/drm/cog-platform-drm.c
```diff
@@ -494,19 +509,30 @@ init_drm(void)
...
     for (int i = 0; i < drm_data.base_resources->count_encoders; ++i) {
         drm_data.encoder = drmModeGetEncoder(drm_data.fd, drm_data.base_resources->encoders[i]);
-        if (drm_data.encoder->encoder_id == drm_data.connector.obj->encoder_id)
+        if (!drm_data.encoder) {
+            /* cannot retrieve encoder, ignoring... */
+            continue;
+        }
+
+        const int32_t crtc_id = find_crtc_for_encoder(drm_data.base_resources, drm_data.encoder);
+        if (crtc_id != 0) {
```
aperezdc, line 521: Either the `find_crtc_for_encoder()` is changed to return zero above, or this should check `crtc_id != -1`.

aperezdc (changes requested):
I've found a small logic error while re-reviewing this. Sorry for not noticing in my previous review 😇 &mdash; could you please take a look and update the PR accordingly?

platform/drm/cog-platform-drm.c
zhani: Fixed it by handing result correctly in calling function

platform/drm/cog-platform-drm.c
zhani: fixed with != -1

aperezdc (approved):
Thanks @zhani, patch looks perfect now!

PR #598, opened 2023-08-28
launcher: Add command-line option for controlling autoplay behavior

### LINKED ISSUE #597: Autoplay policy support
It would be nice to have an option to control autoplay policy, like in GTK's MiniBrowser.

philn:
Fixes #597

### DIFF
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -13,6 +13,7 @@
 #include <string.h>
 
 #define HAVE_WEBKIT_NETWORK_PROXY_API WEBKIT_CHECK_VERSION(2, 32, 0)
+#define HAVE_WEBKIT_AUTOPLAY          WEBKIT_CHECK_VERSION(2, 30, 0)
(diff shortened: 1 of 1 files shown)


launcher/cog-launcher.c
```diff
@@ -1028,6 +1047,32 @@ option_entry_parse_gamepad(const char *option_name, const char *value, void *dat
     return TRUE;
 }
 
+#if HAVE_WEBKIT_AUTOPLAY
+static gboolean
+option_entry_parse_autoplay(const char *option_name, const char *value, void *data, GError **error)
+{
+    if (!g_strcmp0(value, "allow")) {
```
aperezdc, line 1054: We could match the passed value against the possible values for the GEnum class by using `g_enum_get_value_by_nick()` instead of manually checking possible values. I think it's fine to merge this patch as-is and do that as a follow-up later on.

aperezdc (approved):

PR #603, opened 2023-09-25
Fix deleting the view after signal close

pgorszkowski-igalia:
WebKitWebView inherits from GObject so it is not needed to do g_object_ref_sink on it. It fixes the problem with deleting the view after receiving "close" signal.

### DIFF
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -168,7 +168,6 @@ cog_shell_startup_base(CogShell *shell)
     }
 
     g_signal_emit (shell, s_signals[CREATE_VIEW], 0, &priv->web_view);
-    g_object_ref_sink (priv->web_view);
     g_object_notify_by_pspec (G_OBJECT (shell), s_properties[PROP_WEB_VIEW]);
 
     /*


pgorszkowski-igalia:
> Makes sense.
> 
> I assume this mistake was because in WebKitGTK it is a GtkWidget which is GInitiallyUnowned. Indeed in WPE it is a normal GObject.

So maybe I should use `if (G_IS_INITIALLY_UNOWNED (priv->web_view))` ?

aperezdc:
> > Makes sense.
> > I assume this mistake was because in WebKitGTK it is a GtkWidget which is GInitiallyUnowned. Indeed in WPE it is a normal GObject.
> 
> So maybe I should use `if (G_IS_INITIALLY_UNOWNED (priv->web_view))` ?

No need. Cog can only use WPE WebKit these days, we no longer support linking against WebKitGTK.

aperezdc (approved):

aperezdc:
Cherry-picked into the `cog-0.18` branch as commit a888726290cf7d27e46a6dfa7b80eaff681b4cac, too.

Issue #594, opened 2023-08-14
No cursor change on links / buttons and other problems


ancorgs:
We are building a [Linux installer with web interface](https://github.com/openSUSE/agama) that will hopefully become the default installer for the next generation of openSUSE and SUSE distributions.

To enable local installation (ie. traditional Linux installation experience), we are creating a demo live system that automatically opens a browser in full screen. Cog sounds perfect for that, but we have found some problems. Let's start with the main one, illustrated in the following video.

https://github.com/Igalia/cog/assets/3638289/fcf7cbd4-8b72-4834-a466-c8ae35e7937f

As you can see, the cursor doesn't change when we run the mouse over a clickable element. That's quite a show stopper from an usability point of view.

In the same video you can also see another problem which is basically the one reported at #327 (no handler for `input type=file`).

You can try it yourself using this ISO: [agama-live-weston-cog.x86_64-openSUSE.iso](https://download.opensuse.org/repositories/systemsmanagement:/Agama:/Staging/images/iso/agama-live-weston-cog.x86_64-openSUSE.iso) in any real or virtual machine.

In the ISO we use Cog directly on top of Weston, but we have observed the same issue with other compositors as well (eg. on top of a Gnome Wayland session).

¿Anything we can do to have a more reactive cursor?

Just since I also mentioned #327 ¿any guidance on how to handle file inputs in our case?

ancorgs:
Unless I'm mistaken and following a red herring, it looks like this is the code rendering the cursor on Wayland.

https://github.com/Igalia/cog/blob/master/platform/wayland/cog-platform-wl.c#L853
(Originally introduced at https://github.com/Igalia/cog/pull/194).

There is a comment there stating that the mouse cursor will not have the appropriate size in HiDPI. So I guess the lack of a hover icon would not be the only problem with cursors in Wayland.

Pinging @aperezdc since I see he is the main author of all that. What's the status regarding mouse cursor hover on one hand and regarding HiDPI on other hand?

PR #605, opened 2023-10-03
Update cursor on wl, x11 and gtk4

joantolo:
These changes allow the cursor to be updated when hovering on links or on input text fields. And the rest with the default pointer.

From a small conversation with @aperezdc on matrix, the idea was to create a signal at `WebKitWebView` and connect to it. However, the '`mouse-target-changed`' signal already gives some hovering mouse info, but not all. For instance, that signal doesn't inform when hovering over text.

As a first step, I've used '`mouse-target-changed`' signal.
A better signal could be emitted from `PageClientImpl::setCursor`. 
But when debugging cog, I see that `setCursor` isn't used, I don't know why.

The next step would be to use the `setCursor` to expose a signal with all cursor changes and add more cursor names. 

Fixes the cursor part of #594.


aperezdc (changes requested):
@joantolo I have a couple of comments, could you please take a look? While at it, please use the `data/check-style` script to ensure the patch follows the same coding style as the rest of the source code.

Regarding the usage of the signal, I agree that it would be ideal to plumb WebKit's `PageClientImpl::setCursor()` instead to make it work for WPE; but that will need either new public API either in libwpe or in WPE WebKit itself, which would be way more effort.

In the meantime, I would be more than happy to have this implementation merged in Cog, because it's better than nothing. Thanks for working on this!

aperezdc (changes requested):
@joantolo This is looking good, there's only one last detail before we can merge this: we need to add `libxcb-cursor-dev` in the list of packages to install during the CI builds, in `.github/workflows/ci-native.yml`.

If you could have a moment to rebase the patch and do that small change, then we can merge this. Thanks again!

joantolo:
Added `libxcb-cursor-dev` and rebased.

aperezdc (approved):
Thanks @joantolo, this is now ready to merge!

PR #606, opened 2023-10-06
Add file chooser on wl, x11 and 

joantolo:
Use the file chooser provided from the desktop using the `FileChooser` interface of Portal.

The only way I could find to use `XdpParent` for `wl` and `x11` was to add the [parent-header.-private.h](https://github.com/flatpak/libportal/blob/main/libportal/parent-private.h) from [libportal](https://github.com/flatpak/libportal).
And then add an implementation as it is done for [gtk](https://github.com/flatpak/libportal/blob/main/libportal/portal-gtk4.c).

On the wayland platform it was needed the addition of `xdg-foreign` interface which is needed by `XdpParent`. You can see the commits for a more detailed info.

I'm open to suggestions.

This fixes the file chooser part of #594.


aperezdc (changes requested):
@joantolo Thanks for working on this! I actually like the approach of using `libportal` to “offload” the responsibility of providing a suitable file dialog out of Cog, good idea.

There are a couple of things I commented around, mainly:

- [ ] Use `g_autoptr(...)` wherever possible.
- [ ] Declare variables right before first use (C11/C++ style).
- [ ] Make `libportal` suage optional.
- [ ] Use the `data/check-style` script to check that your code following the coding style.

With the above ironed out, I would be super happy to have this PR merged.

Picking the `parent-private.h` header from `libportal.h` I think is fine because the header itself has a comment stating that the license is different for that source file, it is unmodified from the original, and does not result in machine code added to Cog&mdash;only the definitions from it are needed at compile time. IIUC this is allowed as “combined work” by the LGPL, but IANAL so if anyone has a clearer view on this comments are welcome.

joantolo:
Should I modify the copy of xdp-parent-private.h to pass the check-style?

aperezdc:
> Should I modify the copy of `xdp-parent-private.h` to pass the check-style?

I think it's not needed. Let me try to find a way of making the style checker ignore certain files; if not I can merge the PR manually.

aperezdc (approved):
Let's merge this, looks good. Thanks @joantolo!

I have posted #616 to easily mark paths to be ignored by the style checker, prompted by this PR 😄

PR #616, opened 2023-10-19
check-style: Support reading a file with ignored paths

### LINKED ISSUE #615: Support ignoring files in the style checker
The `data/check-style` script should allow reading a list of paths to ignore.

aperezdc:
Add support for reading a `.check-style-ignore` file, with a regex in each line for ignored style checker paths.

Fixes #615

### DIFF
--- a/.check-style-ignore
+++ b/.check-style-ignore
@@ -1,3 +1,7 @@
 # File paths matching the regular expressions in this file (one per line)
 # will be skipped by data/check-style. Empty lines and lines starting with
 # a hash mark are ignored.
+
+platform/common/xdp-parent-private\.h$
+platform/drm/kms\.[hc]$
+platform/wayland/os-compatibility\.[hc]$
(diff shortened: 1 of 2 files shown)


.check-style-ignore
```diff
@@ -0,0 +1,7 @@
+# File paths matching the regular expressions in this file (one per line)
+# will be skipped by data/check-style. Empty lines and lines starting with
+# a hash mark are ignored.
+
+platform/common/xdp-parent-private\.h$
+platform/drm/kms\.[hc]$
+platform/wayland/os-compatibility\.[hc]$
```
psaavedra, line 7: What motivates the exclusion of these files?

aperezdc: These files contain utility code imported from other projects as-is, without modifications. The other projects may not follow the same conventions, so they result in false `check-style` positives (see #606 for an example).

psaavedra (approved):
LGTM

PR #617, opened 2023-10-19
wl: Create cog-utils-wl.h

psaavedra:
Move common definitions to cog-utils.h. The relevant structs are renamed as CogWl(Axis|Keyboard|Output|Pointer|Touch|Xkb)

Also the event_serial is moved inside of the CogWlKeyboard.


aperezdc (approved):
As this only moves definitions into the new `cog-utils-wl.h` header, I suppose this is a first round of cleanups, and that later on we would also split out some functions out from `cog-platform-wl.c`. Which is a plan that sounds good to me&mdash;it's never a bad idea to make the code easier to follow.

PR #618, opened 2023-10-20
Several code cleanups and refactorings

psaavedra:
These changes are motivated towards the progress of being able to have more than one view:

* Move declaration of CogWlPlatform to a header file
* More fix code-style in cog-platform-wl.c
* Encapsulate the seat resources inside of CogWlSeat. All the related resources to an individual wl_seat are encapsulated into a CogWlSeat struct. The CogWlDisplay contains a CogWlSeat (the first registered).
* Redefine wl_data and egl_data as CogWlDisplay. The CogWlDisplay is defined in cog-utils-wl.h. Also wl_data and egl_data definition are renamed as s_display.

### DIFF
--- a/platform/wayland/cog-utils-wl.h
+++ b/platform/wayland/cog-utils-wl.h
@@ -32,6 +32,7 @@
 #endif
 
 typedef struct _CogWlAxis     CogWlAxis;
+typedef struct _CogWlDisplay  CogWlDisplay;
 typedef struct _CogWlKeyboard CogWlKeyboard;
 typedef struct _CogWlOutput   CogWlOutput;
 typedef struct _CogWlPointer  CogWlPointer;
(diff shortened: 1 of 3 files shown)


psaavedra:
> Needs rebasing.

Yes, sorry for that. Done.

psaavedra:
Rebased against master after https://github.com/Igalia/cog/pull/605 being landed. @aperezdc please **r?**.

platform/wayland/cog-utils-wl.h
```diff
@@ -101,6 +104,25 @@ struct _CogWlXkb {
     uint8_t modifiers;
 };
 
+struct _CogWlSeat {
+    struct wl_seat *seat;
+    uint32_t        seat_name;
+    uint32_t        seat_version;
```
aperezdc, line 110: Funny thing: I just realized that `seat_version` is not used anywhere. Let's leave it be for now, but we could make a follow-up patch later on to remove it :wink:

aperezdc (changes requested):
I have a couple of small nits, otherwise the patch LGTM and we could merge it when the small comments are addressed. Thanks for the cleanup!

platform/wayland/cog-utils-wl.h
psaavedra: I am preparing another change for supporting multiple seats. I will keep this in mind.

Issue #357, opened 2021-09-06
Switch to SPDX code headers


philn:
As documented on https://spdx.dev

philn:
Identifier would be `MIT` IIUC?

aperezdc:
> Identifier would be `MIT` IIUC?

Yes, this is correct ✅

PR #619, opened 2023-10-22
wl: Create custom CogWlView 

psaavedra:
**No logical changes in the workflow.**

This change only creates the CogWlView as a specialization of the CogView. We    also move the logic for handling the exported buffers inside of this new    class.

    
This also brings other benefits since this also removes the     access to the global variables by adding context data to the listeners    defined in the CogWlPlatform and in the CogWlView classes. For this  reason, for example, the s_display is not longer needed as global and now is moved and accessible as part of the CogWlPlatform.
    
The are additional cosmetic changes like these:
    
* The struct win_data is renames as CogWlWindow
* Some functions are renamed with the cog_wl_* and cog_wl_view_* prefix
 
This change is motivated towards the progress of being able to have more    than one view.

**DEPENDS ON:** https://github.com/Igalia/cog/pull/618


platform/wayland/cog-view-wl.h
```diff
@@ -0,0 +1,52 @@
+/*
+ * cog-view-wl.h
+ * Copyright (C) 2023 Pablo Saavedra <psaavedra@igalia.com>
+ * Copyright (C) 2023 Adrian Perez de Castro <aperez@igalia.com>
+ *
+ * Distributed under terms of the MIT license.
```
aperezdc, line 6: While at it, let's use SPDX license identifiers for new files, so there will be less to change when we tackle #357 😉 

```suggestion
 * SPDX-License-Identifier: MIT
```

platform/wayland/cog-view-wl.h
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+#ifndef COG_PLATFORM_WL_VIEW_H
+#define COG_PLATFORM_WL_VIEW_H
```
aperezdc, line 10:
```suggestion
#pragma once
```

aperezdc (changes requested):
LGTM, as it's mainly moving code around; I have only a couple of small nits and we can merge this once solved. Thanks!

platform/wayland/cog-view-wl.h
psaavedra: applied

platform/wayland/cog-view-wl.h
psaavedra: applied

platform/wayland/cog-view-wl.c
```diff
@@ -0,0 +1,574 @@
+/*
+ * cog-utils-wl.c
```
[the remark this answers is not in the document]
psaavedra: fixed

platform/wayland/cog-platform-wl.h
```diff
@@ -10,17 +10,32 @@
...
+    CogWlWindow   window;
 };
 
+/*
+ * Method declarations.
+ */
+
+void cog_wl_popup_create(CogWlPlatform *platform, WebKitOptionMenu *);
+
+bool cog_wl_set_fullscreen(CogWlPlatform *platform, bool fullscreen);
```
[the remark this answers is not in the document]
psaavedra: 👌🏼 . Updated

core/cog-view.c
```diff
@@ -19,7 +19,7 @@
  * [struct@WPEWebKit.WebViewBackend] to the current [class@CogPlatform]
  * implementation. Optionally, platform plug-in implementations can
  * provide their own view class implementation by overriding
- * [vfunc@CogPlatform.get_view_type].
+ * the [class@CogPlatform] `get_view_type()` method.
```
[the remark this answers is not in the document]
psaavedra: Fixed. This came from https://github.com/Igalia/cog/commit/d50460e9a8deb4cfe9acdd49b285c04ba41c39c7 as it is.

aperezdc (approved):
Thanks @psaavedra, this is now ready 🆗

PR #622, opened 2023-10-23
[wip]  wl: Add CogWlPlatform fullscreen-enter and fullscreen-exit signals 

psaavedra:
Simply the code in CogWlPlatform by adding the "fullscreen-enter" and    "fullscreen-exit" signals to notify to the view that the window is being   entering or exiting from the fullscreen.
    
This change is motivated towards the progress of being able to have more than one view.

**DEPENDS ON:** https://github.com/Igalia/cog/pull/629

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -103,6 +103,11 @@ static struct {
     .configured = false,
 };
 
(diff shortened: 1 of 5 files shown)


psaavedra:
Discarded for now

PR #623, opened 2023-10-24
wl: Support multiple wl_seats 

psaavedra:
This change extend the WL platform to support more than one WL Seat registered to the Cog app.

Now the  CogWlDisplay contains a default seat. This is the first registered and then the last seat activated. This is used for the shell_popup_surface     and for the on_mouse_target_changed handlers. The CogWlDisplay also     contains the rest of the CogWlPlatform in a list.
    
The listeners associated to input methods (pointer, keyboard and touch) now pass the CogWlSeat as contextual information to the callbacks. The    CogWlPlatform object is still accessible because this is attached    as user_data to the wl_surface on creation time and associated to the    seats as [pointer|keyboard|touch]_target then the input event is activated. These also updates what is the default seat.

This patch also fixes the tracking of the serials for the input methods. Now each different kind of input method for each Seat is stored     individually and updated according. Still it is possible to get the     latest serial using the suitable cog_wl_seat_get_serial() function.

Some other code was cleaned-up and relocated in cot-utils-wl as functions: `cog_wl_display_add_seat; cog_wl_seat_create();    cog_wl_seat_destroy(); void cog_wl_text_input_clear(); cog_wl_text_input_set()`.

Also CogWlDisplay create/destroy logic is wrapped inside of the cog_wl_display_create() and cog_wl_display_destroy() functions and move these to cog-utils-wl.


platform/wayland/cog-platform-wl.c
```diff
@@ -568,44 +565,64 @@ pointer_on_enter(void              *data,
...
-    display->seat.pointer.surface = surface;
+    if (pointer != seat->pointer_obj) {
+        g_critical("%s: Got pointer %p, expected %p.", G_STRFUNC, pointer, seat->pointer_obj);
+        return;
+    }
+
+    seat->display->seat_default = seat;
+
+    CogWlPlatform *platform =
+        wl_surface_get_user_data(surface); // Platform is attached the context of the wl_surface. Check create_window();
```
aperezdc, line 578: There is only one platform object, it's a singleton. Remember that you can always get it using `cog_platform_get_default()` so you don't need to attach it as user data to the surface. You can do the following here and in many other places in this patch:

```c
    CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get_default();
```

Note that we don't need to use `COG_WL_PLATFORM()` to do a checked cast: because there is only one object, if the Wayland platform code is executed, we _know_ that `cog_platform_get_default()` always returns a `CogWlPlatform`.

I did the same kind of simplification in https://github.com/Igalia/cog/pull/565, for example.

aperezdc: By the the way: this also means that there is no need to have a `CogWlPlatform*` saved inside `CogwlView` 😉

platform/wayland/cog-platform-wl.c
```diff
@@ -477,9 +502,17 @@ registry_on_global(void *data, struct wl_registry *registry, uint32_t name, cons
...
+        struct wl_seat *wl_seat = wl_registry_bind(registry, name, &wl_seat_interface, seat_version);
+        CogWlSeat      *seat = cog_wl_seat_create(wl_seat, name);
+        cog_wl_display_add_seat(display, seat);
+        static const struct wl_seat_listener seat_listener = {
+            .capabilities = seat_on_capabilities,
+#ifdef WL_SEAT_NAME_SINCE_VERSION
+            .name = seat_on_name,
+#endif // WL_SEAT_NAME_SINCE_VERSION
+        };
+        wl_seat_add_listener(wl_seat, &seat_listener, seat);
```
psaavedra, line 515: This comes from https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034L1710 but the seat_listener nows passes the CogWlSeat instance  as context.

platform/wayland/cog-platform-wl.c
```diff
@@ -980,11 +1029,21 @@ touch_on_down(void              *data,
...
+    if (touch != seat->touch_obj) {
+        g_critical("%s: Got touch %p, expected %p.", G_STRFUNC, touch, seat->touch_obj);
+        return;
+    }
+
+    seat->display->seat_default = seat;
+
+    CogWlPlatform *platform =
+        wl_surface_get_user_data(surface); // Platform is attached the context of the wl_surface. Check create_window();
+    seat->touch_target = platform;
```
aperezdc, line 1044: I find this counterintuitive; the target of events should be a CogView, or at least a `wl_surface`. Maybe the `*_target` variables need a better name, or the `CogWlPlatform` has no business being the “target”?

I think for now we can use `cog_platform_get_default()` as suggested above, and introduce the `*_target` variables in the future when/if needed. How does that sound to you?

platform/wayland/cog-platform-wl.c
```diff
@@ -524,42 +557,6 @@ registry_on_global(void *data, struct wl_registry *registry, uint32_t name, cons
     g_debug("%s '%s' interface obtained from the Wayland registry.", interface_used ? "Using" : "Ignoring", interface);
 }
 
-#ifdef COG_USE_WAYLAND_CURSOR
-void
-cog_wl_platform_set_cursor(CogWlPlatform *platform, enum cursor_type type)
```
psaavedra, line 609: This is renamed as cog_wl_seat_set_cursor:  https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-b5c8d7e31c380d86e8ae8aea3f2dc76fd5f407590ccc4a6cf2347ca0b9e17749R249

platform/wayland/cog-platform-wl.c
```diff
@@ -1173,6 +1239,14 @@ registry_on_global_remove(void *data, struct wl_registry *registry, uint32_t nam
             break;
         }
     }
+
+    CogWlSeat *seat, *tmp_seat;
+    wl_list_for_each_safe(seat, tmp_seat, &platform->display->seats, link) {
+        if (seat->seat_name == name) {
+            cog_wl_seat_destroy(seat);
```
aperezdc, line 1246: If theIf the seat being destroyed is the current one, we should check that and set it to `NULL`.

platform/wayland/cog-platform-wl.c
```diff
@@ -568,44 +565,64 @@ pointer_on_enter(void              *data,
...
-    CogWlDisplay  *display = platform->display;
+    CogWlSeat *seat = data;
 
-    display->seat.keyboard.serial = serial;
-    display->seat.pointer.serial = serial;
-    display->seat.pointer.surface = surface;
+    if (pointer != seat->pointer_obj) {
+        g_critical("%s: Got pointer %p, expected %p.", G_STRFUNC, pointer, seat->pointer_obj);
+        return;
+    }
```
psaavedra, line 573: These are additional safe checks that they have been added in the input events handlers just to be sure the real implementation meets with the expectations.

aperezdc (changes requested):
Looks good overall. The first commit can land as-is, but I have a few notes on the other one. Please take a look and update the patch, or comment if you think your current approach is better 😸

platform/wayland/cog-platform-wl.c
```diff
@@ -568,44 +565,64 @@ pointer_on_enter(void              *data,
...
+
+    seat->display->seat_default = seat;
```
psaavedra, line 575: This is important. When a input event is processed we update the seat_default reference to that seat associated to the event. The seat_default reference is used    by the im_context (https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R1866) and by the popups (https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R1719)

platform/wayland/cog-platform-wl.c
psaavedra, line 578: This is the way that I did to attach the Platform as a object associated to the surface. Check this: https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R133.

platform/wayland/cog-platform-wl.c
```diff
@@ -618,10 +635,15 @@ pointer_on_button(void              *data,
...
+    CogWlSeat    *seat = data;
+    CogWlDisplay *display = seat->display;
 
-    display->seat.keyboard.serial = serial;
+    if (pointer != seat->pointer_obj) {
+        g_critical("%s: Got pointer %p, expected %p.", G_STRFUNC, pointer, seat->pointer_obj);
+        return;
+    }
+
+    seat->pointer.serial = serial;
```
psaavedra, line 646: Here we are fixing the wrong use of `display->seat.keyboard.serial = serial;` (https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034L624)  for tracking all the serials for the different classes of inputs .  The `cog_wl_seat_get_serial()`gives us the latest serial received (https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-b5c8d7e31c380d86e8ae8aea3f2dc76fd5f407590ccc4a6cf2347ca0b9e17749R234).  For example this is important for the grab action in the popup menu: https://github.com/Igalia/cog/pull/623/commits/a321b21181571ea271b3465d181e2581e9bc38d7#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R1719xdg 

Ref: https://wayland-book.com/xdg-shell-in-depth/popups.html#popup-grabs

platform/wayland/cog-platform-wl.c
psaavedra, line 1044: Associating the Platform object as the context of the touch event (this is a pattern that also happens in the pointer_on_enter and in the keyboard_on_key too). The idea is these events, who they dont get the reference to the wl_surface directly, they gets where this event is being happening thanks to this info in the context. Today they only get a reference to the Platform what is so obvious but later this could be different references, for example, if the View can be contained by difference instances (ViewPort concept, Group of views using different Windows?)

psaavedra (commented):
sadasd adas

platform/wayland/cog-platform-wl.c
psaavedra: Good point and I think I should do this in sake of simplicity having in account there is only one platform. The reason of passing the platform attached always as a user_data is in somehow with the idea of avoiding the use of global references. I think, for the case  of the Platform (since it is a singleton, as you mention) this is not needed and it is a overhead.

There is another justification for this user_data and it is what I wrote here: https://github.com/Igalia/cog/pull/623#discussion_r1377704180. Even with this, it is wrong  add changes for problems that are still not in the code yet. I will update the PR according your suggestions.

platform/wayland/cog-platform-wl.c
psaavedra: DITTO. This is justified in https://github.com/Igalia/cog/pull/623#discussion_r1377704180 but again it is wrong to add changes for problems that are still not in the code yet. I will update the PR according your suggestions.

platform/wayland/cog-platform-wl.c
psaavedra: good point

platform/wayland/cog-platform-wl.c
psaavedra: PR updated.

platform/wayland/cog-platform-wl.c
psaavedra: PR updated.

platform/wayland/cog-platform-wl.c
```diff
@@ -105,90 +104,36 @@ noop()
...
+ * the CogWlPlatform. The rationale behind this is to articulate an easy way
+ * for passing the container as contextual information associated to a
+ * wl_surface. This is interesting for avoiding the use of global
+ * references and, in a near future, will make easier the access to the
+ * object that groups the CogWlViews.
+ *
+ * returns: static struct wl_surface *
+ */
+static struct wl_surface *
+cog_wl_compositor_create_surface(struct wl_compositor *compositor, void *container)
```
aperezdc, line 130: Now that the patch uses `cog_platform_get_default()`, we do not need this anymore. We can remove this helper function and leave the call to `wl_compositor_create_surface()` where it was before.

aperezdc (approved):
That's now nicer and easier to follow, thanks!

Patch LGTM with a small nit you may want to fix before landing.

platform/wayland/cog-platform-wl.c
psaavedra: Good suggestion. Applied in the ultimate revision (https://github.com/Igalia/cog/compare/a43b0fc0cdf500b5ef97210b5129718c00f0186f..5896b0bb752608ea4d4ef70b4102a0cdb4e69381).

PR #624, opened 2023-10-25
Change source headers to use SPDX license identifiers

### LINKED ISSUE #357: Switch to SPDX code headers

aperezdc:
SSIA. Note that there might be some cases in which the style checker trips because the intention is to only modify the license lines in file headers, and add the missing ones.

Fixes #357

----

Note that the style checker may complain in some cases and IMO for this patch which only touches file headers I think it would be reasonable to land the patch ignoring its complaints 😉

### DIFF
--- a/core/cog-directory-files-handler.c
+++ b/core/cog-directory-files-handler.c
@@ -2,7 +2,7 @@
  * cog-directory-files-handler.c
  * Copyright (C) 2017-2018 Adrian Perez <aperez@igalia.com>
  *
- * Distributed under terms of the MIT license.
+ * SPDX-License-Identifier: MIT
  */
 
 #include "cog-directory-files-handler.h"
--- a/core/cog-directory-files-handler.h
+++ b/core/cog-directory-files-handler.h
@@ -2,7 +2,7 @@
  * cog-directory-files-handler.h
  * Copyright (C) 2017-2018 Adrian Perez <aperez@igalia.com>
  *
- * Distributed under terms of the MIT license.
+ * SPDX-License-Identifier: MIT
  */
 
 #ifndef COG_DIRECTORY_FILES_HANDLER_H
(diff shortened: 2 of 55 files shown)


platform/wayland/os-compatibility.h
```diff
@@ -21,6 +21,8 @@
  * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
+ *
+ * SPDX-License-Identifier: MIT
```
psaavedra, line 25: Should we modify imported files from other projects as-is?

platform/wayland/os-compatibility.c
```diff
@@ -21,6 +21,8 @@
  * ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
+ *
+ * SPDX-License-Identifier: MIT
```
psaavedra, line 25: ditto.

psaavedra (approved):
One doubt about how to proceed with the imported files. Other than that. LGTM. r+.

platform/wayland/os-compatibility.h
aperezdc: Good point, maybe it's better to leave them be. I'll revert this change before landing.

platform/wayland/os-compatibility.c
aperezdc: Same here, I'll revert this before landing.

aperezdc: I will also revert the change in `platform/gtk4/cog-gtk-settings-{dialog,cell-renderer-variant}.[hc]`, which were picked from the WebKit MiniBrowser.

aperezdc: Oh, and `platform/drm/kms.[hc]`, which come from libdrm IIRC.

PR #626, opened 2023-10-26
wl: Fix build with libportal enabled after e533fce

aperezdc:
Add missing include and fix access to the window through the view's associated `CogPlatform`.

----

The build with libportal got broken in #619


psaavedra (approved):
Good catch.

PR #627, opened 2023-10-26
core: Warn on both get_view_type and init_web_view defined

aperezdc:
Emit a warning when trying to call `cog_platform_init_web_view()` in case the platform in use defines both `.get_view_type` and `.init_web_view`, because the latter shall not be defined when `.get_view_type` is in use.


psaavedra (approved):
LGTM

PR #629, opened 2023-10-26
wl: Decouple the fullscreen logic in CogWlPlatform and CogWlView

psaavedra:
In the **Platform** side the fullscreen enter/exit logic was cleaned up, simplified and encapsulated in the `cog_wl_platform_[enter|exit]_fullscreen` functions.

The `CogWlPlatform` is not responsible of having in consideration the state of the View. It is limited to do store the previous geometry, to set the new window geometry and to make the adjustments in the corresponding shell. Finally, the **View** associated to the **Platform** is notified using the new `cog_wl_view_[enter|exit]_fullscreen` functions and relies on the View to do the rest. The CogWlView reacts those notifications by sending the `wpe_view_backend_dispatch_did_enter_fullscreen` and `wpe_view_backend_dispatch_did_exit_fullscreen` according with the notification.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -103,6 +103,11 @@ static struct {
     .configured = false,
 };
 
+static void
+noop()
+{
+}
+
 static gboolean
 wl_src_prepare(GSource *base, gint *timeout)
 {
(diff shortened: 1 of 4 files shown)


aperezdc (approved):
Nice cleanup, thanks!

PR #630, opened 2023-10-26
core: Declare public symbol visibilities

aperezdc:
Change the build options to use “hidden” visibility by default, and mark only the symbols that are part of the public API with the COG_API macro. This prevents library private symbols from leaking to the global symbol namespace even if they are not declared as “static”, and also makes the ELF PLT smaller which prevents accidental symbol overriding and improves load times.

### DIFF
--- a/core/cog-view.h
+++ b/core/cog-view.h
@@ -21,6 +21,7 @@ struct wpe_view_backend;
 
 #define COG_TYPE_VIEW (cog_view_get_type())
 
+COG_API
 G_DECLARE_DERIVABLE_TYPE(CogView, cog_view, COG, VIEW, WebKitWebView)
 
 struct _CogViewClass {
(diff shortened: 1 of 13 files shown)


core/cog-view.h
```diff
@@ -32,11 +33,22 @@ struct _CogViewClass {
 
 #define COG_TYPE_VIEW_IMPL (cog_view_get_impl_type())
 
+COG_API
 GType                    cog_view_get_impl_type(void);
```
psaavedra, line 37: Should be this part of the public API? AFAIK this should be used by internals like the cog-shell

core/cog-view.h
```diff
@@ -32,11 +33,22 @@ struct _CogViewClass {
...
+
+COG_API
 CogView                 *cog_view_new(const char *first_property_name, ...);
+
+COG_API
```
psaavedra, line 42: Do we need this object accessible from other part than the platforms plugins? Should not this be an internal detail hidden by cog?

core/cog-view.h
aperezdc: The idea of having this public is to allow using `g_object_new()` instead of forcing people to always use `cog_view_new()`. There is even an example that uses `g_object_new(cog_view_get_impl_type(), ...)` in the [documentation for cog_view_get_impl_type](https://igalia.github.io/cog/type_func.View.get_impl_type.html).

If we are allowing instantiation of the platform-provided custom `CogView` subclasses via `cog_view_new()`, I think there is no problem at all in allowing to use `g_object_new()` in combination with this function in the public API.

psaavedra: I see then I don't have any particular issue with this.

core/cog-view.h
psaavedra: ditto.

psaavedra (approved):

PR #632, opened 2023-10-26
wl: Remove CogWlPlatform::init_web_view()

psaavedra:
All the initialization done in the init_web_view() is now relocated  inside of the cog_wl_view_init(). To make this possible all the code     related to the "run-file-chooser" and "mouse-target-changed" signals     have be moved from the cog-platform-wl to the cog-view-wl.

This change makes the Wayland platform compatible with the check  implemented in the cog_platform_init_web_view(): "CogWlPlatform class    defines both .get_view_type and .init_web_view the latter should be   removed."

Also adds again the early return in cog_platform_init_web_view() if class defines both .get_view_type and .init_web_view.

### DIFF
--- a/platform/wayland/cog-view-wl.c
+++ b/platform/wayland/cog-view-wl.c
@@ -11,6 +11,7 @@
 #include <wpe/fdo-egl.h>
 #include <wpe/fdo.h>
 
+#include "../common/cursors.h"
 #include "../common/egl-proc-address.h"
 
 #include <EGL/egl.h>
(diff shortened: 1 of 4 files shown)


platform/wayland/cog-view-wl.c
```diff
@@ -87,12 +105,20 @@ cog_wl_view_init(CogWlView *self)
...
 
+    // Only one view is managed by the Wayland Platform nowadays.
+    self->platform->view = self;
+
     wl_list_init(&self->shm_buffer_list);
 
+    g_signal_connect(self, "mouse-target-changed", G_CALLBACK(on_mouse_target_changed), NULL);
+#if COG_HAVE_LIBPORTAL
+    g_signal_connect(self, "run-file-chooser", G_CALLBACK(on_run_file_chooser), NULL);
+#endif /* COG_HAVE_LIBPORºTAL */
```
aperezdc, line 121: It looks like your cat got onto the keyboard 🐈:

```suggestion
#endif /* COG_HAVE_LIBPORTAL */
```

aperezdc (commented):
Nice, thanks for the fix. There's only one small nit to fix before landing.

aperezdc (approved):
Feel free to push the small fix and land the PR, @psaavedra 🙏🏼

platform/wayland/cog-view-wl.c
psaavedra: fixed.

PR #635, opened 2023-10-27
wl: Refactor popup_data as CogWlPopup

psaavedra:
There is not longer a static global popup_data definition inside of the platform. Instead of this the Platform will have attached a CogWlPopup instance.

It is the handling of the CogWlPopup responsibility of the CogWlPlatform object. For example, the cog_wl_platform_popup_create() and the cog_wl_platform_popup_destroy() will deal with the set and unset of the reference to the popup object inside the CogWlPlatform.

Because this reason, the CogWlView just relies on the CogWlPlatform for the creation and the rest of the life-cycle of a CogWlPopup now. All this is simply triggered by the on_show_option_menu() handler.

There are also other trivial refactorings:

* The cog_wl_compositor_create_surface() is moved to the cog-utils-wl for convenience.
* The popup auxiliary structs and callback handlers are also moved to the cog-utils-wl.
* The CogWlPopup is now accesible from the CogWlPlatform object but we still try to avoid global access to the platform by passing the platform object, as much as possible, as contextual information to the handlers.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -39,7 +39,6 @@
 #include "cog-im-context-wl-v1.h"
 #include "cog-im-context-wl.h"
 #include "cog-platform-wl.h"
-#include "cog-popup-menu-wl.h"
 #if COG_HAVE_LIBPORTAL
 #    include "cog-xdp-parent-wl.h"
 #endif /* COG_HAVE_LIBPORTAL */
@@ -79,34 +78,10 @@ typedef struct wl_buffer *(EGLAPIENTRYP PFNEGLCREATEWAYLANDBUFFERFROMIMAGEWL)(EG
 
 CogWlOutput output_metrics;
 
-static struct {
-    struct wl_surface *wl_surface;
-
-    struct xdg_positioner *xdg_positioner;
-    struct xdg_surface *xdg_surface;
-    struct xdg_popup *xdg_popup;
-
-    struct wl_shell_surface *shell_surface;
-
-    uint32_t width;
-    uint32_t height;
-
-    CogPopupMenu *popup_menu;
-    WebKitOptionMenu *option_menu;
-
-    bool configured;
-} popup_data = {
-    .configured = false,
-};
-
 static void
 noop()
 {
 }
-
-static void display_popup(void);
-static void update_popup(void);
-static void destroy_popup(void);
 
 static void seat_on_capabilities(void *, struct wl_seat *, uint32_t);
 static void seat_on_name(void *, struct wl_seat *, const char *);
@@ -203,33 +178,6 @@ static const struct wl_shell_surface_listener shell_surface_listener = {
 };
 
 static void
-shell_popup_surface_ping(void *data, struct wl_shell_surface *shell_surface, uint32_t serial)
-{
-    wl_shell_surface_pong(shell_surface, serial);
-}
-
-static void
-shell_popup_surface_configure(void *data,
-                              struct wl_shell_surface *shell_surface,
-                              uint32_t edges,
-                              int32_t width,
-                              int32_t height)
-{
-}
-
-static void
-shell_popup_surface_popup_done (void *data,
-                                struct wl_shell_surface *shell_surface)
-{
-}
-
-static const struct wl_shell_surface_listener shell_popup_surface_listener = {
-    .ping = shell_popup_surface_ping,
-    .configure = shell_popup_surface_configure,
-    .popup_done = shell_popup_surface_popup_done,
-};
-
-static void
 xdg_shell_ping(void *data, struct xdg_wm_base *shell, uint32_t serial)
 {
     xdg_wm_base_pong(shell, serial);
@@ -243,14 +191,7 @@ static void
 xdg_surface_on_configure(void *data, struct xdg_surface *surface, uint32_t serial)
 {
     xdg_surface_ack_configure(surface, serial);
-
-    if (popup_data.xdg_surface == surface && !popup_data.configured) {
-        popup_data.configured = true;
-        display_popup();
-    }
 }
-
-static const struct xdg_surface_listener xdg_surface_listener = {.configure = xdg_surface_on_configure};
 
 static void
 xdg_toplevel_on_configure(void                *data,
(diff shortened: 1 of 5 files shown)


platform/wayland/cog-platform-wl.c
```diff
@@ -1819,6 +1651,30 @@ cog_wl_platform_create_im_context(CogPlatform *platform)
     return NULL;
 }
 
+void
+cog_wl_platform_popup_create(CogWlPlatform *platform, WebKitOptionMenu *option_menu)
+{
+    g_assert(!platform->popup);
+
+    CogWlPopup *popup = cog_wl_popup_create(platform, option_menu);
+    platform->popup = popup;
```
aperezdc, line 1660: What happens if a popup had already been created? Probably the correct thing to do is add a check, and if there was already a popup, close+destroy it before creating the new one:

```suggestion
    g_clear_pointer(&platform->popup, cog_wl_popup_destroy);
    platform->popup = cog_wl_popup_create(platform, option_menu);
```

platform/wayland/cog-platform-wl.c
```diff
@@ -1819,6 +1651,30 @@ cog_wl_platform_create_im_context(CogPlatform *platform)
...
+}
+
+void
+cog_wl_platform_popup_destroy(CogWlPlatform *platform)
+{
+    g_assert(platform->popup);
+    cog_wl_popup_destroy(platform->popup);
+    platform->popup = NULL;
```
aperezdc, line 1668: The `g_clear_pointer()` macro can be used here:

```suggestion
    g_clear_pointer(&platform->popup, cog_wl_popup_destroy);
```

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,180 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
 
+static void
+xdg_popup_on_configure(void *data, struct xdg_popup *xdg_popup, int32_t x, int32_t y, int32_t width, int32_t height)
+{
+}
+
+static void
+xdg_popup_on_popup_done(void *data, struct xdg_popup *xdg_popup)
+{
+    cog_wl_popup_destroy((CogWlPopup *) data);
```
aperezdc, line 198: I am a bit concerned that this will leave a dangling pointer in `platform->popup`, but fixing that would involve a bigger patch and probably turning `CogWlPopup` into a GObject to make `platform->popup` a weak reference that gets reset to `NULL` automatically when the popup is destroyed... Anyway, I am thinking out loud and this is okay for now, no need to change the code.

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,180 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
+    .configure = shell_popup_surface_configure,
+    .popup_done = shell_popup_surface_popup_done,
+};
+
+static void
+xdg_surface_on_configure(void *data, struct xdg_surface *surface, uint32_t serial)
+{
+    xdg_surface_ack_configure(surface, serial);
+
+    CogWlPopup *popup = (CogWlPopup *) data;
```
aperezdc, line 232: FWIW, casting from `void*` to a concrete pointer type does not need the cast, in C the following is valid and preferred:

```suggestion
    CogWlPopup *popup = data;
```

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,180 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
+CogWlPopup *
+cog_wl_popup_create(CogWlPlatform *platform, WebKitOptionMenu *option_menu)
+{
+    CogWlDisplay *display = platform->display;
+
+    CogWlPopup *popup = g_slice_new0(CogWlPopup);
+    g_debug("%s: Create @ %p", G_STRFUNC, popup);
+
+    popup->configured = false;
+
```
aperezdc, line 255: This is not needed, `g_slice_new0()` does zero-initialize the allocated memory itself.

```suggestion
```

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,180 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
+}
+
+void
+cog_wl_popup_destroy(CogWlPopup *popup)
+{
+    g_debug("%s: Destroying @ %p", G_STRFUNC, popup);
+    g_assert(popup);
+
+    if (popup->option_menu == NULL)
+        return;
```
aperezdc, line 318: This check is not needed. All the `g_clear_{pointer,object}` macro uses below already do checks. Also, this check might prevent members from a half-initialized object being freed below and cause memory leaks&mdash;while unlikely, it's possible.

```suggestion
```

I know this is code moved in the refactoring, so we can leave it as-is for now, and do a separate patches later on with cleanups.

aperezdc (approved):
I left notes about some cleanups that could be done, but as this is a refactor that mostly moves existing code around I think it's okay if you prefer to first land this PR and we can have other patches later on with more cleanups.

Thanks for the patch!

platform/wayland/cog-platform-wl.c
psaavedra: in the current implementation, this is not possible since there is an assertion in https://github.com/Igalia/cog/pull/635/files#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R1657. The current implementation relies in a logical management if the Popup by the Platform.

platform/wayland/cog-utils-wl.c
psaavedra: But still should be needed for the webkit_option_menu_close() function invocation, isn't it?.

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,178 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
```
psaavedra, line 316:
```suggestion
    if (popup->option_menu != NULL)
        webkit_option_menu_close(popup->option_menu);
```

platform/wayland/cog-utils-wl.c
```diff
@@ -187,6 +187,178 @@ cog_wl_display_destroy(CogWlDisplay *display)
...
+
+    webkit_option_menu_close(popup->option_menu);
```
psaavedra, line 318:
```suggestion
```

platform/wayland/cog-utils-wl.c
aperezdc: That's right, good catch.  We need a non-NULL check before calling `webkit_option_menu_close()` but also remove the `return` statement so deallocation continues 👍🏼

platform/wayland/cog-platform-wl.c
psaavedra: but ensure the function is used in the expected way. Anyhow, this could be also expressed inside a documentation comment. I will do. Thanks

psaavedra: Documentation added in https://github.com/Igalia/cog/pull/635/files#diff-dfb5db5df9086c64aff35fe3e3f49a702cfadab4aa381fa01e96ad29a52ed8ecR39

platform/wayland/cog-utils-wl.c
psaavedra: Fix submitted in a separated PR: https://github.com/Igalia/cog/pull/643

platform/wayland/cog-utils-wl.c
psaavedra: Fixed in https://github.com/Igalia/cog/pull/643

PR #638, opened 2023-10-31
wl: Make setting surface sizes more robust

aperezdc:
This PR includes two commits which improve robustness when handling setting Wayland surface sizes:

- First, only read the `COG_PLATFORM_WL_VIEW_{WIDTH,HEIGHT}` environment variables only once during instantiation of the `CogWlPlatform` instance. There is no need to read those every time the surface is resized. This also brings in a helper function to do more thorough validation of the values from the environment variables.
- Second, use `g_return_if_fail()` to check that the width and height passed to `cog_wl_platform_configure_geometry()` are non-zero.


psaavedra (approved):
LGTM.

PR #643, opened 2023-11-01
wl: Fix possible invalid access to Platform:popup

psaavedra:
... inducted by the the **xdg_popup::popup_done** handler.

This change fixes the `xdg_popup_on_popup_done()` in `platform/wayland/cog-utils-wl.c`. Now, the function resets the the `Platform::popup` if the destroyed popup is the same than the one
referenced in the **Platform**.


aperezdc (approved):
r=me, thanks!

Issue #634, opened 2023-10-27
Disentangle CogShell from CogPlatform


aperezdc:
Currently `CogShell` and `CogPlatform` are intertwined due to historical hubris, and one _needs to know_ in which order each operation on them has to be invoked to successfully have things work. This is far from ideal, and is a sure footgun in the public API.

Since the switch to GIO modules (#320) and, more recently, the change to allow platform plug-ins to define their own `CogView` subclass (#570), we no longer need to do this dance because the current in-use `CogPlatform` object can always be accessed globally, so the `CogShell` (or anything that replaces it) can fetch it with `cog_platform_get_default()` (as done e.g. in #565 to simplify the launcher).

Finally, currently `CogShell` tries to _both_ handle settings that apply to “Cog as a platform” (device scaling factor, whether to run in automation mode, configuration file settings) and also settings to apply by default to WebKit (`WebKitSettings`, `WebKitWebContext`, etc.)&mdash;and probably it is not a good idea to conflate the two.

Therefore, going forward, we would want to:

- [ ] Make `CogShell` handle application settings *only*, probably renamed to a new `CogConfiguration` class.
- [ ] Remove the handling of WebKit configurations from `CogShell`, or move them to a new utility class which is _optional_ to use, or alternatively have some way of optionally setting *default* instances for `WebKitSettings`, `WebKitWebContext` and friends that would be picked by the custom `CogView` constructor if not specified during construction.

PR #644, opened 2023-11-02
Introduce CogViewport, add support in the Wayland plug-in and example program

aperezdc:
This PR brings in the following:

- A new `CogViewport` class, which represents an “output” to display one of a set of view.
- Adapting `CogShell` to use a viewport instead of a single view. At some point we will want to uncouple the shell (see #634) but this replacement is enough for now.
- Add an `examples/viewport.c` example program which shows how to switch between among the views of a viewport.
- Adapt the Wayland platform plug-in to no longer assume that there is a single view. 

I would recommend reviewing each commit separately.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -97,7 +97,7 @@ cog_wl_platform_configure_geometry(CogWlPlatform *platform, int32_t width, int32
         platform->window.width = width;
         platform->window.height = height;
 
-        cog_wl_view_resize(platform->view);
+        cog_viewport_foreach(platform->viewport, (GFunc) cog_wl_view_resize, NULL);
     }
 }
 
@@ -106,6 +106,11 @@ static void resize_to_largest_output(CogWlPlatform *);
 static void
 cog_wl_platform_enter_fullscreen(CogWlPlatform *platform)
 {
+    if (!cog_viewport_get_n_views(platform->viewport)) {
+        g_debug("%s: No views in viewport, will not fullscreen.", G_STRFUNC);
+        return;
+    }
+
     CogWlDisplay *display = platform->display;
     CogWlWindow  *window = &platform->window;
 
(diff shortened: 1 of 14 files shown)


psaavedra:
One comment after a first quick test:

* The Fullscreen resize is working fine
* The same for the popup option menu. Works OK
* However,  the virtual keyboard stopped working with this change. I don't any obvious reason for this. What change is adding this regression is not obvious to me after reading the patch. I am trying to go deeply inside this issue right now.

aperezdc:
> * However,  the virtual keyboard stopped working with this change. I don't any obvious reason for this. What change is adding this regression is not obvious to me after reading the patch. I am trying to go deeply inside this issue right now.

Thanks for testing, I hadn't tried the virtual keyboard support so I didn't notice. I will try to take a look later.

platform/wayland/cog-platform-wl.c
```diff
@@ -1560,16 +1579,42 @@ static struct wpe_view_backend *
...
+
+    /*
+     * TODO: Once we support multiple viewports, a view may be visible
+     *       without having focus. At that point the input events will
+     *       need to go to the *focused* view instead.
+     *
+     *       For now add the flag and assume that visible == focused.
+     */
+    wpe_view_backend_add_activity_state(cog_view_get_backend((CogView *) view), wpe_view_activity_state_focused);
+    cog_wl_view_update_surface_contents(view);
```
psaavedra, line 1607: Related to https://github.com/Igalia/cog/pull/644#issuecomment-1792067605. You want to set the focus in the visible view always once this is visible (this could change in the future with multiple viewports and so but now it is the way to set the focus in the visible View). The check for view->image only matters for refresh the content with a new frame.

 
```suggestion
    /*
     * TODO: Once we support multiple viewports, a view may be visible
     *       without having focus. At that point the input events will
     *       need to go to the *focused* view instead.
     *
     *       For now add the flag and assume that visible == focused.
     */
    wpe_view_backend_add_activity_state(cog_view_get_backend((CogView *) view), wpe_view_activity_state_focused);

    if (!(view && view->image)) {
        g_debug("%s: No image to show, skipping update.", G_STRFUNC);
        return;
    }

    cog_wl_view_update_surface_contents(view);
```

aperezdc: Oh, wow, excellent catch. Indeed we always want to set the flag, and not only when there is an image to update the surface 🖼️ 

I will update the PR with this change, make the style checker happy, and re-upload the patch.

Thanks!

psaavedra: I think this is ok for now since we will want the focus always activated  in the visible view but in the future it will be necessary to differentiate between these to different concepts (visibility vs focus).

Look at this: `cog set visible view 3 from viewport 2` VS `cog set focus on viewport 2`. The first case should activate the focus on view 3 of the viewport 2 only if the focus was previously active on viewport 2. But it will be not always the case. For example, we may want to make view N of a viewport visible but not steal the focus  from another viewport which is the one with focus at this point.

psaavedra (changes requested):
other than the mentioned regression found with the focus. The rest of the changes looks really great.

I also notice the view->platform is not really needed and removed from the code. I am fine with this also.

I think the PR were be ready for land once you update it according my suggestion and passes the code-style checker.

aperezdc:
@psaavedra Updated, PTAL when you have a minute 🙏🏼

psaavedra (approved):
Awesome. LGTM now.

PR #651, opened 2023-11-06
Fix build with using glib<2.63.1

psaavedra:
Add GLIB_CHECK_VERSION to verify if we really can use the g_warning_once and avoid the undefined reference to g_warning_once link error.

### DIFF
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -125,9 +125,19 @@ cog_platform_init_web_view (CogPlatform   *platform,
     CogPlatformClass *klass = COG_PLATFORM_GET_CLASS(platform);
     if (klass->init_web_view) {
         if (klass->get_view_type) {
+#if GLIB_CHECK_VERSION(2, 63, 1)
             g_warning_once("%s: class %s defines both .get_view_type and .init_web_view, "
                            "the latter should be removed.",
                            G_STRFUNC, G_OBJECT_CLASS_NAME(platform));
+#else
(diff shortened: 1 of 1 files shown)


core/cog-platform.c
```diff
@@ -125,9 +125,19 @@ cog_platform_init_web_view (CogPlatform   *platform,
     CogPlatformClass *klass = COG_PLATFORM_GET_CLASS(platform);
     if (klass->init_web_view) {
         if (klass->get_view_type) {
+#if GLIB_CHECK_VERSION(2, 63, 1)
```
aperezdc, line 128: GLib 2.63.1 was released in October 2019. That's more than four years ago... so now I am wondering in which environment this is a problem.

I suggest we add a check in the top-level `meson.build` file to make sure the GLib version is new enough.

psaavedra: On Yocto Dunfell with long time support until April of 2024 [[ref](https://wiki.yoctoproject.org/wiki/Releases)] so you could expect this issue affecting to environments still supported in [meta-webkit](https://github.com/Igalia/meta-webkit/blob/main/conf/layer.conf#L20). If you think this is not a good argument, I could move it as an adaptation patch inside meta-webkit.

aperezdc: Ahm, fair. Let's merge this and bump the GLib dependency in the next cycle after March-April 2024.

aperezdc (approved):

PR #655, opened 2023-11-09
core: Remove cog_platform_configure() function

aperezdc:
Remove the `cog_platform_configure()` utility function, because at this point it does not do much else than calling `cog_platform_new()` and `cog_platform_setup()`. Instead, make sure that the platform instance is a global singleton and that it is created on-demand on the first call of a the `cog_platform_get()` accessor.

In order to allow manually indicating which platform plug-in to use, a new `cog_init()` function is added. While at it, make it unnecessary for programs to always call `cog_modules_add_directory()` if they only need to use the default module path.

The upside of these changes is that the API gets much more convenient to use, and in the simplest of cases, it is enough to let the library initialize the platform instance. The one-time configuration using `cog_platform_setup()` is still needed, but now the following is enough to properly use the library:

```c
CogShell *shell = cog_shell_new("my-program", FALSE);
if (!cog_platform_setup(cog_platform_get(), shell, "", NULL))
    g_error("Cannot initialize Cog.");
```

Note that the `COG_MODULEDIR` and `COG_PLATFORM_NAME` environment variables are still used when `NULL`s are passed to `cog_init()` or during implicit initialization.

----

- This is an alternative take on #647
- If possible, it would be better to merge #656 first (see [comment below](https://github.com/Igalia/cog/pull/655#issuecomment-1809805319))

### DIFF
--- a/examples/viewport.c
+++ b/examples/viewport.c
@@ -34,12 +34,12 @@ main(int argc, char *argv[])
     }
 
     g_set_prgname("view-stack");
-    cog_modules_add_directory(g_getenv("COG_MODULEDIR"));
 
-    g_autoptr(GError)      error = NULL;
-    g_autoptr(CogShell)    shell = cog_shell_new(g_get_prgname(), FALSE);
-    g_autoptr(CogPlatform) platform = cog_platform_configure(NULL, NULL, "COG", shell, &error);
-    if (!platform)
+    g_autoptr(CogShell) shell = cog_shell_new(g_get_prgname(), FALSE);
+    g_autoptr(GError)   error = NULL;
+
+    CogPlatform *platform = cog_platform_get();
+    if (!cog_platform_setup(platform, shell, g_getenv("COG_PLATFORM_PARAMS") ?: "", &error))
         g_error("Cannot configure platform: %s", error->message);
 
     g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, FALSE);
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -237,11 +237,12 @@ cog_launcher_create_view(CogLauncher *self, CogShell *shell)
         webkit_web_context_set_cache_model(web_context, WEBKIT_CACHE_MODEL_DOCUMENT_VIEWER);
     }
 
+    CogPlatform *platform = cog_platform_get();
+
     g_autoptr(GError) error = NULL;
-    CogPlatform      *platform =
-        cog_platform_configure(s_options.platform_name, s_options.platform_params, "COG", shell, &error);
-    if (!platform)
-        g_error("Cannot instantiate a platform: %s", error->message);
+    if (!cog_platform_setup(platform, shell, s_options.platform_params ?: (g_getenv("COG_PLATFORM_PARAMS") ?: ""),
+                            &error))
+        g_error("Cannot configure platform: %s", error->message);
 
 #if HAVE_WEBKIT_AUTOPLAY
     WebKitWebsitePolicies *website_policies =
@@ -329,6 +330,8 @@ cog_launcher_startup(GApplication *application)
      * prevent GApplication from shutting down immediately after startup.
      */
     g_application_hold(application);
+
+    cog_init(s_options.platform_name, NULL);
 
     CogLauncher *self = COG_LAUNCHER(application);
     self->shell = g_object_new(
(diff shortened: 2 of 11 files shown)


launcher/cog-launcher.c
```diff
@@ -238,11 +238,14 @@ cog_launcher_create_view(CogLauncher *self, CogShell *shell)
     }
 
     g_autoptr(GError) error = NULL;
-    CogPlatform      *platform =
-        cog_platform_configure(s_options.platform_name, s_options.platform_params, "COG", shell, &error);
+    CogPlatform      *platform = cog_platform_new(s_options.platform_name ?: g_getenv("COG_PLATFORM_NAME"), &error);
```
psaavedra, line 241: cog_platform_new doesn't check if there is already created a platform.

psaavedra, line 241: I think the this could be like this  🤔  :

```suggestion
    CogPlatform *platform = cog_platform_get_default();
    if (!platform) {
        platform = cog_platform_new(s_options.platform_name ?: g_getenv("COG_PLATFORM_NAME"), &error);
    }   
```

launcher/cog-launcher.c
```diff
@@ -238,11 +238,14 @@ cog_launcher_create_view(CogLauncher *self, CogShell *shell)
...
     if (!platform)
         g_error("Cannot instantiate a platform: %s", error->message);
 
+    if (!cog_platform_setup(platform, shell, s_options.platform_params ?: (g_getenv("COG_PLATFORM_PARAMS") ?: ""),
```
psaavedra, line 245: ... and this could be immediately after the `cog_platform_new` inside of the  `if (!platform)`

psaavedra: I mean, the:

``` c
if (!platform)
        g_error("Cannot instantiate a platform: %s", error->message);
```

could be inside of 

```
    CogPlatform      *platform = cog_platform_get_default();
	    if (!platform) {
	        platform = cog_platform_new(s_options.platform_name ?: g_getenv("COG_PLATFORM_NAME"), &error);
	    }   
```

psaavedra (changes requested):
I added some comments.

launcher/cog-launcher.c
aperezdc: As agreed in a chat with @psaavedra, changed the patch to move things a bit more around and have a `cog_init()` function + `cog_platform_get()`.

examples/viewport.c
```diff
@@ -34,12 +34,12 @@ main(int argc, char *argv[])
     }
 
     g_set_prgname("view-stack");
-    cog_modules_add_directory(g_getenv("COG_MODULEDIR") ?: COG_MODULEDIR);
 
-    g_autoptr(GError)      error = NULL;
-    g_autoptr(CogShell)    shell = cog_shell_new(g_get_prgname(), FALSE);
-    g_autoptr(CogPlatform) platform = cog_platform_configure(NULL, NULL, "COG", shell, &error);
-    if (!platform)
+    g_autoptr(CogShell) shell = cog_shell_new(g_get_prgname(), FALSE);
```
psaavedra, line 38: I think the platform has to be created first. Even before than the  shell. The shell needs a already created platform. For example, in case of the shell initializes a default viewport containing a default view.

psaavedra (commented):
commented

launcher/cog-launcher.c
aperezdc: The order doesn't really matter anymore: creating a view goes through `cog_view_get_impl_type()`, which uses `cog_platform_get()`, and at that point the platform singleton will be created. The only situation in which the platform singleton needs to be created first is to manually choosing the platform plug-in *using the API*&mdash;environment variables are still honored when letting `cog_platform_get()` create the singleton.

TL;DR with the new approach the only ordering restriction is “call `cog_init()` before anything else if you want to choose the platform plug-in using the API”.

aperezdc:
By the way, this needs #656 because the code in this PR assumes it can do `cog_modules_add_directory(NULL)`, but passing a null pointer needs the other PR to land first.

examples/viewport.c
aperezdc: That is not an issue, because the shell uses `cog_platform_get()`, and that will create the platform instance when needed.

The only case in which the platform instance needs to be created first is to manually specify which one to use with the API, and we have `cog_init()` for that.

aperezdc:
Rebased after #656 was merged, and commented. PTAL @psaavedra 🔍

examples/viewport.c
psaavedra: Using `cog_init("wl", NULL);`

psaavedra (approved):
Tested in several environments.

examples/viewport.c
psaavedra: https://github.com/Igalia/cog/pull/659 makes this even more robust and allows to dont need to use the `cog_init()`.

PR #658, opened 2023-11-13
core: Allow passing NULL as platform setup params

aperezdc:
Change `cog_platform_setup()` to allow passing `NULL` as the parameters string, and fall back to using the value of the `COG_PLATFORM_PARAMS` environment variable in that case. In order to keep the same ABI with existing platform plug-ins, the function ensures that an empty string is always passed when needed.

This change simplifies call sites for the function, which no longer need to handle reading the environment variable themselves, nor taking care of always passing a valid string, even if empty.

While at it, write the missing documentation for `cog_platform_setup()`.

----

This depends on https://github.com/Igalia/cog/pull/655

### DIFF
--- a/core/cog-platform.c
+++ b/core/cog-platform.c
@@ -117,17 +117,37 @@ cog_platform_get(void)
     return platform_singleton;
 }
 
+/**
+ * cog_platform_setup:
+ * @platform: Platform instance.
+ * @shell: The shell being configured.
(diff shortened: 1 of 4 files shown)


psaavedra (approved):
LGTM

aperezdc:
Thanks @psaavedra, I will rebase this, re-upload, and let the CI pass before merging 👍🏼

PR #659, opened 2023-11-14
core: Make cog_modules_add_directory() honor COG_MODULEDIR

aperezdc:
Move handling of the `COG_MODULEDIR` environment variable into the `cog_modules_add_directory()` function, fix the bug that skipped using the environment variable when `cog_init()` is *not* used, and improve robustness:
    
- The path passed is checked to be a valid directory location.
- The environment variable is checked only once for the whole execution of the process, the first time the function is called with a `NULL` argument.
- The path in the environment variable is checked for validity as well. A warning is produced for invalid locations.
- If either the environment variable or the compiled-in default path are used, it is considered that the “default” path was set, and will not be set again to avoid re-scanning directories unnecessarily.

While at it, sprinkle comments on the code to explain the logic.

----

This depends on changes from https://github.com/Igalia/cog/pull/655

### DIFF
--- a/core/cog-modules.c
+++ b/core/cog-modules.c
@@ -183,26 +183,74 @@ cog_modules_foreach(GIOExtensionPoint *extension_point, void (*callback)(GIOExte
 void
 cog_modules_add_directory(const char *directory_path)
 {
-    if (!directory_path)
-        directory_path = COG_MODULEDIR;
+    g_autoptr(GFile) path_file = NULL;
 
+    /*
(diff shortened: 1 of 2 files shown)


psaavedra (approved):
https://github.com/Igalia/cog/pull/659 makes this even more robust and allows to dont need to use the `cog_init()`. 

Ref: https://github.com/Igalia/cog/pull/655#discussion_r1395562760

PR #660, opened 2023-11-14
drm: Update logical input area size on property changes

aperezdc:
In addition to setting the logical input area size during initialization in `init_input()`, also update the size in `cog_drm_platform_set_property()`. The code that picks the correct dimensions depending on the rotation is moved into its own function and reused in both locations.

----

This is a follow-up to https://github.com/Igalia/cog/pull/554


nikolaszimmermann (approved):
Looks good, r=me.

PR #662, opened 2023-11-15
Allow custom viewport implementations

psaavedra:
**excerpt:** This PR allows platform implementations to provide their own viewport class. The WL Viewport class will handle the things associated with the window geometry instead of this being handled directly from the platform object.

**motivation** Optionally, platform plug-in implementations can provide their own viewport class by overriding the CogPlatform.get_viewport_type method. The `CogWlWindow` is moved inside of the `CogWlViewport`. The Window struct contains the information related to the geometry of the surface where the webview has to be painted. This information naturally belongs to the Viewport and not directly to the platform. With this change we  correctly encapsulate the Window information in the right place.


core/cog-viewport.c
```diff
@@ -60,6 +65,24 @@ G_DEFINE_TYPE_WITH_PRIVATE(CogViewport, cog_viewport, G_TYPE_OBJECT)
 
 #define PRIV(obj) ((CogViewportPrivate *) cog_viewport_get_instance_private((CogViewport *) obj))
 
+static void *
+_cog_viewport_get_impl_type_init(GType *type)
+{
+    *type = COG_TYPE_VIEWPORT;
+
+    CogPlatform *platform = cog_platform_get();
+    if (platform) {
```
aperezdc, line 74: Nowadays we can safely assume that `cog_platform_get()` will always return a valid instance because we have `CogFallbackPlatform` always built into `libcogcore`.

I know this is copying what `_cog_view_get_impl_type_init()` is doing, so I think it's okay this way, but this is a reminder to make a follow-up patch later to simplify this 😉

core/cog-viewport.c
```diff
@@ -60,6 +65,24 @@ G_DEFINE_TYPE_WITH_PRIVATE(CogViewport, cog_viewport, G_TYPE_OBJECT)
...
+        CogPlatformClass *platform_class = COG_PLATFORM_GET_CLASS(platform);
+        if (platform_class->get_viewport_type) {
+            GType viewport_type = (*platform_class->get_viewport_type)();
+            *type = viewport_type;
```
aperezdc, line 78: When doing the follow-up patch with cleanups, I think we could assert here that we get a valid type:

```suggestion
            GType viewport_type = (*platform_class->get_viewport_type)();
            g_assert(viewport_type != G_TYPE_INVALID);
            *type = viewport_type;
```

core/cog-viewport.c
```diff
@@ -86,6 +109,52 @@ cog_viewport_get_property(GObject *object, unsigned prop_id, GValue *value, GPar
...
+
+/**
+ * cog_viewport_new: (constructor)
+ *
+ * Creates a new instance of a viewport implementation and sets its properties.
+ *
+ * Returns: (transfer full): A new web view.
+ */
+CogViewport *
+cog_viewport_new()
```
aperezdc, line 152: Missing `void` in the function argument list:

```suggestion
cog_viewport_new(void)
```

core/cog-viewport.c
```diff
@@ -86,6 +109,52 @@ cog_viewport_get_property(GObject *object, unsigned prop_id, GValue *value, GPar
...
+{
+    void *viewport = g_object_new(cog_viewport_get_impl_type(), NULL);
+    return viewport;
```
aperezdc, line 155: This can be written in a single line:

```suggestion
    return g_object_new(cog_viewport_get_impl_type(), NULL);
```

core/cog-viewport.h
```diff
@@ -29,6 +29,14 @@ struct _CogViewportClass {
     void *_padding[8];
 };
 
+#define COG_TYPE_VIEWPORT_IMPL (cog_viewport_get_impl_type())
+
+COG_API
+GType cog_viewport_get_impl_type(void);
+
+COG_API
+CogViewport *cog_viewport_new();
```
aperezdc, line 38: Missing `void` in argument list:

```suggestion
CogViewport *cog_viewport_new(void);
```

aperezdc (approved):
I like the direction of this patch, and I have a couple of minor comments. Please fix the missing `void`s in function argument lists and this can be merged.

(Nice to see one more step in the right direction to allow programs that use the cogcore library to decide themselves how to display the web views, I feel like we are getting close! 👏🏼)

core/cog-viewport.c
psaavedra: Split: https://github.com/Igalia/cog/issues/664

core/cog-viewport.c
psaavedra: This looks good I will add a note in  https://github.com/Igalia/cog/issues/664 to address this here in the CogViewport but also in the CogView

core/cog-viewport.c
psaavedra: fixed

PR #663, opened 2023-11-16
meson: Add an option for libportal support

aperezdc:
Add an option to allow manually choosing at build time whether to enable libportal support. While at it, add logic to determine if the enabled platform plug-ins may use libportal, and only try to add the dependency if that is the case and the option has not been disabled.

Also, move all the logic related to libportal detection into platform/ and remove the definition from the `config.h.in` template, as cogcore library clients should not need to know whether the libportal support has been built: they will still be able to connect to the related signals and handle the file dialogs themselves if needed. Instead, the `COG_HAVE_LIBPORTAL` macro is passed as a compiler command line argument.


psaavedra:
Looks good to me but I'd like to test to build it in other build environments before approval.

psaavedra:
> Looks good to me but I'd like to test to build it in other build environments before approval.

Everything looks good. I will r+ this PR.

psaavedra (approved):

PR #665, opened 2023-11-16
wl: Attach Viewport to the surface

psaavedra:
**motivation:** Every single wl_surface in the Cog apps belongs to a specific Viewport but there is not a way, right now, in the code to know where     is the Viewport associated to the wl_surface.
    
**excerpt:** This change allows a direct way for accessing to the Viewport associated from a given wl_surface. Also opens the possibility of     to handle the input events methods avoiding this assumption: There is     just one single and unique CogView associated to the platform. This change is much more logical than the previous assumption.


aperezdc (approved):
Nice, thanks for the patch @psaavedra

PR #666, opened 2023-11-18
Add in CogView a reference to the CogViewport where it belongs

psaavedra:
Added a convenient reference to make explicit an **bidirectional the relationship between the CogViewport and the CogView**. This makes explicit the relation of membership of the a View respect the Viewport and breaks up the assumption of 'just one viewport in use' in the Platform. 

Also the WL plugin start to use the `cog_view_get_viewport()` to get the View's Viewport. This allows to use the CogViewport assigned to the CogView instead of the global CogViewport assigned to the Platform.

### DIFF
--- a/core/cog-view.c
+++ b/core/cog-view.c
@@ -7,6 +7,7 @@
 
 #include "cog-view.h"
 #include "cog-platform.h"
+#include "cog-view-private.h"
 
 /**
  * CogView:
@@ -26,6 +27,8 @@
 
 typedef struct {
     gboolean use_key_bindings;
+
+    GWeakRef viewport; /* Weak reference to the associated CogViewport */
 } CogViewPrivate;
 
 G_DEFINE_ABSTRACT_TYPE_WITH_PRIVATE(CogView, cog_view, WEBKIT_TYPE_WEB_VIEW)
@@ -37,6 +40,7 @@ struct _CogCoreViewClass {
 enum {
     PROP_0,
     PROP_USE_KEY_BINDINGS,
+    PROP_VIEWPORT,
     N_PROPERTIES,
 };
 
@@ -87,6 +91,9 @@ cog_view_get_property(GObject *object, unsigned prop_id, GValue *value, GParamSp
     case PROP_USE_KEY_BINDINGS:
         g_value_set_boolean(value, cog_view_get_use_key_bindings(self));
         break;
+    case PROP_VIEWPORT:
+        g_value_take_object(value, cog_view_get_viewport(self));
+        break;
     default:
         G_OBJECT_WARN_INVALID_PROPERTY_ID(object, prop_id, pspec);
     }
@@ -111,12 +118,26 @@ cog_view_class_init(CogViewClass *klass)
         g_param_spec_boolean("use-key-bindings", NULL, NULL, TRUE,
                              G_PARAM_READWRITE | G_PARAM_CONSTRUCT | G_PARAM_EXPLICIT_NOTIFY | G_PARAM_STATIC_STRINGS);
 
+    /**
+     * CogView:viewport: (default-value null)
+     *
+     * Viewport where the CogView belongs.
+     *
+     * Since: 0.20
+     */
+    s_properties[PROP_VIEWPORT] = g_param_spec_object("viewport", NULL, NULL, G_TYPE_OBJECT,
+                                                      G_PARAM_READABLE | G_PARAM_CONSTRUCT | G_PARAM_STATIC_STRINGS);
+
     g_object_class_install_properties(object_class, N_PROPERTIES, s_properties);
 }
 
 static void
 cog_view_init(CogView *self)
 {
+    CogViewPrivate *priv = cog_view_get_instance_private(self);
+
+    // Init the weak reference to the CogViewport.
+    g_weak_ref_init(&priv->viewport, NULL);
 }
 
 static inline struct wpe_view_backend *
@@ -379,6 +400,26 @@ cog_view_set_use_key_bindings(CogView *self, gboolean enable)
 }
 
 /**
+ * cog_view_set_viewport
+ * @self: A view.
+ * @viewport: The viewport where the View belongs.
+ *
+ * Since: 0.20
+ */
+void
+cog_view_set_viewport(CogView *self, CogViewport *viewport)
+{
+    g_return_if_fail(COG_IS_VIEW(self));
+
+    CogViewPrivate *priv = cog_view_get_instance_private(self);
+
+    // Sets a reference to the viewport
+    g_weak_ref_set(&priv->viewport, viewport);
+
+    g_object_notify_by_pspec(G_OBJECT(self), s_properties[PROP_VIEWPORT]);
+}
+
+/**
  * cog_view_get_use_key_bindings: (get-property use-key-bindings)
  * @self: A view.
  *
@@ -395,4 +436,21 @@ cog_view_get_use_key_bindings(CogView *self)
 
     CogViewPrivate *priv = cog_view_get_instance_private(COG_VIEW(self));
     return priv->use_key_bindings;
+}
+
+/**
+ * cog_view_get_viewport: (get-property viewport)
+ * @self: A view.
+ *
+ * Gets viewport where the view is attached.
+ *
+ * Returns: (transfer full) (nullable): The viewport what the view belongs.
+ *
+ * Since: 0.20
+ */
+CogViewport *
+cog_view_get_viewport(CogView *self)
+{
+    g_return_val_if_fail(COG_IS_VIEW(self), NULL);
+    return g_weak_ref_get(&((CogViewPrivate *) cog_view_get_instance_private(self))->viewport);
 }
--- a/core/cog-viewport.c
+++ b/core/cog-viewport.c
@@ -7,6 +7,7 @@
 #include "cog-viewport.h"
 
 #include "cog-platform.h"
+#include "cog-view-private.h"
 #include "cog-view.h"
 
 /**
@@ -258,6 +259,8 @@ cog_viewport_add(CogViewport *self, CogView *view)
     CogViewportPrivate *priv = PRIV(self);
     g_return_if_fail(!g_ptr_array_find(priv->views, view, NULL));
 
+    cog_view_set_viewport(view, self);
+
     g_ptr_array_add(priv->views, g_object_ref(view));
     g_signal_emit(self, s_signals[ADD], 0, view);
 
@@ -303,6 +306,8 @@ cog_viewport_remove(CogViewport *self, CogView *view)
         g_warning("Attempted to remove view %p, which was not in viewport %p.", view, self);
         return;
     }
+
+    cog_view_set_viewport(view, NULL);
 
     g_object_ref(view);
     g_ptr_array_remove_index(priv->views, index);
--- a/platform/wayland/cog-view-wl.c
+++ b/platform/wayland/cog-view-wl.c
@@ -302,7 +302,6 @@ cog_wl_view_resize(CogWlView *view)
 {
     CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
     CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
-
     view->should_update_opaque_region = true;
 
     int32_t pixel_width = viewport->window.width * platform->display->current_output->scale;
--- a/platform/wayland/cog-view-wl.c
+++ b/platform/wayland/cog-view-wl.c
@@ -178,7 +178,6 @@ static WebKitWebViewBackend *
 cog_wl_view_create_backend(CogView *view)
 {
     CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
     CogWlView     *self = COG_WL_VIEW(view);
 
     static const struct wpe_view_backend_exportable_fdo_egl_client client = {
@@ -188,8 +187,7 @@ cog_wl_view_create_backend(CogView *view)
 #endif
     };
 
-    self->exportable =
-        wpe_view_backend_exportable_fdo_egl_create(&client, self, viewport->window.width, viewport->window.height);
+    self->exportable = wpe_view_backend_exportable_fdo_egl_create(&client, self, DEFAULT_WIDTH, DEFAULT_HEIGHT);
 
     /* init WPE view backend */
     struct wpe_view_backend *view_backend = wpe_view_backend_exportable_fdo_get_view_backend(self->exportable);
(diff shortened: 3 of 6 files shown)


platform/wayland/cog-view-wl.c
```diff
@@ -188,8 +187,7 @@ cog_wl_view_create_backend(CogView *view)
 #endif
     };
 
-    self->exportable =
-        wpe_view_backend_exportable_fdo_egl_create(&client, self, viewport->window.width, viewport->window.height);
```
psaavedra, line 192: From now on the view will get the right geometry  once the View were added to the Viewport (which is the one who has the Window dimensions). This takes place in: https://github.com/Igalia/cog/pull/666/files#diff-bd42155d30f2b7fed99fdf7a4c9f75f78f54a49509ccf0f6feabb22b20af30e7R391

psaavedra:
@aperezdc this PR is ready for review. Thanks in advance

core/cog-view.c
```diff
@@ -396,3 +442,22 @@ cog_view_get_use_key_bindings(CogView *self)
...
     return priv->use_key_bindings;
 }
+
+/**
+ * cog_view_get_viewport: (get-property viewport)
+ * @self: A view.
+ *
+ * Gets viewport where the view is attached.
+ *
+ * Returns: The viewport what the view belongs.
```
aperezdc, line 452: This is missnig missing `(transfer none)`  and `(nullable)` annotations:

```suggestion
 * Returns: (transfer none) (nullable): The viewport what the view belongs.
```

That is because a new ref is not added to the returned object, and the returned value can be `NULL` while the view has not been added to any viewport.

core/cog-view.c
```diff
@@ -396,3 +442,22 @@ cog_view_get_use_key_bindings(CogView *self)
...
+ *
+ * Since: 0.20
+ */
+CogViewport *
+cog_view_get_viewport(CogView *self)
+{
+    g_return_val_if_fail(COG_IS_VIEW(self), NULL);
+
+    CogViewPrivate *priv = cog_view_get_instance_private(self);
+    return priv->viewport;
```
aperezdc, line 462: No need to change this, but I would have written the same in one line:

```suggestion
    return cog_view_get_instance_private(self)->viewport;
```

core/cog-view.c
```diff
@@ -111,6 +120,16 @@ cog_view_class_init(CogViewClass *klass)
...
 
+    /**
+     * CogView:viewport: (default-value null)
+     *
+     * Viewport where the CogView belongs.
+     *
+     * Since: 0.20
+     */
+    s_properties[PROP_VIEWPORT] =
+        g_param_spec_object("viewport", NULL, NULL, G_TYPE_OBJECT, G_PARAM_READWRITE | G_PARAM_CONSTRUCT);
```
aperezdc, line 131: I would make the property read-only, and update the value of `priv->viewport` only from `cog_viewport_add()` and `cog_viewport_remove()`.

If we allow to use `cog_view_set_viewport()`, then we need to also move the view from one viewport to another and keep them in sync with the propery, which I think is more complicated than we need.

core/cog-view.c
```diff
@@ -74,6 +77,9 @@ cog_view_set_property(GObject *object, unsigned prop_id, const GValue *value, GP
     case PROP_USE_KEY_BINDINGS:
         cog_view_set_use_key_bindings(self, g_value_get_boolean(value));
         break;
+    case PROP_VIEWPORT:
+        cog_view_set_viewport(self, g_value_get_object(value));
+        break;
```
aperezdc, line 82: I think we don't need this if we make the property read-only, please read below about this.

core/cog-view.c
```diff
@@ -378,6 +397,33 @@ cog_view_set_use_key_bindings(CogView *self, gboolean enable)
...
 
+/**
+ * cog_view_set_viewport: (set-property viewport)
+ * @self: A view.
+ * @viewport: The viewport where the View belongs.
+ *
+ * Since: 0.20
+ */
+void
+cog_view_set_viewport(CogView *self, CogViewport *viewport)
```
aperezdc, line 408: This will be a private function, which does not have the `COG_API` macro next to it, and we would add it to a `cog-view-private.h` header that is _not_ installed as an API header.

core/cog-view.c
```diff
@@ -378,6 +397,33 @@ cog_view_set_use_key_bindings(CogView *self, gboolean enable)
...
+{
+    g_return_if_fail(COG_IS_VIEW(self));
+
+    CogViewPrivate *priv = cog_view_get_instance_private(self);
+
+    if (priv->viewport != viewport) {
+        // If the new value is different from the old one, unreference the old value and set the new one.
+        if (viewport)
+            g_object_ref(viewport);
+        if (priv->viewport)
```
aperezdc, line 418: This introduces a circular reference: the view references the viewport, and the viewport its views. Reference cycles are tricky, you would need to add add a custom `GObject.dispose` vfunc implementation that breaks the cycle and the object needs to allow public API function calls while in half-dead state while `priv->viewport` is null.

On the other hand, we have to allow `priv->viewport` to be null anyway, because the API has one function to create a view, and a different one to add (or remove) to a viewport. Taking that into account, I think it would be better to make `priv->viewport` a weak reference: this will make the `priv->viewport` pointer set to `NULL` when the viewport gets destroyed, and we do not need to handle cyclic references.

(As a bonus, using the weak reference approach we would allow programs which use the API to check if the viewport is set to `NULL` in a signal handler, and move the orphaned view to some other viewport. But this is not be the main reason to use a weak reference, I'd say.)

core/cog-viewport.c
```diff
@@ -165,6 +165,9 @@ cog_viewport_dispose(GObject *object)
 static void
 cog_viewport_finalize(GObject *object)
 {
+    // Unset the reference to this viewport in the attached views.
+    cog_viewport_foreach((CogViewport *) object, (GFunc) cog_view_set_viewport, NULL);
+
```
aperezdc, line 170: We don't need this if we make the views keep a weak reference to their current viewport, as per my suggestion above.

aperezdc (changes requested):
I think we should make views use weak references to their viewports, see my comments above.

If allowing switching views from one viewport to another makes things too complicated, alternatively we may want to have the viewport as a mandatory construct-only property of the view, remove `cog_viewport_{add,remove}` from the public API, and have the view constructor/destructor handle adding/removing the view from its viewport instead.

core/cog-view.c
psaavedra: Sounds reasonable

core/cog-viewport.c
psaavedra: I agree. A weak references works much better and it is a simpler solution. Latests changes goes in this direction. Thanks for the suggestion.

core/cog-view.c
psaavedra: I think this particular code are safe of this situation as it is but it is true that then I could have the problem in the for-each (https://github.com/Igalia/cog/pull/666#discussion_r1400789045). Totally agree.

core/cog-view.c
```diff
@@ -111,12 +118,26 @@ cog_view_class_init(CogViewClass *klass)
...
 
+    /**
+     * CogView:viewport: (default-value null)
+     *
+     * Viewport where the CogView belongs.
+     *
+     * Since: 0.20
+     */
+    s_properties[PROP_VIEWPORT] =
+        g_param_spec_object("viewport", NULL, NULL, G_TYPE_OBJECT, G_PARAM_READABLE | G_PARAM_CONSTRUCT);
```
aperezdc, line 129: Let's also have the `G_PARAM_STATIC_NAME` flag here, so GLib won't copy the `"viewport"` string:

```suggestion
        g_param_spec_object("viewport", NULL, NULL, G_TYPE_OBJECT, G_PARAM_READABLE | G_PARAM_CONSTRUCT | G_PARAM_STATIC_NAME);
```

core/cog-view.c
aperezdc: By the way, we do not need to provide documentation to `gi-docgen` for private functions. Of course it's fine to have the documentation comment for developers.

core/cog-view.c
```diff
@@ -396,3 +437,20 @@ cog_view_get_use_key_bindings(CogView *self)
...
     return priv->use_key_bindings;
 }
+
+/**
+ * cog_view_get_viewport: (get-property viewport)
+ * @self: A view.
+ *
+ * Gets viewport where the view is attached.
+ *
+ * Returns: (transfer none) (nullable): The viewport what the view belongs.
```
aperezdc, line 447: This should be `(transfer full)` because the value returned from `g_weak_ref_get()` is also `(transfer full)`. A reference is added to the returned value when the weak reference is fetched to prevent the object to be freed while being used by the caller.

```suggestion
 * Returns: (transfer full) (nullable): The viewport what the view belongs.
```

platform/wayland/cog-view-wl.c
```diff
@@ -214,8 +212,7 @@ cog_wl_view_create_backend(CogView *view)
 bool
 cog_wl_view_does_image_match_win_size(CogWlView *view)
 {
-    CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
+    CogWlViewport *viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
```
aperezdc, line 215: The `cog_view_get_viewport()` function returns a `(transfer full)` value with an added reference, so we have to unref after using it. The easiest way is to use `g_autoptr`:

```suggestion
    g_autoptr(CogWlViewport) viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
```

platform/wayland/cog-view-wl.c
```diff
@@ -214,8 +212,7 @@ cog_wl_view_create_backend(CogView *view)
...
     return view->image && wpe_fdo_egl_exported_image_get_width(view->image) == viewport->window.width &&
            wpe_fdo_egl_exported_image_get_height(view->image) == viewport->window.height;
```
aperezdc, line 217: We also need to check that the viewport is valid, because `cog_view_get_viewport()` may return `NULL`:

```suggestion
    return viewport && view->image &&
           wpe_fdo_egl_exported_image_get_width(view->image) == viewport->window.width &&
           wpe_fdo_egl_exported_image_get_height(view->image) == viewport->window.height;
```

(Or whichever indentation the style checker prefers.)

platform/wayland/cog-view-wl.c
```diff
@@ -250,8 +246,8 @@ cog_wl_view_exit_fullscreen(CogWlView *view)
 static bool
 cog_wl_view_handle_dom_fullscreen_request(void *data, bool fullscreen)
 {
-    CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
+    CogWlView     *view = data;
+    CogWlViewport *viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
```
aperezdc, line 250: Use `g_autoptr` for the `viewport` pointer, and it needs to be null-checked.

platform/wayland/cog-view-wl.c
```diff
@@ -278,7 +274,7 @@ static void
 cog_wl_view_request_frame(CogWlView *view)
 {
     CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
+    CogWlViewport *viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
```
aperezdc, line 277: Same here: use `g_autoptr` for the `viewport` pointer, and it needs to be null-checked.

platform/wayland/cog-view-wl.c
```diff
@@ -301,8 +297,12 @@ void
 cog_wl_view_resize(CogWlView *view)
 {
     CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
+    g_assert(platform);
 
+    if (!platform->display || !platform->display->current_output)
+        return;
+
+    CogWlViewport *viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
```
aperezdc, line 305: Another one: `g_autoptr` for the `viewport` pointer and a null-check.

I think we may want to do a follow-up patch later in a separate PR to do all the null-checking on the viewport, but at least let's make sure we use `g_autoptr` for the pointers returned by `cog_view_get_viewport()` to avoid leaks.

aperezdc (changes requested):
Much nicer using weak pointers, thanks for the update @psaavedra. I still have a couple of comments regarding reference counting. There are also a number of null-checks missing but I think we can leave those for a separate patch.

core/cog-view.c
psaavedra: Ok but let's leave it as it is.

platform/wayland/cog-view-wl.c
```diff
@@ -228,8 +228,8 @@ cog_wl_view_enter_fullscreen(CogWlView *view)
         return;
 
 #if HAVE_FULLSCREEN_HANDLING
-    CogWlPlatform *platform = (CogWlPlatform *) cog_platform_get();
-    CogWlViewport *viewport = COG_WL_VIEWPORT(platform->viewport);
+    g_autoptr(CogWlViewport) viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
+    g_assert(viewport != NULL);
```
psaavedra, line 232: Handled with an assert since we should not fall in this situation I think.

platform/wayland/cog-view-wl.c
```diff
@@ -393,16 +406,15 @@ validate_exported_geometry(uint32_t width, uint32_t height)
...
+    CogWlView               *view = data;
+    g_autoptr(CogWlViewport) viewport = COG_WL_VIEWPORT(cog_view_get_viewport((CogView *) view));
 
     struct wl_resource   *exported_resource = wpe_fdo_shm_exported_buffer_get_resource(exported_buffer);
     struct wl_shm_buffer *exported_shm_buffer = wpe_fdo_shm_exported_buffer_get_shm_buffer(exported_buffer);
 
     uint32_t image_width = wl_shm_buffer_get_width(exported_shm_buffer);
     uint32_t image_height = wl_shm_buffer_get_height(exported_shm_buffer);
-    if (!validate_exported_geometry(image_width, image_height)) {
+    if (!viewport || !validate_exported_geometry(viewport, image_width, image_height)) {
```
psaavedra, line 417: Still we need to dispatch and release the generated frame

platform/wayland/cog-view-wl.c
psaavedra: Hopefully I added now all the null checks required.

core/cog-view.c
```diff
@@ -87,6 +91,9 @@ cog_view_get_property(GObject *object, unsigned prop_id, GValue *value, GParamSp
     case PROP_USE_KEY_BINDINGS:
         g_value_set_boolean(value, cog_view_get_use_key_bindings(self));
         break;
+    case PROP_VIEWPORT:
+        g_value_set_object(value, cog_view_get_viewport(self));
```
aperezdc, line 95: We need to use `g_value_take_object()` here, because `cog_view_get_viewport()` adds a ref to the object, and then `g_value_set_object()` adds _another_ ref. The `_take_object()` function “takes ownership” of the object, that is: does not add a reference, and assumes it can unref the object when it does not need it anymore.

(Sorry that I didn't notice this in my previous review.)

platform/wayland/cog-view-wl.c
aperezdc: That makes sense, yes 🆗

aperezdc (changes requested):
We need only a small fix to get this ready for merging, we're almost there now 🏃🏼‍♂️

aperezdc (approved):
We're good to go now, thanks @psaavedra!

core/cog-view.c
psaavedra: understood.

PR #668, opened 2023-11-23
Add CogView::cog_view_[set|is]_visible() accessors

psaavedra:
Add convenient accessors to the "visible" property the to the **CogView** external API:

* `CogView::cog_view_set_visible()` sets the view as the one visible inside of the viewport where the view is attached.
* `CogView::cog_view_is_visible()` gets if the view is visible inside of the viewport where the view is attached.

### DIFF
--- a/core/cog-view.c
+++ b/core/cog-view.c
@@ -454,3 +454,23 @@ cog_view_get_viewport(CogView *self)
     g_return_val_if_fail(COG_IS_VIEW(self), NULL);
     return g_weak_ref_get(&((CogViewPrivate *) cog_view_get_instance_private(self))->viewport);
 }
+
+/**
+ * cog_view_is_visible:
+ * @self: A view.
+ *
+ * Gets whether the view is visible in a viewport.
+ *
+ * Returns: True if the view is visible in its viewport and false in any
+ * other case.
+ *
+ * Since: 0.20
+ */
+gboolean
+cog_view_is_visible(CogView *self)
+{
+    g_return_val_if_fail(COG_IS_VIEW(self), false);
+
+    g_autoptr(CogViewport) viewport = cog_view_get_viewport(self);
+    return viewport && self == cog_viewport_get_visible_view(viewport);
+}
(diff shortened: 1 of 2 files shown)


core/cog-view.c
```diff
@@ -454,3 +454,51 @@ cog_view_get_viewport(CogView *self)
...
+
+/**
+ * cog_view_set_visible:
+ * @self: A view.
+ *
+ * Set the view as the one visible inside of the viewport where the view is
+ * attached.
+ *
+ * Returns: True if the view succesfully set as visible in its viewport and
+ * false in any other case.
```
aperezdc, line 489: Maybe the following is easier to parse (well, it is for me):

```suggestion
 * Set the view as visible in its viewport.
 *
 * Returns: %TRUE if the view was set as visible, or %FALSE if the view
 *    does is not attached to a viewport.
```

core/cog-view.c
```diff
@@ -454,3 +454,51 @@ cog_view_get_viewport(CogView *self)
...
+ *
+ * Since: 0.20
+ */
+gboolean
+cog_view_set_visible(CogView *self)
+{
+    g_assert(COG_IS_VIEW(self));
+
+    g_autoptr(CogViewport) viewport = cog_view_get_viewport(self);
+    g_return_val_if_fail(viewport != NULL, false);
```
aperezdc, line 499: I would return `FALSE` here if there is no viewport. If we want to consider that it is a programmer error to call this function when the view is not attached to a viewport, then the function does not need to return any value and use `g_return_if_fail()` here.

But I think it's better to do:

```suggestion
    if (!viewport)
        return FALSE;
```

core/cog-view.c
```diff
@@ -454,3 +454,51 @@ cog_view_get_viewport(CogView *self)
...
+
+    cog_viewport_set_visible_view(viewport, self);
+
+    return true;
```
aperezdc, line 503: The `true` value is for the `bool` type, but for `gboolean` we must use `TRUE`:

 ```suggestion
    return TRUE;
```

aperezdc (changes requested):
I agree on the idea for the new API, but I have a few change requests. If you look into them, I think we can merge this today. Thanks @psaavedra! 🙏🏼

psaavedra:
> I agree on the idea for the new API, but I have a few change requests. If you look into them, I think we can merge this today. Thanks @psaavedra! 🙏🏼

Done.

aperezdc (approved):

PR #672, opened 2023-11-23
platform: Refactor cursor code, and avoid compiler warnings

aperezdc:
While fixing an issuie related to cursors (PR coming later), I spent a few minutes cleaning up the code first, to avoid compiler warnings, and to factor out a bit the code. The logic that mapped a `WebKitHitTestResult` to the type of mouse cursor was repeated the same in three different places.

These two commits mainly move things around and define type aliases. The C type system has its warts and many would call it deficient, but anyway it's good to use it to make the code clearer.


psaavedra (approved):
DITTO  in Ditto in https://github.com/Igalia/cog/pull/673#pullrequestreview-1746599307

PR #673, opened 2023-11-23
x11: Do not attempt to set an invalid cursor

aperezdc:
Only set the cursor when the resolved cursor is _not_ `XCB_CURSOR_NONE`.


aperezdc:
This also includes the changes from #672 but I can rebase it on top of `master` if that's more convenient.

psaavedra (approved):
LGTM. I spent a bit more time because I have to learn about `WebKitHitTestResult`  and  how this work to identify the type of HTML clicked or interacted elements.

I appreciate also the restructuration done in  the first commit. The creation of the platform/common/cog-cursors.c and the encapsulation of the cursos types fixes should fix also an annoying build warning due to an unused definition. 

**r+**

PR #675, opened 2023-11-24
wl: Shell handles many viewports

psaavedra:
Add new functions in the CogShell API for handling multiple Viewports:
    
* `cog_shell_viewport_lookup(CogShell *, CogViewportKey)`
* `cog_shell_viewport_new(CogShell *, CogViewportKey)`
* `cog_shell_viewport_remove(CogShell *, CogViewportKey)`
    
Also added the `CogShell::create-viewport` signal. This signal is emitted when the shell creates a new **CogViewport**. With handling this     signal is possible to customize the required actions to be done during     the creation of a new viewport.
    
This change modifies the `cog_shell_get_viewport()`. Now it creates a default viewport in order to keep the current behavior. The creation     of the default viewport is lazy and it is delayed until the invocation of the `cog_shell_get_viewport()`.

Add a new multiple-stack example program which shows how to attach 2 viewports to the CogShell. Each viewport iterates over a couple of views.


platform/wayland/cog-platform-wl.c
```diff
@@ -1335,24 +1357,19 @@ cog_wl_platform_setup(CogPlatform *platform, CogShell *shell G_GNUC_UNUSED, cons
...
-        return FALSE;
-    }
-
     /* init WPE host data */
     wpe_fdo_initialize_for_egl_display(display->egl_display);
 
+    /* get a reference to the hashmap with viewports added to the shell */
+    self->viewports = cog_shell_get_viewports(shell);
+
+    g_signal_connect_swapped(shell, "create-viewport", G_CALLBACK(cog_wl_platform_on_create_viewport), self);
```
psaavedra, line 1366: This allows the specific platforms to react to it and run specific steps. Like here https://github.com/Igalia/cog/pull/675/files#diff-7342248ee4ad822cd294d9c8341a52fd10b3d1df4dd2a202ccb926df67d51034R1327  where the WL platform creates the CogWlWindow.

examples/viewports.c
```diff
@@ -0,0 +1,83 @@
+/*
+ * viewports.c
```
psaavedra, line 2: For a better user experience, I suggest to use this example with Sway or a similar  tiling Wayland-based window manager.

examples/viewports.c
```diff
@@ -0,0 +1,83 @@
...
+    if (!cog_platform_setup(platform, shell, g_getenv("COG_PLATFORM_PARAMS") ?: "", &error))
+        g_error("Cannot configure platform: %s", error->message);
+
+    g_autoptr(GMainLoop) loop = g_main_loop_new(NULL, FALSE);
+
+    CogViewport *viewport0 = cog_shell_get_viewport(shell);
+
+    for (int i = 1; i < argc; i++) {
+        g_autoptr(CogView) view = cog_view_new(NULL);
+        cog_platform_init_web_view(platform, WEBKIT_WEB_VIEW(view));
```
aperezdc, line 51: Having to call `cog_platform_init_web_view()` by hand is annoying. I'm sure we can change it in the future, we can add a `CogView:viewport` property (most of the code is already there) and then view code that needs late initialization when the view is known to have a viewport assigned can monitor the property. Currently what the `x11`, `drm`, and `gtk4` plug-ins do is that (late initializations) or other configurations which could be moved to the `GObject.constructed` virtual function. 

Anyway, this is not something we need to change now, I am writing my thoughts here so we can add an issue and make a patch later.

aperezdc (changes requested):
I have doubts about using a `GHashTable` and exposing that to the public `CogShell` API, because it looks to me that the code could be simpler and more consistent with the rest of the library using an array&mdash;please see the comments inline in the review.

If there is some reason I am missing to use the hash table, please let me know and we can chat about it 😺

examples/viewports.c
aperezdc: Issue created to handle this later: https://github.com/Igalia/cog/issues/678 🏗️

core/cog-shell.c
```diff
@@ -716,5 +737,121 @@ CogViewport *
...
+    if (size == 0) {
+        viewport = cog_viewport_new();
+        cog_shell_add_viewport(shell, viewport);
+    } else
+        viewport = cog_shell_get_nth_viewport(shell, COG_SHELL_DEFAULT_VIEWPORT_INDEX);
+    return viewport;
+}
+
+GPtrArray *
+cog_shell_get_viewports(CogShell *shell)
```
aperezdc, line 754: I'm not a fan of this, because in general it's a bad idea to use `GPtrArray` in public API (it is not introspectable) but provided that we would do #634 later I think we can land this patch as-is.

core/cog-shell.h
```diff
@@ -16,10 +16,14 @@
 
 G_BEGIN_DECLS
 
+#define COG_SHELL_DEFAULT_VIEWPORT_INDEX 0
+
 typedef struct _CogViewport   CogViewport;
 typedef struct _WebKitWebView WebKitWebView;
 
-#define COG_TYPE_SHELL  (cog_shell_get_type ())
+typedef gpointer CogViewportKey;
```
aperezdc, line 24: We don't need this `typedef` anymore 😉

aperezdc (approved):
We can land this, assuming we will continue improving things and we will get rid of `cog_shell_get_viewports()` later on.

PTAL at the comment about removing the `typedef` before merging.

😺

PR #679, opened 2023-11-30
x11: Support xcb-keysyms as fallback when XKB support not available

aperezdc:
These patch set pulls the parts of the code from #303 which implement usage of `xcb-keysyms` to handle keyboard input event translation, adding a few niceties on top:

- Allow passing `-Dx11_keyboard=` to Meson to choose which methods are supported. It is possible to use `xkb`, `xcb-keysyms`, or both. Passing an empty list disables keyboard input to web views.
- If support for both XKB and `xcb-keysyms`, then XKB will be preferred, and if not available then `xcb-keysyms` is used as fallback.
- Lastly, add a Wrap subproject definition that teaches Meson how to download a tarball and build a static `xcb-keysyms` library when not available. The library is a single `.c` file so it does not add much weight to the X11 platform plug-in, and we make it less likely that people may end up without keyboard input support.

Finally, I want to thank @agordon who was the person that originally wrote the code 🙏🏼

### DIFF
--- a/platform/x11/cog-platform-x11.c
+++ b/platform/x11/cog-platform-x11.c
@@ -23,7 +23,9 @@
 #    include <xcb/xcb_keysyms.h>
 #endif /* COG_X11_USE_XCB_KEYSYMS */
 
-#include <xkbcommon/xkbcommon-x11.h>
+#ifdef COG_X11_USE_XKB
+#    include <xkbcommon/xkbcommon-x11.h>
+#endif
 
 #if COG_HAVE_LIBPORTAL
 #    include "../common/cog-file-chooser.h"
@@ -87,6 +89,7 @@ struct CogX11Display {
         GSource *source;
     } xcb;
 
+#ifdef COG_X11_USE_XKB
     struct {
         int32_t             device_id;
         struct xkb_context *context;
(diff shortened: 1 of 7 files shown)


platform/x11/cog-platform-x11.c
```diff
@@ -899,8 +899,10 @@ cog_x11_platform_setup(CogPlatform *platform, CogShell *shell G_GNUC_UNUSED, con
         return FALSE;
     }
 
-    if (!init_keyboard(error))
-        return FALSE;
+    if (!init_keyboard(error)) {
```
psaavedra, line 906: **minor**: This could be squashed into https://github.com/Igalia/cog/pull/679/commits/27c30fbbd711a5fb13174b8c03af657a90701932  since affects to code added in this previous commit from this same PR.

psaavedra (approved):
LGTM. r+

Notice, I0've not tested alternatives builds with/without the COG_X11_USE_XCB_KEYSYMS manually. I assume this works.

platform/x11/cog-platform-x11.c
aperezdc: Done, squashed.

aperezdc:
> LGTM. r+
> 
> Notice, I've not tested alternatives builds with/without the COG_X11_USE_XCB_KEYSYMS manually. I assume this works.

I tested the build options here before submitting the PR 😄

Issue #614, opened 2023-10-17
REGRESSION(47dee03): Broke DRM plugin on RPi4


clopez:
Commit 47dee03 broke the DRM plugin on the RPi4 (My test build is based on Yocto version 4.2 (Mickeledore) with Mesa drivers)

I bisected it and reverting this commit fixes the issue.

To test this on the RPi4 you have to pass to Cog the `renderer=gles` platform parameters (see below how to execute cog)

On the terminal you can see this:

```
# cog --platform=drm --platform-params=renderer=gles https://igalia.com
EGLDisplay Initialization failed: EGL_NOT_INITIALIZED
Cog-Core-Message: 20:50:55.787: <https://igalia.com/> Load started.
Cog-Core-Message: 20:50:55.820: <https://www.igalia.com/> Redirected.
Cog-Core-Message: 20:50:56.157: <https://www.igalia.com/> Loading...

(cog:6469): Cog-DRM-WARNING **: 20:50:59.149: cog_drm_gles_renderer_handle_egl_image: Cannot set mode (Invalid argument)
Cog-Core-Message: 20:51:00.285: <https://www.igalia.com/> Loaded successfully.

```

And nothing gets rendered on the screen

The error `(cog:6469): Cog-DRM-WARNING **: 20:50:59.149: cog_drm_gles_renderer_handle_egl_image: Cannot set mode (Invalid argument)` is new.

If you run with env var ` G_MESSAGES_DEBUG=all`  then you get this: http://ix.io/4JfT

Related: https://github.com/Igalia/cog/pull/590

 //cc @zhani  @psaavedra  @aperezdc

clopez:
The following patch fixes the issue

```
diff --git a/platform/drm/cog-platform-drm.c b/platform/drm/cog-platform-drm.c
index be8521b..25a4f22 100644
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -359,6 +359,7 @@ find_crtc_for_encoder(const drmModeRes *resources, const drmModeEncoder *encoder
         const uint32_t crtc_mask = 1 << i;
         const uint32_t crtc_id = resources->crtcs[i];
         if (encoder->possible_crtcs & crtc_mask) {
+            //g_assert (encoder->crtc_id == crtc_id); // should this two values match? should be an if instead of an assert?
             return crtc_id;
         }
     }
@@ -517,6 +518,9 @@ init_drm(void)
             continue;
         }
 
[...]

PR #680, opened 2023-12-12
REGRESSION(47dee03): Broke DRM plugin on RPi4 (#614)

clopez:
`find_crtc_for_encoder()` should not only check the bitmask but also the ID of the CRTC.

See #614 for a detailed description of the issue that this PR fixes.


aperezdc (approved):
LGTM, seems reasonable after reading #614 to understand the change.

PR #685, opened 2023-12-20
core: Remove view and viewport handling from CogShell

aperezdc:
Remove handling of viewports and views from `CogShell`. Tracking live viewports can be done using `CogPlatform` since #683 and platform plug-ins no longer depend on `CogShell` for that. The following are no longer needed in `CogShell`:

- The `:web-view` and `:viewport` properties, and their accessors.
- The `::create-view` and `::create-viewport` signals.
- The `.create_view()` virtual function and its default implementation.
- The `cog_shell_add_viewport()`, `cog_shell_remove_viewport()`, `cog_shell_get_n_viewports()`, and `cog_shell_get_nth_viewport()` methods.

As a side effect of the above removals, the examples have been changed to have them create their views using `cog_view_new()` instead of relying on the default implementation of the `CogShell.create_view()` virtual function.

The launcher also needed a number of changes, which now manages itself a viewport as it cannot rely on the shell for that. Some code was moved around to accommodate creation of the view in `cog_launcher_startup()` without relying on the `CogShell::create-view` signal, and similarly for web view creation when running in automation mode.

There are more opportunities for further simplifications that would be better done in separate patches: this one focuses on the minimum amount of changes needed to get view and viewport handling out of `CogShell`.

### DIFF
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -121,25 +121,6 @@ request_handler_map_entry_register (const char             *scheme,
 }
 
 static void
-cog_shell_startup_base(CogShell *shell)
-{
-    CogShellPrivate *priv = PRIV(shell);
-
-    if (priv->request_handlers) {
-        g_hash_table_foreach (priv->request_handlers,
-                              (GHFunc) request_handler_map_entry_register,
-                              priv->web_context);
-    }
-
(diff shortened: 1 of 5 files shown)


psaavedra:
tested on my rpi3.

psaavedra (approved):

aperezdc:
Thanks @psaavedra!

PR #689, opened 2024-01-08
wl: Specify button modifier in listeners

### LINKED ISSUE #688: MouseEvent buttons property not set in wayland backend
To replicate the issue you can visit https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons#result, in the `Result` pane try to click. In Firefox the correct button is detected, while in cog it is always 0.

This breaks MUI Slider component, you can try it [here](https://mui.com/material-ui/react-slider/)

Tested with cog master

matperc:
TODO:
- [ ] Is it necessary to fix the button value passed through WPE? It doesn't seem to affect the result in my tests
- [X] ~Implement back and forward buttons (I don't have the hardware right now to test it).~ I can't test the extra buttons since they change the browser page (obviously...). So I'll just removed them. The behavior is the same as before for those buttons

Closes #688

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -11,6 +11,7 @@
 
 #include <glib-object.h>
 #include <glib.h>
+#include <linux/input-event-codes.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


aperezdc (approved):
This looks good to me, thanks @matperc and apologies for the late review 🙏🏼

aperezdc:
Backported into the `cog-0.18` branch as commit f2bbcbf8973d5a582a695d7f7d436df8e2900e6c

PR #692, opened 2024-01-26
automation: Save automated view and avoid dangling ref when closing window

lauromoura:
The view created for automation was not being saved in the viewport, leaving some required signals unhandled, like "close". This causes the view to never be destroyed when called from WebKit's WebPageProxy::close(), leading to infinite loop in the close window WebDriver algorithm, as the list of open windows would never change.

After hooking the close signal, there was the issue with cog_viewport_add ref'ing the view when storing the view with g_ptr_array_add but not releasing this reference when removing, doing just a ref/unref pair to preserve during the removal.

This commit removes the extra g_object_ref during the cleanup, effectively relying on the previously saved ref before unref'ing it

This should fix the WPT webdriver fixtures that rely on creating and destroying views for testing.

Note: This commit makes creating new windows for automation behave like creating new tabs, regardless of the "create-web-view::window" detail, which maybe could be implemented creating a new viewport in a future commit.

### DIFF
--- a/launcher/cog-launcher.c
+++ b/launcher/cog-launcher.c
@@ -240,7 +240,7 @@ on_automation_session_create_web_view(WebKitAutomationSession *session, CogLaunc
         webkit_website_policies_new_with_policies("autoplay", s_options.autoplay_policy, NULL);
 #endif /* HAVE_WEBKIT_AUTOPLAY */
 
-    WebKitWebView *new_view =
+    g_autoptr(WebKitWebView) new_view =
         (WebKitWebView *) cog_view_new("settings", cog_shell_get_web_settings(launcher->shell), "web-context",
                                        cog_shell_get_web_context(launcher->shell), "is-controlled-by-automation", TRUE,
                                        "zoom-level", s_options.scale_factor, "use-key-bindings", FALSE,
@@ -253,8 +253,6 @@ on_automation_session_create_web_view(WebKitAutomationSession *session, CogLaunc
                                        NULL);
     // FIXME New window should be new viewport
     cog_viewport_add(launcher->viewport, (CogView *) new_view);
-    // Let viewport own the view
-    g_object_unref(new_view);
 
     return new_view;
 }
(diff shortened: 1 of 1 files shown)


launcher/cog-launcher.c
```diff
@@ -230,21 +230,33 @@ on_web_view_create(WebKitWebView *web_view, WebKitNavigationAction *action)
...
+    }
 #if HAVE_WEBKIT_AUTOPLAY
     g_autoptr(WebKitWebsitePolicies) website_policies =
         webkit_website_policies_new_with_policies("autoplay", s_options.autoplay_policy, NULL);
 #endif /* HAVE_WEBKIT_AUTOPLAY */
 
-    return (WebKitWebView *) cog_view_new("settings", cog_shell_get_web_settings(launcher->shell), "web-context",
-                                          cog_shell_get_web_context(launcher->shell), "is-controlled-by-automation",
-                                          TRUE, "zoom-level", s_options.scale_factor, "use-key-bindings", FALSE,
+    WebKitWebView *new_view =
```
aperezdc, line 243: You could use `g_autoptr()` here and then we do not need the `g_object_unref()` below:

```suggestion
    g_autoptr(WebKitWebView) new_view =
```

launcher/cog-launcher.c
```diff
@@ -230,21 +230,33 @@ on_web_view_create(WebKitWebView *web_view, WebKitNavigationAction *action)
...
 #if HAVE_WEBKIT_AUTOPLAY
-                                          "website-policies", website_policies,
+                                       "website-policies", website_policies,
 #endif /* HAVE_WEBKIT_AUTOPLAY */
-                                          NULL);
+                                       NULL);
+    // FIXME New window should be new viewport
+    cog_viewport_add(launcher->viewport, (CogView *) new_view);
+    // Let viewport own the view
+    g_object_unref(new_view);
```
aperezdc, line 257: Then you don't need this.

```suggestion
```

aperezdc (approved):
Patch LGTM, but please check the comment first for a small simplification you may want to take into account before merging 😺

PR #695, opened 2024-02-26
Fixed the 0 device scale factor issue in the drm platform.

Mikael-Eliasson:
Copied the code for fetching the device scale factor from the gtk platform.

The problem was that a scale factor of 0.0 was read before which just results in a black screen.

I am not certain if the "g_signal_connect" call is where you want it.

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -1319,6 +1319,12 @@ key_repeat_source_dispatch(CogDrmPlatform *platform)
 }
 
 static void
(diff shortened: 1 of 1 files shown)


aperezdc (approved):
@Mikael-Eliasson Thanks, this looks good!

The style checker is having some trouble and giving false positives. I am going to approve and merge this anyway.

aperezdc:
Cherry-picked into the `cog-0.18` branch as commit 2c8fc7c4da29281f75f5101ea949f3030a897ba9

PR #700, opened 2024-03-07
wl: Prefer SSD via zxdg_decoration_manager_v1

aperezdc:
Ask compositors that support the `zxdg_decoration_manager_v1_interface` protocol to provide server-side decorations (SSD). This way it is more likely that the Cog surfaces will have window decorations. Note that compositors will still hide the decorations of fullscreen surfaces, but maximized ones will keep them, and this is the expected behaviour. Tested on Sway and Labwc, and should work in most (if not all) wlroots based compositors.


aperezdc:
Screenshot, running under Labwc:

Screenshot of Cog running under Labwc, with window decorations

PR #701, opened 2024-03-09
wl: Add safe-guards for touch and pointer events

psaavedra:
In some compositors it could be possible these 2 cases:

* to receive touch up/motion events immediately after a touch up (touch deinit the touch target).
* to receive pointer motion/leave event immediately after a touch leave (deinit the touch target).

This patch adds safeguards in the touch events callbacks

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -271,7 +271,13 @@ pointer_on_enter(void              *data,
 static void
 pointer_on_leave(void *data, struct wl_pointer *pointer, uint32_t serial, struct wl_surface *surface)
 {
+    if (data == NULL || pointer == NULL)
+        return;
+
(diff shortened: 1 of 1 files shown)


platform/wayland/cog-platform-wl.c
```diff
@@ -271,8 +271,14 @@ pointer_on_enter(void              *data,
 static void
 pointer_on_leave(void *data, struct wl_pointer *pointer, uint32_t serial, struct wl_surface *surface)
 {
+    if (data == NULL || pointer == NULL)
```
aperezdc, line 274: In general it's not needed to explicitly compare with `NULL`, which tends to be clearer:

```suggestion
    if (!data || !pointer)
```

aperezdc (approved):
Thanks @psaavedra 👌🏼

PR #702, opened 2024-03-13
wl: Fix popup menu item 6 & 7 not rendering

gebausim:
When HTML select had 6 or 7 options those respective options were not rendered, they still functioned fine just not visible.

Before:

before

After:

after


aperezdc (approved):
Thanks for the fix @gebausim, good catch!

aperezdc:
Cherry-picked into the `cog-0.18` branch as commit abe690795371ef05d0f9bbcfb37324b4b7a5794a

PR #706, opened 2024-04-08
platform: fix compatibility with ln(1) without relative support

q66:
This reduces the assumptions allowing cog to build cleanly with any version of ln.


aperezdc (approved):
Thanks for the fix @q66!

aperezdc:
Backported to `cog-0.18` as commit 6dd7a29908f854e6b3a9526423797bc3770cf385

PR #707, opened 2024-04-24
Fix documentation of COG_PLATFORM_DRM_VIDEO_MODE

septatrix:
The docs stated that it accept a refresh rate which is not true


aperezdc (approved):
Thanks @septatrix, good documentation update... and sorry for the delay in reviewing this.

PR #716, opened 2024-06-13
core, wl: Factor out view fullscreening

aperezdc:
Move view fullscreening to a new vfunc in the `CogView` class, which gets in turn implemented by `CogWlView`. This moves the entry point to common code, which would allow other platform plug-ins to implement the functionality if desired. Also, handling its key binding can be moved along to common code.


aperezdc:
Ping reviewers 🏓

csaavedra (approved):

PR #717, opened 2024-06-13
wl: Only use xdg-decoration if it's supported by the compositor

mnme:
Using the xdg-compositor protocol on a compositor that doesn't support it produces a segfault. Guard against this.


TingPing:
CI suggests some minor code style changes, otherwise LGTM.

mnme:
Thanks, I amended the necessary changes. Already thought the line has gotten a big long, good to know you have a style checker script. Might be worth to mention that in the contribution guide.

TingPing (approved):

PR #729, opened 2024-08-22
Set entry->registered to TRUE when custom URI scheme handler is regis…

### LINKED ISSUE #727: RequestHandlerMapEntry->registered not being set to TRUE when handler is registered.
It appears `RequestHandlerMapEntry->registered` is not being set to `TRUE` in `cog_shell.c:request_handler_map_entry_register()` after `webkit_web_context_register_uri_scheme()` is called.

It's not a fatal error, but `webkit_web_context_register_uri_scheme()` does report an error if a program calls `cog_shell_set_request_handler()` and then calling `cog_shell_startup()`.

Setting `entry->registered` to `TRUE` inside `cog_shell.c:request_handler_map_entry_register()` suppresses the error message.

ziebarthw:
…tered

Sets the `entry->registered` flag to `TRUE` after registering the custom URI handler with WebKit.

Fixes #727

### DIFF
--- a/core/cog-shell.c
+++ b/core/cog-shell.c
@@ -117,6 +117,7 @@ request_handler_map_entry_register (const char             *scheme,
                                                 handle_uri_scheme_request,
                                                 entry,
                                                 NULL);
+        entry->registered = TRUE;
     }
 }
 


aperezdc (approved):
Good catch, thanks!

aperezdc:
Merged into the `cog-0.18` branch as commit d51b26e758f028b5623457aa50ed96676bfabe3b

PR #730, opened 2024-08-28
wl: fix key repeat timeout

rawoul:
Key repeat rate must be converted from keys/second to a period to schedule the repeat timer.


aperezdc (approved):
Good catch, thanks @rawoul!

aperezdc:
Cherry-picked into the `cog-0.18` branch as commit 31886ec11de49b6932cb0eaa3d149ea48e43cdbc

PR #731, opened 2024-08-28
build: make libmanette support explicit

rawoul:
This allows disabling libmanette support when libmanette is installed.


aperezdc (approved):
Thanks for the improvement, @rawoul!

PR #732, opened 2024-08-28
launcher: set preferred language list from system locale

rawoul:
This changes the Accept-Language header to match the system locale by default.


launcher/cog-launcher.c
```diff
@@ -318,6 +320,24 @@ cog_launcher_startup(GApplication *application)
 
     g_object_set(self->shell, "device-scale-factor", s_options.device_scale_factor, NULL);
 
+    /* build Accept-Language from locale, filtering out encodings */
+    locales = g_get_language_names();
+    if (locales) {
+        guint   i, n, length = g_strv_length((gchar **) locales);
+        gchar **languages = g_new0(gchar *, length);
```
TingPing, line 327: `length + 1`, to ensure `NULL` termination?

launcher/cog-launcher.c
```diff
@@ -318,6 +320,24 @@ cog_launcher_startup(GApplication *application)
...
+
+        for (n = 0, i = 0; locales[i]; i++) {
+            const gchar *lang = locales[i];
+            if (!strcmp(lang, "C") || !strcmp(lang, "POSIX") || strchr(lang, '.'))
+                continue;
+            languages[n++] = g_strdup(lang);
```
TingPing, line 333: If we immediately free this why copy it?

launcher/cog-launcher.c
```diff
@@ -318,6 +320,24 @@ cog_launcher_startup(GApplication *application)
 
     g_object_set(self->shell, "device-scale-factor", s_options.device_scale_factor, NULL);
 
+    /* build Accept-Language from locale, filtering out encodings */
+    locales = g_get_language_names();
+    if (locales) {
```
TingPing, line 325: `g_get_language_names()` never returns `NULL`.

rawoul:
Thanks for the review, I've fixed the issues you raised and factored the code in a separate function for clarity.

TingPing:
So the patch looks good to me.

I have one concern though, this exposes system information directly to the web which adds information for fingerprinting. Browsers like Safari specifically filter and reduce this list to make it less unique.

launcher/cog-launcher.c
```diff
@@ -267,6 +267,29 @@ on_automation_started(WebKitWebContext *context, WebKitAutomationSession *sessio
...
+static void
+cog_launcher_set_preferred_languages(CogLauncher *self)
+{
+    const gchar *const *locales;
+    const gchar       **languages;
+    guint               i, n, length;
+
+    /* build Accept-Language from locale, filtering out encodings */
+    locales = g_get_language_names();
+    length = g_strv_length((gchar **) g_get_language_names());
```
TingPing, line 279: I think we might have to reformat the languages.

MDN says the format is `en-US` yet this will return `en_US`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

PR #745, opened 2024-11-27
docs: webdriver: update code for selenium 4

jonrebm:
The desired_capabilities property has been deprecated in Selenium 4.0 Alpha 7 and dropped in Selenium 4.17.0.

Use options.capabilities instead.


aperezdc (approved):
Thanks for the update, @jonrebm!

PR #752, opened 2025-02-12
CI: Update to latest actions version and base image

aperezdc:
SSIA 😉

### DIFF
--- a/.github/workflows/ci-native.yml
+++ b/.github/workflows/ci-native.yml
@@ -13,12 +13,16 @@ on:
 
 jobs:
   build:
+    # TODO: It would be nice to move away from Ubuntu 22.04, but that would
+    #       need either building WPE WebKit, finding a PPA for it, or using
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-native.yml
```diff
@@ -13,12 +13,16 @@ on:
 
 jobs:
   build:
+    # TODO: It would be nice to move away from Ubuntu 22.04, but that would
```
TingPing, line 16: The webkit-container-sdk is a 24.04 image with the dependencies :)

Can make an image based on it with webkit built.

TingPing (approved):

.github/workflows/ci-native.yml
aperezdc: Yes, we could but GitHub self-hosted runners won't take an arbitrary image. That would mean switching over to our own runner, so I would rather do it in separately.

PR #760, opened 2025-05-29
wl: fix the button of pointer button events

carlosgcampos:
We should pass the right value and for both press and release events.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -372,14 +372,12 @@ pointer_on_button(void              *data,
 
     seat->pointer.serial = serial;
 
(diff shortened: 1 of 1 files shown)


aperezdc (approved):
Good catch 😮

aperezdc:
Backported into the `cog-0.18` branch as commit 93de1e44907415315e750b9a2db8ce5465873388

aperezdc:
> Hello, we're waiting for this fix, so would it be possible to release version 0.18.5 containing this fix in near future?

[Cog 0.18.5 has been released](https://wpewebkit.org/release/cog-0.18.5.html) and it includes this fix.

Issue #474, opened 2022-08-18
cogctl communication to multiple instances does not work


ventosus:
With two cog instances (built with system DBUS):

    cog --gapplication-app-id=com.example.ONE uri
    cog --gapplication-app-id=com.example.TWO uri

There is no way to talk to both of them with cogctl by using the object path:

    cogctl --system --object-path=/com/example/ONE reload -> OK

    cogctl --system --object-path=/com/example/TWO reload
    GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path ?/com/example/TWO?

There is no way to talk to both of them with cogctl by using the appid:

    cogctl --system --appid=com.example.ONE reload
    GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Name "com.example.ONE" does not exist

    cogctl --system --appid=com.example.TWO reload
    GDBus.Error:org.freedesktop.DBus.Error.NameHasNoOwner: Name "com.example.TWO" does not exist

How should I talk to different cog instances? Am I doing something wrong?

ventosus:
# cog --version
    0.12.4 (WPE WebKit 2.36.3)

chrisvollorenew:
Hello @ventosus - I also have been having trouble with cogctl and dbus. 
I wonder if theres something I am missing with my set up. 
I am running cog as root in an embedded environment. 
The system bus is at /run/dbus/...
The environment variables for dbus are not getting set "automatically" in my case. 
Have you tried checking your environment variables (such as DBUS_SYSTEM_BUS_ADDRESS) for dbus?

Please find my post here:
https://github.com/Igalia/cog/issues/532

ventosus:
@chrisvollorenew dbus ist set up just fine on my side. my problem is addressing multiple instances of cog. addressing a single instance works just fine.

aperezdc:
The same identifiers passed to `cog --gapplication-app-id=` must be passed to `cogctl --appid=`:

```sh
# Start two instances with different application identifiers
cog --gapplication-app-id=com.example.Cog1 wpewebkit.org
cog --gapplication-app-id=com.example.Cog2 igalia.com
# Load different pages on each using cogctl
cogctl --appid=com.example.Cog1 open webkit.org
cogctl --appid=com.example.Cog2 open apple.com
```
You can see this in action in the screenshot below (note that `-A` is the same as `--appid=`):

[...]

PR #762, opened 2025-06-13
cog-launcher: fix system dbus appid connection

woutervanh:
On system dbus connection, the appid passed to the launcher was not taken into consideration yet, so taking the default id. This made it impossible to address multiple browser instances on the system bus. I moved the call to a place where we can take the appid. The end user is still responsible for creating the correct policy file!

Fixes:
https://github.com/Igalia/cog/issues/474


aperezdc (approved):
Patch looks good to me, thanks!

woutervanh:
done

PR #779, opened 2026-03-30
drm: Respect XDG_SEAT for libinput seat selection

alchark:
Instead of hardcoding "seat0" for enumerating available inputs under the DRM platform, use the value optionally provided by the standard environment variable XDG_SEAT, and fall back to the previous default if unavailable.

This allows running cog on a dedicated seat with its own display and input devices, while leaving the default seat0 available for other uses.

### DIFF
--- a/platform/drm/cog-platform-drm.c
+++ b/platform/drm/cog-platform-drm.c
@@ -68,6 +68,7 @@ struct _CogDrmPlatform {
     CogDrmRenderer        *renderer;
     CogGLRendererRotation  rotation;
     GList                 *rotatable_input_devices;
+    char                  *seat_name;
     bool                   draw_cursor;
     bool                   dispatch_pointer;
     bool                   use_gles;
@@ -246,11 +247,24 @@ parse_bool(const char *value, bool *out)
 }
 
 static void
+set_seat_name(CogDrmPlatform *self, const char *seat_name)
+{
+    g_free(self->seat_name);
+    self->seat_name = g_strdup(seat_name);
+}
+
+static void
 init_config(CogDrmPlatform *self, CogShell *shell, const char *params_string)
 {
     drm_data.device_scale = cog_shell_get_device_scale_factor (shell);
     g_debug ("init_config: overriding device_scale value, using %.2f from shell",
              drm_data.device_scale);
+
+    {
+        const char *env_seat_name = g_getenv("XDG_SEAT");
+        if (env_seat_name && *env_seat_name)
+            set_seat_name(self, env_seat_name);
+    }
 
     self->draw_cursor = g_getenv("COG_PLATFORM_DRM_CURSOR") != NULL;
     self->dispatch_pointer = g_getenv("COG_PLATFORM_DRM_POINTER") != NULL;
@@ -287,6 +301,12 @@ init_config(CogDrmPlatform *self, CogShell *shell, const char *params_string)
                 self->use_gles = false;
             else if (value)
                 g_warning("Invalid renderer '%s', using default.", value);
+        }
+
+        {
+            g_autofree char *value = g_key_file_get_string(key_file, "drm", "seat", NULL);
+            if (value && *value)
+                set_seat_name(self, value);
         }
 
         {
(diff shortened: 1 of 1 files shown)


platform/drm/cog-platform-drm.c
```diff
@@ -1292,7 +1293,12 @@ init_input(CogDrmPlatform *platform)
...
 
-    int ret = libinput_udev_assign_seat (input_data.libinput, "seat0");
+    if (!seat_name || !*seat_name)
+        seat_name = "seat0";
+
+    g_debug("%s: assigning libinput to seat '%s'", __func__, seat_name);
+
+    int ret = libinput_udev_assign_seat(input_data.libinput, seat_name);
     if (ret)
         return FALSE;
```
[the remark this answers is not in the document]
alchark: Not related to this change, separate issue

aperezdc (changes requested):
@alchark I like the idea of allowing configuring the seat, but rather than using only an environment variable, it would be better to allow setting it in the configuration file and/or via the `-O`/`--platform-params=` command line option. Would you be open to try and implement that in the `init_config()` function?

P.S: Please don't use Copilot for code reviews.

alchark:
@aperezdc sure, let me try and add config/cmdline parsing too. I believe the precedence should be:
1. Use cmdline value if available
2. Else use config value if available
3. Else use environment variable value if available
4. Else fallback to `seat0`

As for Copilot, I don't think I did anything to trigger it, but it might be some "helpful" default buried somewhere in Github settings. Any pointers welcome :-D

aperezdc:
> @aperezdc sure, let me try and add config/cmdline parsing too. I believe the precedence should be:
> 
>     1. Use cmdline value if available
> 
>     2. Else use config value if available
> 
>     3. Else use environment variable value if available
> 
>     4. Else fallback to `seat0`

This makes sense, indeed. 

Also I was checking around and it looks like the name of the `XDG_SEAT` environment variable is commonly used by Wayland compositors and other applications. So we may want to actually keep using that for consistency with the rest of the ecosystem.

> As for Copilot, I don't think I did anything to trigger it, but it might be some "helpful" default buried somewhere in Github settings. Any pointers welcome :-D

It looks like it was some repo setting that triggered it automatically, hopefully it is disabled now after some fiddling with the repo settings (not they make it easy to disable 🙃).

platform/drm/cog-platform-drm.c
```diff
@@ -226,13 +227,25 @@ static struct {
...
 static void
 init_config(CogDrmPlatform *self, CogShell *shell, const char *params_string)
 {
     drm_data.device_scale = cog_shell_get_device_scale_factor (shell);
     g_debug ("init_config: overriding device_scale value, using %.2f from shell",
              drm_data.device_scale);
 
+    {
+        const char *env_seat_name = g_getenv("XDG_SEAT");
+        set_seat_name(self, (env_seat_name && *env_seat_name) ? env_seat_name : "seat0");
```
aperezdc: I would set the default `seat0` value in `cog_drm_platform_init()`, and here only modify it if the environment variable is defined and non-empty. I think that would make the logic easier to follow:


```suggestion
        const char *env_seat_name = g_getenv("XDG_SEAT");
        if (env_seat_name && *env_seat_name)
            set_seat_name(self, env_seat_name);
```

aperezdc (dismissed):
@alchark Thanks for iterating on this. I left one small comment, and otherwise the patch looks good to me. Could you take a look?

(Don't worry about the failed ARM build in the CI, the job is having some issue picking the toolchain, so that's on us.)

platform/drm/cog-platform-drm.c
alchark: Thanks Adrian, this makes sense. Your version is more readable than the nested ternary operator.

zhovner:
@alchark can we resolve conflict please?

So we can ask Patrick @TingPing to look here too.

TingPing (approved):

TingPing:
> This branch cannot be rebased due to conflicts

alchark:
> > This branch cannot be rebased due to conflicts

That's weird. Let me rebase manually and re-publish, sorry for the noise.

alchark:
@TingPing would you mind giving this another look? No conflicts now.

PR #781, opened 2026-04-19
platform-wl: Reuse wl_display between being checked and being used

TingPing:
In normal cases this shouldn't meaningfully change behavior.

However if WAYLAND_SOCKET is set to an fd then wl_display_disconnect() closes that fd leaving it in a state where it cannot be reused.

### DIFF
--- a/platform/wayland/cog-platform-wl.c
+++ b/platform/wayland/cog-platform-wl.c
@@ -1155,12 +1155,15 @@ static const struct wpe_video_plane_display_dmabuf_receiver video_plane_display_
 };
 #endif
 
+/* Saved by check_supported(), consumed once by init_wayland(). */
+static struct wl_display *checked_wl_display = NULL;
+
 static gboolean
 init_wayland(CogWlPlatform *platform, GError **error)
 {
     g_debug("Initializing Wayland...");
 
-    CogWlDisplay *display = cog_wl_display_create(NULL, error);
+    CogWlDisplay *display = cog_wl_display_create(g_steal_pointer(&checked_wl_display), error);
     platform->display = display;
     display->registry = wl_display_get_registry(display->display);
     g_assert(display->registry);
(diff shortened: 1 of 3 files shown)


platform/wayland/cog-platform-wl.c
```diff
@@ -1155,12 +1155,15 @@ static const struct wpe_video_plane_display_dmabuf_receiver video_plane_display_
...
+/* Saved by check_supported(), consumed once by init_wayland(). */
+static struct wl_display *checked_wl_display = NULL;
+
 static gboolean
 init_wayland(CogWlPlatform *platform, GError **error)
 {
     g_debug("Initializing Wayland...");
 
-    CogWlDisplay *display = cog_wl_display_create(NULL, error);
+    CogWlDisplay *display = cog_wl_display_create(g_steal_pointer(&checked_wl_display), error);
```
aperezdc, line 1166: Maybe we could do this only if `WAYLAND_SOCKET` was set? Or as an alternative, if the environment variable is set, we can `dup()` the file descriptor and open the duplicate for the check.

Either way, the idea is to avoid keeping a connected `wl_display` in case Cog ends up picking a different plugin in the end. But also I am not sure if it's worth the effort 🤔

aperezdc (approved):
Even when I have a small doubt (see comment), this solves the actual issue of not being able to open the connection to the Wayland compositor when `WAYLAND_SOCKET` is defined in the environment, so I think we should merge it.

If we find later on that we do not want to keep the `wl_display` alive we make a patch later for that.

aperezdc:
The ARM cross-compilation is broken (#787) so let's merge this anyway.

platform/wayland/cog-platform-wl.c
TingPing: I tried `dup` but it didn't work. It’s left in a bad state if it’s reused anyway. 

Originally i also limited it to when that's set but AFAICT if the plugin is ever checked for support and returns TRUE it's always used, so there is no risk of it left open.

